PR #42, opened 2026-03-30
fix: remove double /api prefix in analytics API call

shapor:
## Summary
- Fix `Analytics.tsx` requesting `/api/api/analytics` instead of `/api/analytics`
- The axios instance has `baseURL: '/api'`, so the path should be `/analytics` not `/api/analytics`

Before:
Screenshot from 2026-03-30 04-46-31

## Test plan
- [ ] Analytics Dashboard page loads data (was showing 404 error)
- [ ] Network tab shows request to `/api/analytics?timeRange=7d` (not `/api/api/...`)

### DIFF
--- a/frontend/src/pages/Analytics.tsx
+++ b/frontend/src/pages/Analytics.tsx
@@ -159,7 +159,7 @@ export default function Analytics() {
     setLoading(true)
     setError(null)
     try {
-      const response = await api.get(`/api/analytics?timeRange=${timeRange}`)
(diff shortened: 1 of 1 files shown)


mattmorr1:
- Puts {decision_id} back in first position, before /stats and /pending-feedback. This re-introduces the original bug where those endpoints 404.
- Breaks analytics page due to the change in routing. 

The correct fix if they wanted to use the baseURL properly would be api.get('analytics?...')

PR #47, opened 2026-03-31
fixed render issue with ci/cd

mattmorr1:
AutoInvestigateTab.tsx — Replaced 13 MUI v6-style <Grid size={{ xs, sm }}> props with MUI v5-compatible <Grid item xs={N} sm={N}> across the    
  Agent Limits, Cost Guardrails, Timing & Advanced, and model/directory field sections. This was the root cause of the TypeScript build failures  
  in CI.                                                                                                                                          
                                                                                                                                                
  DetectionRulesTab.tsx — Two fixes:                                                                                                              
  - Removed unused const theme = useTheme() and its import (was causing an unused variable TS error)
  - Narrowed format state type from string to the proper union 'sigma' | 'splunk' | 'elastic' | 'kql' | 'auto' so it satisfies the addSource() API
   type                                                                                                                                           
                                                                                                                                                  
  ClaudeDrawer.tsx — Changed two let declarations (thinkingContent, textContent) to const since the arrays are never reassigned (only mutated via
  .push()). Likely a lint/TS strict-mode fix. 
  
  imperfect fix, will need to migrate to mui v6 (will get to this week)

### DIFF
--- a/frontend/src/components/settings/DetectionRulesTab.tsx
+++ b/frontend/src/components/settings/DetectionRulesTab.tsx
@@ -73,6 +73,10 @@ const FORMAT_COLORS: Record<string, string> = {
   auto: '#607d8b',
 }
 
+const RULE_FORMATS = ['sigma', 'splunk', 'elastic', 'kql', 'auto'] as const
+type RuleFormat = typeof RULE_FORMATS[number]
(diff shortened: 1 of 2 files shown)


mattmorr1:
@mando222 pls merge when you have the chance :)

PR #51, opened 2026-04-06
fix(monitoring): wire Sentry and Prometheus — both were dead code

mando222:
Sentry: init_sentry() was defined but never called. Added call in backend/main.py at startup (no-op when SENTRY_DSN is unset).

Prometheus: init_prometheus_metrics() defined metrics as local variables that were immediately garbage-collected. Moved metrics to module level, added PrometheusMiddleware (records request count + duration), and added /metrics endpoint on the backend at port 6987.

Adds 13 unit tests covering Sentry init behavior, before_send filter, Prometheus availability, middleware bypass for /metrics, and the get_metrics_response helper. Tests run without sentry-sdk or prometheus-client installed via sys.modules stubs.

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -66,6 +66,7 @@ from api.detection_rules import router as detection_rules_router
 from api.orchestrator import router as orchestrator_router
 
 from core.rate_limit import rate_limit_dependency
+from monitoring import init_sentry, PROMETHEUS_AVAILABLE, PrometheusMiddleware, get_metrics_response
 
 # Configure logging
 log_dir = Path.home() / '.deeptempo'
@@ -81,6 +82,9 @@ logging.basicConfig(
 )
 
 logger = logging.getLogger(__name__)
+
+# Initialize Sentry as early as possible (no-op if SENTRY_DSN is unset)
+init_sentry()
 
 # Create FastAPI app
 app = FastAPI(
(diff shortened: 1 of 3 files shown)


mattmorr1:
a few other notes: reading the metrics we are using, not sure if these are the only ones we want: maybe expose cost (i was also going to update cost to pull api cost in future api since rn it's hard coded) 

active_cases_total, findings_processed_total are defined but not set properly. 

claude had some ideas:
 Cost/LLM                                                                                        
  - vigil_llm_input_tokens_total, vigil_llm_output_tokens_total, vigil_llm_cost_usd_total
  - vigil_orchestrator_hourly_cost_usd gauge (budget guardrail visibility)                                                                      
                                                            
  Investigations/Agents                                                                                                                         
  - vigil_orchestrator_active_agents gauge                  
  - vigil_orchestrator_investigations_completed_total / failed_total                                                                            
  - vigil_agent_tool_calls_total{tool_tier} — especially approval-tier calls
                                                                                                                                                
  Pipeline health                                                                                                                               
  - vigil_poller_findings_total{source} + vigil_poller_errors_total{source} — detect broken integrations                                        
  - vigil_processor_triage_duration_seconds histogram — LLM triage latency                                                                      
  - vigil_responder_auto_executed_total vs vigil_responder_escalated_total
                                                                                                                                                
  MCP tools                                                                                                                                     
  - vigil_mcp_tool_calls_total{tool_name} + vigil_mcp_tool_errors_total — tells you which integrations are flaky                                
                                                                                                                                                
  SOC-specific ratios to alert on:                          
  - findings_ingested / investigations_started — alert fatigue signal                                                                           
  - dedup_prevented_total — shared intel effectiveness                                                                                          
  - approvals_requested vs approvals_granted/rejected — human-in-loop bottleneck

PR #73, opened 2026-04-15
fix(scripts): improve install reliability for fresh server deployment

mando222:
## Summary
- **Fix inline comments in `env.example`** that broke `source .env` — daemon crashed with `ValueError` when parsing values like `"300"  # 5 minutes`
- **Add `pg_trgm` extension** to `database/init/01_init_schema.sql` — backend reported `database_available: false` because GIN index creation failed
- **Fix `.env` parsing in `scripts/start_database.sh`** — replaced fragile `export $(cat .env | xargs)` with `set -a; source .env; set +a`
- **Add `BIND_HOST` env var** to `start_daemon.sh` and `start_web.sh` — defaults to `127.0.0.1`, set to `0.0.0.0` for remote/VM deployments
- **Add prerequisite checks** (Docker, Node.js 18+, npm, python3-venv) to `setup_dev.sh`, `start_daemon.sh`, and `start_web.sh`

## Test plan
- [ ] Run `bash -n` syntax check on all modified scripts
- [ ] `source env.example` in a clean shell — verify no parse errors
- [ ] Fresh clone + `./setup_dev.sh` on a machine missing Docker — verify clear error message
- [ ] `BIND_HOST=0.0.0.0 ./start_daemon.sh` — verify backend listens on all interfaces
- [ ] Fresh PostgreSQL container picks up `pg_trgm` from init SQL


G-r-ay (approved):
There would be concerns about the addition of the pg_trgm extension without migration, but I think that's maybe overkill right now

mando222:
@claude — addressed all three issues you flagged in commit edd45e5:

1. **`setup_dev.sh` wrong filename** — changed `cp .env.example .env` → `cp env.example .env`. The repo file has no leading dot, so the `cp` failed under `set -e` on every fresh clone.

2. **`BIND_HOST` silently overwritten by `source .env`** — both `start_daemon.sh` and `start_web.sh` now save the caller-supplied `BIND_HOST` before sourcing `.env` and restore it afterward. `BIND_HOST=0.0.0.0 ./start_daemon.sh` now actually binds to all interfaces as documented.

3. **Node.js warning doesn't suppress frontend startup** — introduced `SKIP_FRONTEND` flag in both `start_daemon.sh` and `start_web.sh`. When node is absent or <v18, the flag is set to `1` and both the npm-install block and frontend-start block are guarded by it, so the script no longer contradicts its own warning.

All three scripts pass `bash -n` syntax checks; BIND_HOST override/default logic verified in isolated shell tests.

PR #160, opened 2026-04-27
Remove spaces around equals signs

glennbarrett:
The spaces around the equals signs in the example .env file were causing the source command to fail with "Command Not Found" for all the variable names, since it thought they were commands.


mando222 (approved):

mando222:
Thank you so much for the PR! It has been merged

PR #180, opened 2026-04-30
attack.py (added in the recent ATT&CK tab PR) imports get_time_range …

johnvanlowe:
…from backend.api.analytics, but since api/__init__.py imports both modules during startup, it creates a cycle — attack tries to import from analytics before analytics is fully initialized.

### DIFF
--- a/backend/api/attack.py
+++ b/backend/api/attack.py
@@ -1,13 +1,25 @@
 """ATT&CK framework API endpoints."""
 
-from datetime import datetime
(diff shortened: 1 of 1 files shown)


joshuacox:
fixes #181

johnvanlowe:
seems to no longer be needed as claude/hotfix-attack-circular-import in PR #179 addressed same issue

mando222:
Yeah the #179 hotfix should take care of this.  Closing.

PR #314, opened 2026-06-02
V0.2.0

hubert-s:
First set of tweaks/bugfixes needed to get Vigil working in my environment.

Summary:
- when running one of the skills, i noticed that IOCs were not getting attached to the case.  This required an additional change in mcp-servers/servers/deeptempo_findings.py
- when running in production mode without vite, the /assets were not getting served
- we were not able to get bifrost working with litellm, so we are using your native claude connector.  our llm proxy is very strict with additional keyword, so we needed to sanitize the payload.

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -156,6 +156,13 @@ app = FastAPI(
     version=__version__,
 )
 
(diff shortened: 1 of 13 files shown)


johnvanlowe (approved):
LGTM

PR #336, opened 2026-06-10
fix: provider delete failures, single-default enforcement, cascade cleanup

### LINKED ISSUE #329: fix: AI provider management — delete failures, single-default enforcement, and cascade cleanup
## What to build

Three related bugs in the AI provider settings surface that should be fixed together:

**1. Delete fails on the top provider row**
The delete action errors for at least the first provider in the list. Likely a FK constraint violation or a missing guard in the API handler. The error surface and root cause need investigation.

**2. Default toggle has no UX feedback and allows invalid state**
- The UI doesn't explain why deselect/select is blocked when interacting with the default toggle, leaving users confused.
- No DB-level or API-level constraint prevents multiple rows having `is_default = true` simultaneously. Setting a new default should atomically unset the previous one.

**3. Deleting a provider leaves dangling references**
When a provider is deleted, `provider_id` references in `ai_model_configs`, agent configs, and any other FK'd tables are not cleaned up. This silently breaks model resolution and agent routing for any component that was pinned to the deleted provider.

## Acceptance criteria

- [ ] Deleting any provider row (including the first/top one) succeeds without error
- [ ] Only one provider can have `is_default = true` at a time — setting a new default atomically clears the previous one
- [ ] The UI shows a clear reason when a default toggle action is blocked (e.g. "At least one default provider is required")
- [ ] Deleting a provider nullifies or removes all references to its `provider_id` in `ai_model_configs` and related tables
- [ ] Deleting the current default provider either blocks the action with a clear message or promotes another active provider to default automatically

## Blocked by

None — can start immediately

mattmorr1:
Closes #329

## Summary

- **Delete no longer fails with FK violation** — `ai_model_configs` rows referencing the deleted provider are removed before the provider row is deleted (the FK is `ON DELETE RESTRICT` so this must happen first)
- **Deleting the only active default is blocked** — returns 409 with a clear message; if another active provider exists the next one is auto-promoted to default (with a `db.flush()` before promotion to satisfy the unique partial index)
- **Unsetting `is_default` via PUT is guarded** — returns 409 if no other default exists for that provider type
- **Default star icon is now `disabled`** with a tooltip explaining "to change, set another provider as default first" instead of silently ignoring the click

## Test plan

- [x] `DELETE /api/llm/providers/<only-active-default>` → 409 with message
- [x] `DELETE /api/llm/providers/<default-with-other-active>` → 200, other provider auto-promoted to default, `ai_model_configs` assignments cleared
- [x] `PUT /api/llm/providers/<only-default>` with `{"is_default": false}` → 409
- [x] Default star in UI is disabled; tooltip explains how to change


frontend/src/components/settings/LLMProvidersTab.tsx
```diff
@@ -71,11 +71,27 @@ export default function LLMProvidersTab({ setMessage }: Props) {
     }
   }
 
-  const handleDelete = async (id: string) => {
-    if (!window.confirm(`Delete provider "${id}"? This also removes its stored API key.`)) return
+  const handleDelete = async (p: LLMProvider) => {
```
nestor-deeptempo, line 74: update UI changes to new redesign path

mattmorr1: remediated with rework in #357 , but shouldn't cause any breakage even if merged seperately.

backend/api/llm_providers.py
```diff
@@ -312,12 +331,41 @@ async def delete_provider(
...
+    # deleting — the FK is ON DELETE RESTRICT so this must happen first.
+    db.execute(sa_delete(AIModelConfig).where(AIModelConfig.provider_id == provider_id))
+
     if row.api_key_ref:
         try:
             delete_secret(row.api_key_ref)
         except Exception as exc:  # noqa: BLE001
             logger.warning("Failed to delete secret %s: %s", row.api_key_ref, exc)
-        # Wipe the corresponding value on Bifrost too.
         push_provider_key(row.provider_type, "")
```
nestor-deeptempo, line 369: This may clear the credential for a provider that's still in use. push_provider_key is keyed by provider_type, and Bifrost stores only one key per type (bifrost_admin.py:56). So when we delete a provider that has a same-type sibling — including the one promoted on line 358 — blanking the key here also removes the credential that sibling relies on. Nothing restores it afterward, since _schedule_catalog_resync re-syncs the model allow-list but not the keys, leaving the entire provider type without a working key until a restart or a manual key update.

Could we limit the clearing to the case where it's the last provider of its type, and otherwise re-push the surviving provider's key?

nestor-deeptempo (approved):

PR #337, opened 2026-06-10
ux: replace inherit/chat-default checkbox with single model dropdown

### LINKED ISSUE #331: ux: redesign model selection — replace inherit/chat-default checkbox with dropdown
## What to build

The current model selection UI uses a "chat default" checkbox that greys out all other options when checked. The inherit vs chat-default distinction is hard to follow and the visual hierarchy is confusing.

Proposed replacement: a single dropdown per component/agent that shows the currently active model. When using the provider's default, the dropdown label reads "Default (model-name)" so the actual model is always visible. An explicit override option lets users pin a specific model without needing to understand the inherit/chat-default concept.

This removes the checkbox entirely and makes the selection self-explanatory: what you see in the dropdown is what gets used.

## Acceptance criteria

- [ ] Model selection for each component/agent is a single dropdown — no separate "chat default" checkbox
- [ ] Dropdown shows the resolved model name when set to default (e.g. "Default (llama3:latest)") so users always see what will be called
- [ ] Selecting a specific model overrides the default for that component only
- [ ] Resetting to default is a clear action (e.g. "Use provider default" option at the top of the dropdown)
- [ ] Behaviour of `ai_model_configs` resolution is unchanged — only the UI representation changes

## Blocked by

None — can start immediately

mattmorr1:
Closes #331

## Summary

- Removed the separate **Provider** dropdown column and **Inherit** checkbox column from the Model Assignment table
- Each component row now has a **single combined dropdown** — the first option is `Use Chat Default (model-name)` making the fallback chain self-explanatory without needing to understand "inherit"
- Models are grouped by provider using `ListSubheader` inside a single `<Select>`
- The resolved model name (including provider default) is always visible in the label — no more invisible/greyed-out state
- Capabilities chips (Tools, Thinking, Vision) now also show when inheriting, using the resolved chat_default model's capabilities
- `ai_model_configs` resolution logic in `services/model_registry.py` is unchanged — only the UI representation changed

## Test plan

- [x] Table renders 3 columns: Component / Model / Capabilities
- [x] Non-`chat_default` rows show "Use Chat Default (llama3.2:1b)" as the first dropdown option
- [x] Selecting a model pins that component; selecting "Use Chat Default" clears the assignment
- [x] Models are grouped by provider with a subheader in the dropdown
- [x] Build passes cleanly (`npm run build`)
- [x] Pre-existing test failure in `LLMProvidersTab.test.tsx` confirmed unrelated to this change (also fails on `main` with no changes)

### DIFF
--- a/frontend/src/components/settings/ModelAssignmentTab.tsx
+++ b/frontend/src/components/settings/ModelAssignmentTab.tsx
@@ -109,7 +109,7 @@ export default function ModelAssignmentTab({ setMessage }: Props) {
   const chatDefaultModel = useMemo(() => {
     const a = initialAssignments[CHAT_DEFAULT_KEY]
     if (!a) return null
-    return models.find((m) => m.model_id === a.model_id) ?? null
+    return models.find((m) => m.provider_id === a.provider_id && m.model_id === a.model_id) ?? null
   }, [initialAssignments, models])
 
   const chatDefaultLabel = chatDefaultModel
@@ -219,7 +219,8 @@ export default function ModelAssignmentTab({ setMessage }: Props) {
       </Typography>
       <Typography variant="caption" color="text.secondary" sx={{ display: 'block', mb: 2 }}>
         Pick a model for each system component. Unassigned rows fall back to the{' '}
-        <code>Chat Default</code>. Model list is live-queried from each provider.
+        <code>Chat Default</code>, then to the active provider's default model if Chat Default is
+        also unset. Model list is live-queried from each provider.
       </Typography>
 
       {providerIds.length === 0 ? (
@@ -286,6 +287,10 @@ export default function ModelAssignmentTab({ setMessage }: Props) {
                             {chatDefaultLabel}
                           </Typography>
                         </MenuItem>
+                      )}
+                      {/* Hidden sentinel so MUI doesn't warn when chat_default has no assignment */}
+                      {isChatDefault && (
+                        <MenuItem value={INHERIT_VALUE} sx={{ display: 'none' }} />
                       )}
 
                       {/* Models grouped by provider */}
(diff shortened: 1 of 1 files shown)


mattmorr1:
@nestor-deeptempo ready for review :)

frontend/src/components/settings/ModelAssignmentTab.tsx
```diff
@@ -229,39 +218,45 @@ export default function ModelAssignmentTab({ setMessage }: Props) {
...
               const meta = COMPONENT_LABELS[c] || { label: c, description: '' }
-              const row = rows[c] || { inherit: true, providerId: '', modelId: '' }
+              const row = rows[c] || { providerId: '', modelId: '' }
               const isChatDefault = c === CHAT_DEFAULT_KEY
-              const providerModels = row.providerId
-                ? modelsByProvider[row.providerId] || []
-                : []
-              const selectedModel =
-                providerModels.find((m) => m.model_id === row.modelId) || null
+              const dropdownValue = encode(row.providerId, row.modelId)
```
nestor-deeptempo, line 246: One edge case on dropdownValue: if a component is pinned to a model that's dropped out of /ai/models (provider disabled, or an Ollama model removed locally), encode() returns a pid::mid that matches no MenuItem, so the <Select> renders blank (+ MUI out-of-range warning) and the capability chips disappear. The row looks unassigned even though it's still pinned in the DB, so the user can't see what it's set to.

Could we render a disabled MenuItem for an unrecognized value, e.g. Pinned: {modelId} (unavailable), so the stale assignment stays visible?

nestor-deeptempo (changes requested):

mattmorr1:
No need since redesign remediates these issues

PR #341, opened 2026-06-11
Ollama fixes

mattmorr1:
feat(ollama): Non-Anthropic chat routing + centralised DEFAULT_MODEL

 **Summary**

  - Sliding window context compression — replaces blocking LLM-based auto-summarisation with a pure rolling window. SOC entities (finding IDs, case IDs, IPs, hashes, CVEs) are
  extracted via regex and folded into a running summary string in a daemon thread, so no Anthropic key is needed at request time and latency is zero.
  - Ollama/OpenAI chat routing — chat_stream now detects the active provider before touching ClaudeService. Non-Anthropic providers route through LLMRouter.dispatch_stream()
  via Bifrost's OpenAI-compatible surface; the Anthropic SDK is never instantiated when Ollama or OpenAI is configured.
  - Centralised DEFAULT_MODEL constant — introduces services/defaults.py with DEFAULT_MODEL = os.getenv("DEFAULT_MODEL", "claude-sonnet-4-6") and replaces all 41 hardcoded
  model-string fallbacks across 11 files. Operators on Ollama-only deployments can set DEFAULT_MODEL=llama3.2:1b so the failsafe never tries to call an Anthropic endpoint.
  - Bifrost Ollama provider config — adds the ollama provider block to docker/bifrost/config.json.

  Test plan

  - [x] Start with Ollama as the only configured provider — confirm chat_stream returns text chunks without an Anthropic key
  - [x] Start with Anthropic as the provider — confirm existing chat behaviour is unchanged
  - [x] Set DEFAULT_MODEL=llama3.2:1b in env, restart backend, confirm no claude-sonnet-* strings appear in LLM requests
  - [x] Confirm long conversatcions roll off the context window without triggering a summarisation LLM call
  - [x] Confirm GET /api/health and backend startup succeed with no Anthropic key set

### DIFF
--- a/services/chat/context_manager.py
+++ b/services/chat/context_manager.py
@@ -1,0 +1,447 @@
+"""Context lifecycle management.
+
+Responsibilities:
+- Token estimation (char-based heuristic)
+- History windowing (sliding window, no LLM call)
+- Rolling summary compression (fold aged-out messages into a summary string
+  asynchronously; zero LLM calls at request time, provider-agnostic)
+- Prompt-cache control injection (Anthropic cache_control blocks)
+- Tool filtering and per-tool response budgets
+"""
+from __future__ import annotations
+
+import json
+import logging
+import re
+from typing import Any, Dict, List, Optional, Tuple
+
+from services.defaults import DEFAULT_MODEL
+
+logger = logging.getLogger(__name__)
+
+# Rolling summary is capped to avoid unbounded growth.
+_SUMMARY_MAX_CHARS = 8000  # ~2k tokens
+
+
+def _flatten_content_to_text(content: Any) -> str:
+    """Extract plain text from any message content shape."""
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts = []
+        for block in content:
+            if isinstance(block, str):
+                parts.append(block)
+            elif isinstance(block, dict):
+                if block.get("type") == "text":
+                    parts.append(block.get("text", ""))
+                elif block.get("type") == "tool_result":
+                    inner = block.get("content", "")
+                    parts.append(_flatten_content_to_text(inner))
+                elif block.get("type") in ("tool_use",):
+                    parts.append(json.dumps(block.get("input", {})))
+            elif hasattr(block, "text"):
+                parts.append(getattr(block, "text", ""))
+        return " ".join(p for p in parts if p)
+    return ""
+
+
+def _prepend_summary_block(summary: str, windowed: List[Dict]) -> List[Dict]:
+    if not summary:
+        return list(windowed)
+    summary_msg = {
+        "role": "user",
+        "content": f"[INVESTIGATION CONTEXT - rolling summary of earlier conversation]\n{summary}\n[END CONTEXT]",
+    }
+    # The summary is injected as a synthetic user turn. To preserve the strict
+    # user/assistant alternation the Anthropic API requires, only follow it with
+    # the synthetic acknowledgement when the windowed history resumes with a user
+    # turn. If the window was sliced so it now starts with an assistant turn,
+    # that turn already follows the summary user turn correctly — adding the ack
+    # would produce two consecutive assistant messages.
+    if windowed and windowed[0].get("role") == "assistant":
+        return [summary_msg, *windowed]
+    return [
+        summary_msg,
+        {
+            "role": "assistant",
+            "content": "Context noted. Continuing the investigation.",
+        },
+        *windowed,
+    ]
+
+
+class ContextManager:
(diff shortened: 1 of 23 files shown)


mattmorr1:

services/claude_service.py
```diff
@@ -127,17 +134,20 @@ def __init__(
         self.thinking_budget = thinking_budget
         self.use_agent_sdk = use_agent_sdk and AGENT_SDK_AVAILABLE
 
-        # Session management for multi-turn conversations
-        # L1: in-memory (fast); L2: MemPalace (durable across restarts)
-        self.sessions: Dict[str, List[Dict]] = {}
```
nestor-deeptempo, line 132: there are still a lot of calls to self.sessions in this file, like methods create_session/get_session/clear_session and spots where those methods are called. could you clean those up?

services/chat/context_manager.py
```diff
@@ -0,0 +1,447 @@
...
+        max_msgs = window * 2 if window > 0 else len(messages)
+
+        if len(messages) > max_msgs:
+            overflow = list(messages[:-max_msgs])
+            windowed = list(messages[-max_msgs:])
+        else:
+            overflow = []
+            windowed = list(messages)
+
+        prepared = _prepend_summary_block(summary, windowed)
```
nestor-deeptempo: prepare_context prepends summary but never folds the overflow it just computed. 

summary is always "" here because get_summary (session_manager.py:127) reads an in-memory dict that's empty on every per-request ClaudeService (backend/api/claude.py:327), with nothing calling _session_mgr.get() to hydrate it from disk. So aged-out messages are dropped, not compressed. Since there's an overflow in hand, fold it synchronously and prepend it this request:

could do this to address this

folded = self.fold_overflow(overflow, summary) if overflow else summary
prepared = _prepend_summary_block(folded, windowed)

services/claude_service.py
```diff
@@ -3197,17 +2798,14 @@ async def chat_stream(
...
+            # Sliding window + rolling summary compression (no LLM call).
+            messages, windowed_out = await self._prepare_context_async(
+                messages, effective_system_prompt, model=model, session_id=session_id
             )
-            if summarized_count > 0:
+            if windowed_out > 0:
                 yield {
-                    "type": "context_summarized",
-                    "summarized_messages": summarized_count,
+                    "type": "context_windowed",
```
nestor-deeptempo, line 2797: This renames the event to context_windowed/windowed_messages, but the three consumers still listen for the old context_summarized/summarized_messages: backend/api/claude.py:799, Chat.tsx:498, and ClaudeDrawer.tsx:603. So the "context compressed" banner never renders and the backend log line is dead. Either keep the old event name here, or update all three.

services/claude_service.py
```diff
@@ -1289,570 +1331,133 @@ def _strip_thinking_blocks(self, messages: List[Dict]) -> List[Dict]:
...
-
-Be thorough. This summary replaces the conversation history — anything you omit is lost.
-
-CONVERSATION ({len(conversation_text)} chars):
-{conversation_text}
-
-Provide a structured summary preserving all critical context."""
+        return ContextManager.build_summary_prompt(conversation_text)
 
     def _summarize_messages_sync(
```
nestor-deeptempo: These delegate to self._context_mgr._summarize_messages_sync/_async, but ContextManager doesn't define those methods (it kept the summarization clients per the docstring at context_manager.py:78, but not the methods). Any call raises AttributeError. Nothing calls them right now, but they're dead shims that throw. Either drop both methods (_summaritze_messages_sync/_async) or port the actual summarization logic onto ContextManager.

services/llm_router.py
```diff
@@ -503,6 +503,12 @@ async def dispatch_openai_stream(
                     yield {"type": "text", "content": content}
         except Exception as exc:
             await _wrap_budget_errors(_raise_async(exc))
+        finally:
+            # AsyncOpenAI holds an httpx connection pool; close it on normal
+            # completion, error, AND consumer disconnect (GeneratorExit, e.g.
+            # the SSE client goes away mid-stream) so file descriptors /
+            # connections don't accumulate under load.
+            await client.close()
```
nestor-deeptempo, line 517: 👍 on closing the client here. The same finally: await client.close() is missing in the two non-streaming dispatchers that dispatch() uses: _dispatch_bifrost_openai (client at :397, create at :413) and _dispatch_anthropic (:557). chat() now routes the non-streaming path through dispatch() → _dispatch_bifrost_openai (backend/api/claude.py:407), so each call leaks an httpx pool until GC. Wrap those create() calls in try/finally: await client.close() too

nestor-deeptempo (changes requested):

nestor-deeptempo (approved):
lg 👍

PR #343, opened 2026-06-12
fix(db): URL-encode Postgres credentials (#306)

### LINKED ISSUE #306: [Bug] PostgreSQL connection fails when password contains complex/special characters
Users are unable to successfully establish a PostgreSQL database connection if their password contains certain complex or special characters (e.g., `@`, `#`, `?`, `/`, `%`, `&`). 

It appears the application may be failing to properly escape or URL-encode these characters when parsing the database connection string, resulting in authentication failures or malformed URI errors.

The value needs to be set during the initial configuration from start scripts.

G-r-ay:
Fixes #306

### DIFF
--- a/database/connection.py
+++ b/database/connection.py
@@ -7,6 +7,7 @@ Handles database connections, session management, and connection pooling.
 import os
 import logging
 from typing import Optional, Generator, TYPE_CHECKING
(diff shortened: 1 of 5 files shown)


nestor-deeptempo:
good work 🫡

PR #351, opened 2026-06-19
[codex] fix entity graph node click stability

ShmalexM:
## Summary
- normalize entity graph link endpoints after react-force-graph mutates source/target ids into node objects
- reuse that normalization for link highlighting and painting
- add a focused unit test for mutated node endpoints

## Root Cause
react-force-graph mutates link source/target fields from string ids into node objects during simulation. A click-triggered re-render recalculated filteredLinks with String(link.source), which produced [object Object] and dropped links from the graph.

## Validation
- npm run test:run -- EntityGraph.test.ts
- ./node_modules/.bin/eslint src/components/graph/EntityGraph.tsx src/components/graph/entityGraphUtils.ts src/components/graph/EntityGraph.test.ts --ext ts,tsx --report-unused-disable-directives
- npm run build

Note: full npm run lint still fails on existing unrelated lint errors outside this change.


nestor-deeptempo:
only addresses the graph on the old ui, but merging in case we use the old one as a foundation for a new one

PR #357, opened 2026-06-24
feat(llm): native OpenAI-format agent loop for Ollama/OpenAI providers

mattmorr1:
Agentic OLLAMA + OPENAI SUPPORT

KEY FEATURES: 

-> `OpenAIAgentService` — agent loop for OpenAI-format providers
New `services/openai_agent_service.py`: streaming multi-turn tool calling spoken in native OpenAI chat-completions wire format end-to-end. Tool execution reuses the same MCP + backend dispatch as `ClaudeService`, so Ollama/OpenAI agents get the full Vigil toolset. Structurally parallel to the Anthropic loop, with matching guardrails:
- Multi-turn tool loop (max 30 iterations) with a 300s wall-clock timeout
- Exponential backoff between late iterations
- Infinite-loop detection (identical tool signatures ×3)
- Per-agent `recommended_tools` filtering
- Context-window management (history window + token-based trimming)
- Per-iteration `LLMInteractionLog` logging + Bifrost correlation UUID
- Tool-result truncation + prompt-injection wrapping
- Budget enforcement via VK headers
- `anthropic_tools_to_openai()` schema converter

-> Chat routing wiring (`backend/api/claude.py`)
The `/chat/stream` path now branches on provider type:
- **Non-Anthropic + tool-capable model** → `OpenAIAgentService.stream()` (full agent loop)
- **Non-Anthropic + no tool support** → `LLMRouter.dispatch_openai_stream()` with a context-only, no-fabrication system prompt
- **Anthropic** → `ClaudeService.chat_stream()` (unchanged path)

Tool capability is resolved per-model via `ModelRegistry`, including **Ollama capability inference** (tool-use + vision detection) in `services/provider_model_discovery.py` / `services/model_registry.py`.

-> ClaudeService.chat_stream parity
Brought the Anthropic loop up to the same feature set: prompt caching, refusal handling, an explicit max-iteration message, and `recommended_tools` filtering.

Also updated wording for vigil login + added profile icon for username/email signing box.

### DIFF
--- a/services/claude_service.py
+++ b/services/claude_service.py
@@ -2598,7 +2598,6 @@ Your goal is to help SOC analysts work more efficiently by leveraging all availa
             content_blocks = []
             # Add images first
             content_blocks.extend(images)
-            # Add text message
             content_blocks.append({"type": "text", "text": message})
             return content_blocks
 
--- a/services/llm_format.py
+++ b/services/llm_format.py
@@ -1,0 +1,196 @@
+"""Anthropic <-> OpenAI wire-format translation.
+
+The interactive chat path (``OpenAIAgentService``), the daemon's autonomous
+tool loop (``daemon/agent_runner`` via ``LLMRouter``), and the workflow engine
+all build conversations in **Anthropic** shape — messages whose ``content`` is a
+list of typed blocks (``text``, ``thinking``, ``tool_use``, ``tool_result``)
+and tools described as ``{"name", "description", "input_schema"}``.
+
+Non-Anthropic providers (Ollama, OpenAI, Groq) are reached through Bifrost's
+OpenAI-compatible ``/v1`` surface, which expects a different shape: assistant
+``tool_calls`` arrays, standalone ``role: "tool"`` result messages, and tools
+described as ``{"type": "function", "function": {...}}``.
+
+Centralizing the translation here keeps the router and the agent loops from
+drifting. Both translators are **idempotent / tolerant**: a message whose
+``content`` is already a plain string (the ordinary OpenAI shape) passes
+through untouched, so callers that never used Anthropic blocks are unaffected.
+"""
+
+from __future__ import annotations
+
+import json
+import uuid
+from typing import Any, Dict, List
+
+
+def anthropic_tools_to_openai(
+    tools: List[Dict[str, Any]],
+) -> List[Dict[str, Any]]:
+    """Convert Anthropic tool schemas to OpenAI function-calling format.
+
+    Anthropic: ``{"name", "description", "input_schema": {json-schema}}``
+    OpenAI:    ``{"type": "function", "function": {"name", "description",
+               "parameters": {json-schema}}}``
+
+    A tool that is already in OpenAI shape (has a ``function`` key) is passed
+    through unchanged so this is safe to call on mixed/pre-converted lists.
+    """
+    converted: List[Dict[str, Any]] = []
+    for tool in tools:
+        if "function" in tool and tool.get("type") == "function":
+            converted.append(tool)
+            continue
+        converted.append(
+            {
+                "type": "function",
+                "function": {
+                    "name": tool["name"],
+                    "description": tool.get("description", ""),
+                    "parameters": tool.get(
+                        "input_schema",
+                        {"type": "object", "properties": {}},
+                    ),
+                },
+            }
+        )
+    return converted
+
+
+def _flatten_tool_result_content(content: Any) -> str:
+    """Reduce an Anthropic ``tool_result`` block's content to plain text.
+
+    The content may be a string, or a list of blocks (``{"type": "text",
+    "text": ...}``). OpenAI ``tool`` messages carry a single string body.
+    """
+    if content is None:
+        return ""
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts: List[str] = []
+        for block in content:
+            if isinstance(block, dict):
+                if block.get("type") == "text":
+                    parts.append(str(block.get("text", "")))
+                else:
+                    parts.append(json.dumps(block, default=str))
+            else:
+                parts.append(str(block))
+        return "\n".join(parts)
+    return str(content)
+
+
+def _assistant_block_message(content: List[Dict[str, Any]]) -> Dict[str, Any]:
+    """Translate an Anthropic assistant block list to one OpenAI message."""
+    text_parts: List[str] = []
+    tool_calls: List[Dict[str, Any]] = []
+    for block in content:
+        if not isinstance(block, dict):
+            continue
+        btype = block.get("type")
+        if btype == "text":
+            text_parts.append(str(block.get("text", "")))
+        elif btype == "tool_use":
+            tool_calls.append(
+                {
+                    "id": block.get("id") or f"call_{uuid.uuid4().hex[:12]}",
+                    "type": "function",
+                    "function": {
+                        "name": block.get("name", ""),
+                        "arguments": json.dumps(block.get("input") or {}),
+                    },
+                }
+            )
+        # 'thinking' blocks have no OpenAI equivalent — drop them.
+    msg: Dict[str, Any] = {"role": "assistant"}
+    text = "".join(text_parts)
+    if tool_calls:
+        msg["tool_calls"] = tool_calls
+        # OpenAI requires content present (may be null) alongside tool_calls.
+        msg["content"] = text or None
+    else:
+        # An assistant turn with neither text nor tool calls is invalid; emit
+        # an empty string rather than a key-less message.
+        msg["content"] = text
+    return msg
+
+
+def _user_block_messages(content: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """Translate an Anthropic user block list to OpenAI message(s).
+
+    ``tool_result`` blocks become standalone ``role: "tool"`` messages; any
+    plain ``text``/``image`` content becomes a regular user message. Tool
+    results are emitted first so they immediately follow the assistant
+    ``tool_calls`` turn they answer (OpenAI ordering requirement).
+    """
+    tool_messages: List[Dict[str, Any]] = []
+    user_parts: List[Dict[str, Any]] = []
+    for block in content:
+        if not isinstance(block, dict):
+            user_parts.append({"type": "text", "text": str(block)})
+            continue
+        btype = block.get("type")
+        if btype == "tool_result":
+            tool_messages.append(
+                {
+                    "role": "tool",
+                    "tool_call_id": block.get("tool_use_id"),
+                    "content": _flatten_tool_result_content(block.get("content")),
+                }
+            )
+        elif btype == "text":
+            user_parts.append({"type": "text", "text": str(block.get("text", ""))})
+        elif btype == "image":
+            source = block.get("source", {})
+            if isinstance(source, dict) and source.get("type") == "base64":
+                media_type = source.get("media_type", "image/jpeg")
+                data = source.get("data", "")
+                user_parts.append({
+                    "type": "image_url",
+                    "image_url": {"url": f"data:{media_type};base64,{data}"}
+                })
+            else:
+                user_parts.append(block)
+        else:
+            user_parts.append({"type": "text", "text": json.dumps(block, default=str)})
+
+    out: List[Dict[str, Any]] = list(tool_messages)
+    if user_parts:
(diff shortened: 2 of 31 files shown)


services/openai_agent_service.py
```diff
@@ -0,0 +1,683 @@
...
+                    get_security_detection_tools,
+                )
+
+                security_tools = get_security_detection_tools()
+                handler = getattr(security_tools, tool_name, None)
+                if handler:
+                    result = await handler(**arguments)
+                    return _truncate(json.dumps(result, default=str)), False
+
+            if tool_name in (
```
nestor-deeptempo: I don't think these tool names match the DatabaseDataService methods. It's get_findings/get_nearest_neighbors/get_cases, and search/stats/resolution have no direct method, so getattr returns None and 6 of 11 data tools fall through to "Unknown backend tool". 

Map names → methods, or add a shared tool_manager.execute_backend_tool.

services/claude_service.py
```diff
@@ -88,11 +86,6 @@ def _resolve_summarization_model() -> str:
 
 logger = logging.getLogger(__name__)
 
-# Sub-module imports (lazy to avoid circular deps at module load)
-from services.chat.session_manager import SessionManager  # noqa: E402
```
nestor-deeptempo, line 92: This inlines ContextManager/SessionManager/ToolExecutor into claude_service, which orphans services/chat/context_manager.py / session_manager.py / tool_executor.py., so nothing imports them anymore. Let's delete these three modules in this PR so we don't leave dead code behind. 

Also test_context_manager.py currently tests the now-dead path, so it'll need to be removed or repointed at the inlined logic

services/openai_agent_service.py
```diff
@@ -0,0 +1,683 @@
...
+
+Tool execution reuses the same MCP + backend dispatch as ClaudeService.
+
+Parity with ClaudeService.chat_stream:
+  - Multi-turn tool loop (up to 30 iterations)
+  - Wall-clock timeout (300s)
+  - Exponential backoff between iterations
+  - Infinite loop detection (identical tool signatures × 3)
+  - Per-agent recommended_tools filtering
+  - Context window management (history window + token-based trimming)
```
nestor-deeptempo: Docstring at the beginning of the file claims "token-based trimming" but the history window is only applied once before the loop, so no trimming as tool results accumulate. To make sure they agree can we either trim inside the loop or drop the claim

services/openai_agent_service.py
```diff
@@ -0,0 +1,683 @@
...
+
+        oai_messages: List[Dict[str, Any]] = []
+        if system_prompt:
+            oai_messages.append({"role": "system", "content": system_prompt})
+        oai_messages.extend(messages)
+
+        # Context window management
+        oai_messages = self._apply_history_window(oai_messages, history_window)
+
+        client = AsyncOpenAI(
```
nestor-deeptempo: AsyncOpenAI(...) is created per call and never closed → httpx connection leak, either add an async with or close in finally

services/openai_agent_service.py
```diff
@@ -0,0 +1,683 @@
...
+                    model=model,
+                    request_messages=[],
+                    thinking_enabled=False,
+                    response_content=(text_content[:2000] if text_content else None),
+                    tool_calls=(
+                        [{"iteration": iteration}] if tool_calls_count > 0 else []
+                    ),
+                    tool_results=[],
+                    stop_reason=finish_reason,
+                    input_tokens=0,
```
nestor-deeptempo: log_interaction hardcodes tokens/cost to 0, so OpenAI/Groq usage is invisible in analytics

nestor-deeptempo (changes requested):

mattmorr1:
Previously the native OpenAI-format loop only existed on the interactive chat path; the daemon and workflows were hard-bound to Anthropic. This wires them in and adds one shared wire-format translation layer plus consistent tool-safety gating everywhere.

**What changed**

Shared translation — services/llm_format.py
- anthropic_tools_to_openai + anthropic_messages_to_openai convert the Anthropic shape (tool_use/tool_result blocks, input_schema) to the OpenAI shape (tool_calls/role:tool, function.parameters). Idempotent for string-content messages, so existing callers are unaffected.

Interactive chat — OpenAIAgentService
- Streaming multi-turn tool loop with the same guardrails as ClaudeService: 30-iteration cap, 300s wall-clock, exponential backoff, infinite-loop detection, sliding context window, per-iteration audit logging with real token usage + cost.
- chat_stream runs the agentic loop when the selected non-Anthropic model supports tool calling; falls back to the no-tools router stream otherwise.

Daemon (autonomous)
- llm_router's OpenAI dispatch now translates tools + multi-turn messages and normalizes OpenAI tool-call objects back to {id,name,input}. It previously forwarded Anthropic-shaped tools/messages raw — so the daemon's tool loop would have broken on turn 2 for non-Anthropic providers. This is the key correctness fix.
- daemon/config.py keeps the provider_id from resolve_model_for_component (it was being discarded); agent_runner threads it to the gateway, disables extended thinking for non-Anthropic, and prices calls against the real provider.
- The daemon's WORKDIR state machine and tier guardrails are untouched.

Workflows
- workflows_service resolves each agent's provider/model via ai_model_configs (per-agent). Anthropic keeps ClaudeService.chat; non-Anthropic uses OpenAIAgentService (new non-streaming run() helper). Approval gates between phases preserved.

Tool-safety parity
- TOOL_TIERS / get_tool_tier moved to services/tool_manager.py so the daemon, the OpenAI agent loop, and workflows share one policy.
- OpenAIAgentService now refuses forbidden tools and creates a pending approval action for requires_approval tools instead of executing them — closing the Responder-safety gap on the non-Anthropic path.


services/llm_format.py
```diff
@@ -0,0 +1,189 @@
...
+                    "tool_call_id": block.get("tool_use_id"),
+                    "content": _flatten_tool_result_content(block.get("content")),
+                }
+            )
+        elif btype == "text":
+            user_parts.append({"type": "text", "text": str(block.get("text", ""))})
+        elif btype == "image":
+            # Pass image blocks through as OpenAI multimodal parts when shaped
+            # that way; otherwise fall back to a text marker.
+            user_parts.append(block)
```
nestor-deeptempo: Confirmed bug (reproduced): vision blocks aren't converted to OpenAI shape.

The PR adds Ollama vision capability detection, but this image branch forwards the Anthropic block unchanged. Ran anthropic_messages_to_openai verbatim with an image user message — the output keeps {"type":"image","source":{…}} and emits no image_url, so /v1 (Ollama/OpenAI) silently ignores the image. Needs translation to OpenAI's {"type":"image_url","image_url":{"url":"data:<media_type>;base64,<data>"}} here (and the same gap exists on the raw stream() path).

services/provider_model_discovery.py
```diff
@@ -326,6 +327,134 @@ def _ollama_context_from_show(show_payload: Dict[str, Any]) -> int:
...
+        )
+    else:
+        supports_tools = _name_matches_family(name_lower, _OLLAMA_TOOL_CAPABLE_FAMILIES)
+        if not supports_tools:
+            for arch_family in families_in_info:
+                if arch_family in ("general", "tokenizer"):
+                    continue
+                for known in _OLLAMA_TOOL_CAPABLE_FAMILIES:
+                    normalized = known.replace("-", "").replace(".", "")
+                    if arch_family.startswith(normalized) or normalized.startswith(
```
nestor-deeptempo: **Confirmed bug (reproduced against real Ollama): arch-family heuristic marks non-tool models as tool-capable.**

The bidirectional match here — `normalized.startswith(arch_family)` — lets a short arch family like `llama` (reported by llama2 / codellama / llama3) match `llama31`, so `supports_tools=True` for models that can't do structured tool calling. They then get routed into the agent loop, where the calls fail.

Also worth noting: on this Ollama build `/api/tags` returns **empty capabilities for every model**, so the "live caps are authoritative" precedence never fires — this heuristic is the *only* signal in practice, not a rare fallback.

Suggested fix: drop the `normalized.startswith(arch_family)` direction, and let the heuristic trust a known non-tool arch (a negative) rather than only ever upgrading to `True`.

services/tool_manager.py
```diff
@@ -0,0 +1,492 @@
...
+    if tool_name in _TOOL_TIER_LOOKUP:
+        return _TOOL_TIER_LOOKUP[tool_name]
+    short = tool_name.split("_", 1)[-1] if "_" in tool_name else tool_name
+    return _TOOL_TIER_LOOKUP.get(short, "unknown")
+
+
+# --- Backend tool dispatch ------------------------------------------------
+#
+# Single source of truth for executing built-in (non-MCP) backend tools.
+# Both agent loops route here so the tool-name -> handler mapping can't drift
```
nestor-deeptempo: Only OpenAIAgentService routes through it; ClaudeService+daemon still use claude_service._execute_backend_tool (agent_runner.py:1256, 1517), so there are two dispatch tables and the docstring's "both agent loops route here" isn't true yet

Either confirm to be separate for now or consolidate

services/openai_agent_service.py
```diff
@@ -0,0 +1,785 @@
...
+                # Accumulate streamed response
+                text_buffer = ""
+                tool_calls_buffer: Dict[int, Dict[str, Any]] = {}
+                finish_reason: Optional[str] = None
+                input_tokens = 0
+                output_tokens = 0
+                iter_start = time.monotonic()
+
+                try:
+                    stream = await client.chat.completions.create(**kwargs)
```
nestor-deeptempo: calls the OpenAI client directly, bypassing LLMRouter/_pre_dispatch_sanitize (so scan_for_injection never runs), and backend tool results aren't wrap_tool_result-wrapped (:630) though _execute_mcp_tool is (:676) and ClaudeService wraps both (claude_service.py:1855, :1955).

services/openai_agent_service.py
```diff
@@ -0,0 +1,785 @@
...
+        # approval and are NOT executed until a human resolves it.
+        tier = tool_manager.get_tool_tier(tool_name)
+        if tier == "forbidden":
+            logger.warning("Blocked forbidden tool call: %s", tool_name)
+            return (
+                f"Tool '{tool_name}' is forbidden for autonomous agents and "
+                "was not executed.",
+                True,
+            )
+        if tier == "requires_approval":
```
nestor-deeptempo: requires_approval doesn't pause the run (openai_agent_service.py:564) — we create the pending action and return is_error=False, but the loop just continues. Nothing dangerous executes without approval (good), but on the workflow path the phase then reports success while the containment action sits pending and un-taken — so a "completed" run can quietly have left a required step undone. The daemon instead enters waiting_approval and halts until a human decides.

Could we either gate the run like the daemon, or at least not mark a phase/run as success when it leaves a requires_approval action pending?

nestor-deeptempo (changes requested):

services/openai_agent_service.py
mattmorr1: fixed!

mattmorr1:
@nestor-deeptempo ready for re-review — issues addressed`22ee714`. Per-thread breakdown:

1. **Backend tool name/method mismatch** — Fixed. Backend tools now dispatch through the shared `tool_manager.execute_backend_tool` with an explicit name ladder (`get_finding`→`get_finding`, `nearest_neighbors`→`get_nearest_neighbors`, `get_case`→`get_case`, plus `search_findings`/`get_findings_stats`/`add_resolution_step`/`list_findings`/`list_cases`). 
2. **AsyncOpenAI connection leak** — Fixed. The agent no longer constructs `AsyncOpenAI` directly; it routes through `router.stream_openai_raw`, which closes the client in a `finally`.
3. **Vision blocks not converted** (confirmed bug) — Fixed. `anthropic_messages_to_openai` now translates base64 `image` blocks to OpenAI `{"type":"image_url","image_url":{"url":"data:<media_type>;base64,<data>"}}`..
4. **Bypassing sanitization** — Fixed. `scan_for_injection` now runs on the OpenAI path via `_pre_dispatch_sanitize` inside `stream_openai_raw`, and `wrap_tool_result` is applied to both backend and MCP tool results.
5. **`requires_approval` doesn't pause the run** — Fixed. Approval-gated tools raise `PendingApprovalError`; the loop creates the pending action and blocks on `_await_approval`, halting on timeout/rejection so a run can't complete with a containment step silently un-taken.
6. **Two dispatch tables / docstring** -> created issue for followup: https://github.com/Vigil-SOC/vigil/issues/393

nestor-deeptempo (approved):

PR #359, opened 2026-06-25
Replace password validator with zxcvbn entropy scoring

mattmorr1:
fixed pw validation logic. previously was just basic checker for requirements + common pw .txt file. switched to zxcvbn (open source standard for entropy-based pw checking)

- Replaces basic character-class checks (letter + digit) with zxcvbn entropy-based scoring that catches dictionary words, keyboard patterns, l33t subs, repeated sequences, and date patterns
- Keeps the bundled common-password blocklist as a fast first-pass reject
- Adds `user_inputs` parameter so passwords derived from the username/email are penalized
- Configurable threshold via `AUTH_MIN_ZXCVBN_SCORE` env var (0-4 scale, default 3 = ~10^8 guesses)
- `PasswordPolicyError` now carries a `suggestions` list with actionable feedback from zxcvbn

### DIFF
--- a/backend/services/password_validator.py
+++ b/backend/services/password_validator.py
@@ -10,6 +10,12 @@ from zxcvbn import zxcvbn
 logger = logging.getLogger(__name__)
 
 _MIN_LENGTH = 12
+# bcrypt rejects inputs longer than 72 *bytes* (it raises, it does not
+# truncate), so cap here to match — otherwise a long password passes
+# validation and then 500s at hash time. A byte cap also bounds zxcvbn's
+# super-linear matching. 72 bytes == 72 chars for ASCII, comfortably above
+# NIST's "at least 64 characters" recommendation.
+_MAX_BYTES = 72
 _MIN_SCORE = 3
 _BLOCKLIST_PATH = Path(__file__).resolve().parent.parent.parent / "data" / "common_passwords.txt"
 _blocklist: Optional[frozenset] = None
@@ -19,6 +25,17 @@ class PasswordPolicyError(ValueError):
     def __init__(self, message: str, *, suggestions: Optional[List[str]] = None):
         super().__init__(message)
         self.suggestions = suggestions or []
+
+    def as_detail(self) -> str:
+        """Render message + suggestions as a single client-facing string.
+
+        Callers surface this via HTTPException(detail=...); the frontend
+        renders ``detail`` as a plain string, so suggestions are folded
+        in here rather than returned as a nested object.
+        """
+        if not self.suggestions:
+            return str(self)
+        return f"{self} ({'; '.join(self.suggestions)})"
 
 
 def _env_int(key: str, default: int) -> int:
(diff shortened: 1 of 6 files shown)


backend/services/password_validator.py
```diff
@@ -1,95 +1,90 @@
...
-    block = frozenset(b.lower() for b in blocklist) if blocklist is not None else _blocklist_cached()
+    block = frozenset(b.lower() for b in blocklist) if blocklist is not None else _get_blocklist()
     if password.lower() in block:
         raise PasswordPolicyError(
-            "Password is too common — choose something more unique"
+            "Password is too common — choose something more unique",
+            suggestions=["Avoid passwords from known breach lists."],
         )
+
+    result = zxcvbn(password, user_inputs=user_inputs or [])
```
nestor-deeptempo, line 106: Can we add a password length cap before calling zxcvbn (or max_length on the schema fields)? Since there's no cap rn super long passwords can block the entire app

backend/services/password_validator.py
```diff
@@ -1,95 +1,90 @@
-"""
```
nestor-deeptempo, line 1: Two of the new features aren't wired into the callers yet, so they're no-ops in this pr rn

user_inputs (username/email penalization):  none of the call sites (auth.py change-password / reset, users.py create-user) pass it.
PasswordPolicyError.suggestions:  the handlers do detail=str(exc), which drops the suggestions before they reach the client.

Could pull those call sites into this PR to actually exercise both?

requirements.txt
```diff
@@ -108,3 +108,6 @@ opentelemetry-instrumentation-redis>=0.46b0
 opentelemetry-instrumentation-logging>=0.46b0
 opentelemetry-semantic-conventions-ai>=0.4.0
 
+# Password strength
+zxcvbn-python>=4.4.24
```
nestor-deeptempo: just a heads up, zxcvbn-python stops at 4.4.24 so >= only ever resolves that one version. the maintained package is zxcvbn (4.5.0), same import, probably better long term to switch

backend/services/password_validator.py
```diff
@@ -1,95 +1,90 @@
...
+    """Raise PasswordPolicyError if password fails policy.
 
-    Arguments are overridable for testing; production code calls this with
-    no arguments and picks up env / bundled-blocklist defaults.
+    All kwargs are test overrides; production uses env/defaults.
+    user_inputs: terms (username, email) zxcvbn penalizes if embedded.
     """
-    limit = min_length if min_length is not None else _min_length()
+    limit = min_length if min_length is not None else _env_int("AUTH_MIN_PASSWORD_LENGTH", _MIN_LENGTH)
+    threshold = min_score if min_score is not None else _env_int("AUTH_MIN_ZXCVBN_SCORE", _MIN_SCORE)
```
nestor-deeptempo, line 80: Worth clamping to 0–4 since setting it to 5 silently rejects every password, 0 disables the check, would also match the comment in env.example

nestor-deeptempo (changes requested):

mattmorr1:
@nestor-deeptempo fixed issues

nestor-deeptempo (approved):
good work 🙇

PR #361, opened 2026-06-25
feat(auth): harden authentication for production readiness

mattmorr1:
Production auth hardening, extracted from the bundled #357 and tidied. Two commits: the hardening (cherry-picked) and a behaviour-preserving condensing pass.

-> Hardening
- Access-token TTL → 30 min, refresh → 7 days (configurable)
- MFA secrets encrypted at rest with Fernet (key derived from `JWT_SECRET_KEY`)
- 10 single-use, bcrypt-hashed MFA recovery codes
- JWT bound to client session via IP+UA fingerprint (`sfp` claim)
- Token revocation **fails closed** when Redis is unreachable
- CSRF default-on; new `cryptography` dependency; docs in `docs/PRODUCTION_SECURITY.md`

-> Condensing (behaviour-preserving)
- Hoist `base64`/`hashlib`/`cryptography` imports to module scope
- Extract `_session_fingerprint`, `_generate_recovery_codes`, `_fernet` helpers (dedup across token-gen / MFA / encrypt-decrypt)
- Collapse multi-line single-expressions; `session = session or get_db_session()`
- Drop two pre-existing unused imports; black-format the three files
- **Left as-is:** fail-closed revocation branches + the bcrypt recovery-code loop (auditability)

### DIFF
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -190,8 +190,14 @@ async def login(
             )
 
     # Generate tokens
-    access_token = AuthService.generate_jwt_token(user, "access")
-    refresh_token = AuthService.generate_jwt_token(user, "refresh")
+    _ip = request.client.host if request.client else None
+    _ua = request.headers.get("user-agent")
+    access_token = AuthService.generate_jwt_token(
+        user, "access", request_ip=_ip, user_agent=_ua
+    )
+    refresh_token = AuthService.generate_jwt_token(
+        user, "refresh", request_ip=_ip, user_agent=_ua
+    )
 
     # Extract exp claims so the cookie Max-Age matches the JWT lifetime.
     access_payload = AuthService.verify_jwt_token(access_token) or {}
@@ -348,8 +354,14 @@ async def refresh_token(
         )
 
     # Generate new tokens
-    access_token = AuthService.generate_jwt_token(user, "access")
-    refresh_token = AuthService.generate_jwt_token(user, "refresh")
+    _ip = request.client.host if request.client else None
+    _ua = request.headers.get("user-agent")
+    access_token = AuthService.generate_jwt_token(
+        user, "access", request_ip=_ip, user_agent=_ua
+    )
+    refresh_token = AuthService.generate_jwt_token(
+        user, "refresh", request_ip=_ip, user_agent=_ua
+    )
 
     access_payload = AuthService.verify_jwt_token(access_token) or {}
     refresh_payload = AuthService.verify_jwt_token(refresh_token) or {}
--- a/backend/api/auth.py
+++ b/backend/api/auth.py
@@ -7,7 +7,7 @@ Handles login, logout, token refresh, password management, and MFA.
 import logging
 import os
 from datetime import datetime
-from typing import Optional
+from typing import List, Optional
 from fastapi import APIRouter, HTTPException, Depends, Header, Request, Response, status
 from pydantic import BaseModel, EmailStr
 from sqlalchemy.orm import Session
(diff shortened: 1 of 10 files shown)


backend/api/auth.py
```diff
@@ -514,105 +532,104 @@ async def change_password(
...
     qr_uri = AuthService.get_mfa_qr_uri(current_user.user_id, session)
     if not qr_uri:
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            detail="Failed to generate QR code"
+            detail="Failed to generate QR code",
         )
-    
+
     return MFASetupResponse(secret=secret, qr_uri=qr_uri)
```
nestor-deeptempo, line 573: we generate + store recovery codes in setup_mfa but never return them, the response is only secret + qr_uri and get_mfa_recovery_codes has no caller. docs say setup returns them, so either add them to the response or wire an endpoint

database/init/06_auth_tables.sql
```diff
@@ -26,6 +26,7 @@ CREATE TABLE IF NOT EXISTS users (
     is_verified BOOLEAN NOT NULL DEFAULT FALSE,
     mfa_enabled BOOLEAN NOT NULL DEFAULT FALSE,
     mfa_secret VARCHAR(255),
+    mfa_recovery_codes JSONB NOT NULL DEFAULT '[]',
```
nestor-deeptempo, line 29: this is inside CREATE TABLE IF NOT EXISTS, so existing DBs won't get the column — but the ORM maps it, so the first User query fails with UndefinedColumn on upgrade. Can we add an ALTER TABLE ... ADD COLUMN IF NOT EXISTS that actually runs in the init path?

we know of some users so would be best to add this to make sure they don't have any disturbances

backend/services/auth_service.py
```diff
@@ -262,252 +302,314 @@ def authenticate_user(username_or_email: str, password: str, session: Optional[S
...
+        """Decrypt an MFA secret; fall back to plaintext for pre-migration rows."""
+        try:
+            return AuthService._fernet().decrypt(encrypted.encode()).decode()
+        except Exception:
+            # Fallback: secret may be stored unencrypted (pre-migration rows)
+            return encrypted
+
     @staticmethod
-    def get_mfa_qr_uri(user_id: str, session: Optional[Session] = None) -> Optional[str]:
+    def get_mfa_qr_uri(
```
nestor-deeptempo, line 485: builds the QR from the encrypted secret now, pyotp.TOTP(user.mfa_secret) gets the Fernet ciphertext. the authenticator app will derive codes from the ciphertext while verify_mfa_code decrypts first, so they'll never match and MFA can't be enabled. Needs a _decrypt_mfa_secret(user.mfa_secret) here

backend/middleware/auth.py
```diff
@@ -131,14 +131,25 @@ async def get_current_user(
...
     # Reject revoked tokens (logout / password change / role change).
-    # Fail-open if Redis is down — see token_blacklist.is_token_revoked.
     if await is_token_revoked(payload):
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
             detail="Token has been revoked",
             headers={"WWW-Authenticate": "Bearer"},
         )
 
+    # Validate session fingerprint (IP + UA binding).
```
nestor-deeptempo: the fingerprint includes client IP, so any IP change within the token lifetime (mobile/NAT/VPN) → 401 and forced re-login. And behind a reverse proxy it's the same IP for everyone, so it adds nothing. maybe bind to UA only / a trusted forwarded-for header?

nestor-deeptempo (changes requested):

nestor-deeptempo (approved):

PR #362, opened 2026-06-25
chore(scripts): extract scripts/lib.sh, slim start scripts, move test runner

mattmorr1:
## Summary
cleaned up verbosity in start scripts. 

- New `scripts/lib.sh` with shared helpers (docker-compose wrapper, Python 3.10+ discovery, venv/deps, wait-for-url/postgres).
- Slim `start.sh` / `setup_dev.sh` / `shutdown_all.sh` to source `lib.sh`.
- Drop standalone `start_database.sh`, `stop_database.sh`, `start_daemon.sh`, `start_web.sh`.
- Delete stale `scripts/run_tests.sh` (zero-assert curl smoke script); move the pytest+vitest runner to `tests/run-tests.sh` (sourcing `lib.sh`); update `CONTRIBUTING.md` and `docs/TESTING_GUIDE.md`.
- ./start_web.sh still works (deprecate soon).


start.sh
```diff
@@ -1,513 +1,142 @@
...
-if [ "$SKIP_FRONTEND" -eq 0 ] && [ -d "frontend" ] && [ -f "frontend/package.json" ]; then
-    if [ ! -d "frontend/node_modules" ]; then
-        echo "Installing frontend dependencies (may take a few minutes)..."
-        cd frontend
-        npm install
-        cd ..
-        echo "✓ Frontend dependencies installed"
-    else
-        echo "✓ Frontend dependencies OK"
+start_frontend() {
```
nestor-deeptempo, line 78: FRONTEND_PID=$(start_frontend) hangs, backgrounded npm run dev keeps the command-substitution pipe open so it never returns. maybe use (cd frontend && npm run dev) & FRONTEND_PID=$! instead

scripts/lib.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
+# scripts/lib.sh — shared helpers for Vigil scripts. Source this, don't execute.
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+# --- Docker Compose wrapper ---
+dc() { docker compose -f "$REPO_ROOT/docker/docker-compose.yml" "$@"; }
```
nestor-deeptempo: hardcodes docker compose (v2); the old scripts fell back to docker-compose v1. v1 only hosts break on every dc call

start.sh
```diff
@@ -1,513 +1,142 @@
...
-    echo "🔐 Default Login Credentials:"
-    echo "   Username: admin"
-    echo "   Password: admin123"
-    echo "   (⚠️  Change in production!)"
-    if [ "$DEV_MODE" == "true" ]; then
-        echo ""
-        echo "⚠️  DEV_MODE ENABLED - Auth bypassed!"
+    # Open browser once frontend is ready
+    if [ "$SKIP_FRONTEND" -eq 0 ]; then
+        (sleep 3 && open "http://localhost:6988/redesign" 2>/dev/null || xdg-open "http://localhost:6988/redesign" 2>/dev/null) &
```
nestor-deeptempo: /redesign is now just a redirect to / (App.tsx), open http://localhost:6988/ instead

shutdown_all.sh
```diff
@@ -1,202 +1,48 @@
...
 
-# Check Docker status
-if command -v docker &> /dev/null && [ -d "docker" ]; then
-    echo ""
-    if [ "$DOCKER_SHUTDOWN" == "true" ]; then
-        echo "Docker status (after shutdown):"
+# Docker
+if [ "$DOCKER_STOP" -eq 1 ]; then
+    if [ "$FULL" -eq 1 ]; then
+        dc down -v
```
nestor-deeptempo: unguarded under the inherited set -e, so shutdown_all.sh -d aborts before the summary if Docker isn't available. old code guarded with command -v docker.

scripts/lib.sh
```diff
@@ -0,0 +1,98 @@
...
+        curl -sf --max-time 2 "$url" >/dev/null 2>&1 && return 0
+        sleep 1; i=$((i + 1))
+    done
+    return 1
+}
+
+# --- Ensure a docker service is running ---
+ensure_container() {
+    local name="$1" service="$2"
+    if docker ps --format '{{.Names}}' | grep -q "$name"; then
```
nestor-deeptempo: unanchored substring match, a deeptempo-postgres-test container would make us skip starting real postgres. use docker ps -q -f "name=^${name}$"

scripts/lib.sh
```diff
@@ -0,0 +1,98 @@
...
+        set -a; source "$REPO_ROOT/.env"; set +a
+    fi
+}
+
+# --- Ensure venv exists and is activated ---
+ensure_venv() {
+    local py="${1:-$(find_python)}"
+    if [ ! -d "$REPO_ROOT/venv" ]; then
+        echo "Creating virtual environment..."
+        "$py" -m venv "$REPO_ROOT/venv"
```
nestor-deeptempo: aborts with a raw error under set -e if python3-venv isn't installed. we lost the old "apt install python3-venv" hint, more of a nit..

start.sh
```diff
@@ -1,513 +1,142 @@
...
-    echo ""
-    echo "Starting backend server..."
-    nohup uvicorn backend.main:app \
-        --host "$BIND_HOST" \
-        --port 6987 \
-        --reload \
-        --reload-dir backend \
-        --reload-dir services \
-        --reload-dir database \
+    nohup uvicorn backend.main:app --host "$BIND_HOST" --port 6987 --reload \
```
nestor-deeptempo, line 123: no liveness check here anymore, daemon mode prints "Ready" even if uvicorn died on startup. can we re-add a quick ps -p/health check before the banner

nestor-deeptempo (changes requested):
deleting start_database.sh leaves dangling refs that still point users at it: backend/main.py:712, backend/api/storage_status.py:74 & :196, and the active UI frontend/src/redesign/screens/settings/DeveloperSection.tsx:105

mattmorr1:
@nestor-deeptempo fixed :)

nestor-deeptempo (approved):
lg 🫡

PR #372, opened 2026-06-29
feat(theme): add Background tweak deriving full ramp from base

G-r-ay:
## Summary
Adds a **Background** tweak on Settings → Appearance that works like the existing accent tweak, but recolors the entire surface/line/text ramp from a single base color. Presets (slate/ink/navy/espresso/light) + custom color picker + hex field, mirroring the accent control 1:1. The whole ramp is derived by mixing the base toward white/black, with dark vs. light chosen by relative luminance. Per the chosen design, the background base drives light/dark mode (and stays in lockstep with the existing mode toggle). The accent tweak is untouched; the two compose independently.

### DIFF
--- a/frontend/src/redesign/shell/theme.tsx
+++ b/frontend/src/redesign/shell/theme.tsx
@@ -150,9 +150,10 @@ export function RedesignThemeProvider({ children }: { children: ReactNode }) {
       const base = BG_PRESETS[key]
       if (!base) return
       setBg({ key, base })
-      setMode(isDarkBase(base) ? 'dark' : 'light')
+      const next = isDarkBase(base) ? 'dark' : 'light'
+      if (next !== mode) setMode(next)
     },
-    [setMode],
+    [mode, setMode],
   )
 
   const setBgHex = useCallback(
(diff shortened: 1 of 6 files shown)


frontend/src/redesign/shell/theme.tsx
```diff
@@ -92,9 +145,30 @@ export function RedesignThemeProvider({ children }: { children: ReactNode }) {
     return true
   }, [])
 
+  const setBgPreset = useCallback(
+    (key: string) => {
+      const base = BG_PRESETS[key]
+      if (!base) return
+      setBg({ key, base })
+      setMode(isDarkBase(base) ? 'dark' : 'light')
```
nestor-deeptempo: Every background change saves the light/dark setting to the server, even when light/dark didn't actually change. So dragging the color picker fires dozens of save requests per second, and clicking between the dark presets (slate/ink/navy/espresso — all "dark") re-saves "dark" every time. The light/dark toggle already guards against this; the background control doesn't.

Fix — only save when it actually flips, in both setBgPreset and setBgHex:

```
const next = isDarkBase(base) ? 'dark' : 'light'
if (next !== mode) setMode(next)
```
(mode is in scope already; add it to the useCallback deps.)

nestor-deeptempo (changes requested):

frontend/src/redesign/shell/theme.tsx
nestor-deeptempo: more of a nit

PR #374, opened 2026-06-30
drop legacy v2.0 branding and wire start.sh version from release

mattmorr1:
simple aesthetic change (no conflicts)
-> VIGIL startup gives actual version #  on startup 
(good for traceability)


nestor-deeptempo (approved):
👍

PR #377, opened 2026-07-01
[codex] route service LLM calls through configured provider

ShmalexM:
## Summary

Routes service-layer LLM calls through the configured provider instead of assuming Anthropic/Claude.

- Adds `services.configured_llm` to resolve component model assignments, select the active provider, route non-Anthropic providers through Bifrost, and preserve the existing Anthropic `ClaudeService` path.
- Updates finding enrichment, workflow execution, workflow cost estimates, and custom integration generation to use the shared configured-provider helper.
- Stamps finding enrichment output with `provider_id`, `provider_type`, and `model` metadata.
- Adds focused unit tests for Ollama routing, stale Claude model fallback, Anthropic fallback, and missing-key behavior.

## Root Cause

Several service endpoints still instantiated `ClaudeService` directly and passed `DEFAULT_MODEL`, so an Ollama-only local setup could still call `claude-sonnet-*` and fail with `no keys found that support model` even when `ollama-local` was configured as the default provider.

## Validation

- `venv/bin/python -m py_compile services/configured_llm.py backend/api/findings.py services/workflows_service.py services/custom_integration_service.py tests/unit/test_configured_llm.py`
- `venv/bin/pytest tests/unit/test_configured_llm.py tests/unit/test_claude_chat_routing.py -q` -> 23 passed
- Live local smoke: `POST /api/findings/f-20260626-e6d4b34a/enrich?force_regenerate=true` returned 200 with `provider_id=ollama-local`, `provider_type=ollama`, `model=qwen3.5:latest`.

## Notes

Unrelated local VStrike doc/script changes were intentionally left out of this branch.


mattmorr1:
## Review — route service LLM calls through configured provider

Good direction: consolidating provider/model resolution and routing non-Anthropic providers through Bifrost is the right fix for the "no keys found that support model" failure on Ollama-only setups. `estimate_configured_cost`'s signature matches `cost_estimator.estimate_cost`, and the `import asyncio` removals are safe (verified the remaining refs live inside a prompt-template string). Findings:

1. **HIGH — overlaps #392.** Both PRs rewrite `get_or_generate_enrichment` in `backend/api/findings.py`, both delete the same `ClaudeService` block and `DEFAULT_MODEL` import, and both stamp `provider_id`/`provider_type`/`model`. #392 also inlines a *second* provider resolver (`get_registry().resolve_model_for_component` + `LLMRouter().dispatch`). Reconciliation: keep this PR's `configured_llm` bridge as the single routing path and fold #392's Ollama recovery/`raw_response`/`/no_think`/token-cap into the reporting caller — do not ship two resolvers. This PR should land first.

2. **MEDIUM/HIGH — non-Anthropic workflows silently drop all tools but report success.** `configured_llm.py` hardcodes `tools=None` on the non-Anthropic branch, but `workflows_service` calls it with `use_backend_tools=True, use_mcp_tools=True, recommended_tools=...`. A playbook on Ollama/OpenAI returns `success=True` with a text blob and zero tool execution — worse than the previous hard error, because it *looks* successful. At minimum stamp the run record as text-only so operators don't trust a playbook that ran nothing.

3. **MEDIUM — duplicates `backend/api/claude.py` (slop).** `resolve_configured_llm` ≈ `_select_active_provider` (claude.py:205-237); `_model_for_provider` is a near-verbatim copy of `_router_model` (claude.py:240-250); `NO_PROVIDER_CONFIGURED` duplicates `NO_PROVIDER_DETAIL` (claude.py:98-105). Hoist the stale-`claude-*`-model guard and the no-provider detail into one shared module both call sites reuse.

4. **LOW/MEDIUM — sync DB I/O on the event loop.** `resolve_configured_llm` runs synchronous registry/provider lookups (`get_db_session`) but isn't wrapped in `asyncio.to_thread`, while the Anthropic call correctly is. CLAUDE.md prohibits blocking I/O in async paths.

5. **LOW** — a registry/DB exception in `resolve_configured_llm` propagates raw, so `findings.py` (which only catches `NoConfiguredLLMProvider`) returns 500 instead of the intended 503. **LOW** — `_model_for_provider` line 167 returns `provider.default_model` directly (can be falsy) inconsistent with the guarded fallback one line above.

**Tests:** no test drives the findings endpoint through the bridge, the custom-integration non-Anthropic branch, `estimate_configured_cost`, or the workflow non-Anthropic tool-dropping path (fixtures pin an Anthropic selection).

mattmorr1:
Pushed `3e13475` addressing the verified findings from the review above:
- **async DB I/O** — `resolve_configured_llm` now runs via `asyncio.to_thread` in both `generate_configured_text` and `estimate_configured_cost`, so registry/provider lookups no longer block the event loop.
- **silent tool-drop** — added a `text_only` flag to `ConfiguredTextResult`; `workflows_service` now records a warning on the run/phase when a non-Anthropic path drops requested tools, so a playbook that executes nothing no longer reports clean success.
- **None-model guard** — the stale-`claude-*` rewrite falls back to `DEFAULT_MODEL` instead of returning a possibly-falsy `provider.default_model`.
- added two `configured_llm` tests covering the text-only flag (set/unset).

Left for the merge with #392: the duplication of `_select_active_provider`/`_router_model`/`NO_PROVIDER_DETAIL` from `claude.py` (slop finding) is best resolved by hoisting to a shared module once both provider-routing PRs land together — flagged so it isn't done twice.

mattmorr1:
closed. we can pick this up later with agent refactor next week :)

PR #378, opened 2026-07-01
[codex] improve finding investigation UI

ShmalexM:
## Summary
- add agent selection from the main Findings table and finding detail modal
- add bulk finding selection with correlated investigation prompts
- wire Findings export button to the backend export API
- make Chat honor seeded agent choices and configured AI model metadata
- replace the Entity Graph placeholder with a live graph built from findings

## Verification
- npm run build
- npx eslint src/redesign/SocConsole.tsx src/redesign/screens/dashboard/DashboardScreen.tsx src/redesign/screens/dashboard/FindingPopup.tsx src/redesign/shared/types.ts src/redesign/shell/Chat.tsx --ext ts,tsx --report-unused-disable-directives

## Notes
- Full npm run lint still fails on pre-existing unrelated lint errors outside this PR, including legacy case/detail components, VStrike iframe context, Settings, Skills, Timesketch, and CustomIntegrationBuilder.

### DIFF
--- a/frontend/src/redesign/SocConsole.tsx
+++ b/frontend/src/redesign/SocConsole.tsx
@@ -81,6 +81,7 @@ function SocConsoleInner() {
   const { mode, accent } = useSocTheme()
   const [chatOpen, setChatOpen] = useState(false)
   const [chatSeed, setChatSeed] = useState<string | null>(null)
+  const [chatSeedAgentId, setChatSeedAgentId] = useState<string | undefined>()
   const [viewFull, setViewFull] = useState(false)
   // runtime-dynamic rail membership (mirrors production NavigationRail):
   // integrations are fetched once, orchestrator status is polled every 10s. No
@@ -113,9 +114,12 @@ function SocConsoleInner() {
     })
   }, [])
 
-  const openChat = useCallback((prompt?: string) => {
+  const openChat = useCallback<ScreenProps['openChat']>((prompt, options) => {
     setChatOpen(true)
-    if (prompt) setChatSeed(prompt)
+    if (prompt) {
+      setChatSeed(prompt)
+      setChatSeedAgentId(options?.agentId)
+    }
   }, [])
   const closeChat = useCallback(() => setChatOpen(false), [])
 
@@ -232,7 +236,16 @@ function SocConsoleInner() {
         </div>
 
         {/* Vigil chat dock */}
-        <Chat open={chatOpen} onClose={closeChat} seed={chatSeed} onSeedConsumed={() => setChatSeed(null)} />
+        <Chat
+          open={chatOpen}
+          onClose={closeChat}
+          seed={chatSeed}
+          seedAgentId={chatSeedAgentId}
+          onSeedConsumed={() => {
+            setChatSeed(null)
+            setChatSeedAgentId(undefined)
+          }}
+        />
       </div>
 
       {/* floating Vigil assistant button — hidden while the chat dock is open
--- a/frontend/src/redesign/screens/dashboard/DashboardScreen.tsx
+++ b/frontend/src/redesign/screens/dashboard/DashboardScreen.tsx
@@ -7,13 +7,16 @@ import { Icon } from '../../shared/icons'
 import { Pie, Hbars } from '../../shared/charts'
 import { useFindings, useDashboardKpis } from './useFindings'
 import type { Finding } from '../../data/data'
+import { agentsApi, findingsApi } from '../../../services/api'
+import { notificationService } from '../../../services/notifications'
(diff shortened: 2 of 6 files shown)


mattmorr1:
## Review — improve finding investigation UI

Solid feature work — the export button is correctly wired with try/catch error handling, and all referenced APIs exist. This is the base of a three-PR set (#401 resizable dock, #403 linked cases both stack on this). Findings:

1. **HIGH — this branch is built on a base that predates #379** (empty states + settings deep-links, already merged to main). Current `main` has `DashboardScreen({ openChat, goSettings })`, `FindingsTab({ openChat, goSettings })` with the `goSettings('integrations')` empty-state deep-link, and `FindingPopup`'s `onConfigureAi` prop wired via `goSettings('ai-config')`. This branch's context still shows the pre-#379 signatures. Rebasing onto current main will conflict in `DashboardScreen.tsx` and `FindingPopup.tsx`, and a careless resolution silently drops the #379 deep-links. **Rebase onto current main and re-add `goSettings`/`onConfigureAi` to the new signatures before merge.**

2. **MEDIUM — model dropdown renders a blank selection.** `Chat.tsx`: live models are keyed with composite ids (`provider_id::model_id`) and the `Select` options use `value: m.id`, but `model` state defaults to the bare `savedSettings.model` (`claude-sonnet-4-6`). `selectedModel` recovers via the `modelId === model` fallback (so cost/context still compute), but `<Select value={model}>` has no matching option when there's no `chat_default` assignment — the picker shows no label. Normalise `model` to the matched composite id once models load, or match the `Select` on `modelId`.

3. **LOW — bulk-investigation count/body mismatch.** `DashboardScreen.tsx`: the prompt header says "Investigate these N selected findings" but the body enumerates only `slice(0, 25)`. Cap the header at the slice length or say "first 25 of N".

4. **LOW — seed effect ignores agent change.** The `openInvestigation` dedup key now includes the agent id, but the seed effect still guards on `seed` alone (`seedRef`). Investigating the same finding with agent A then agent B produces an identical seed, so the second seeding is dropped. Include `seedAgentId` in the `seedRef` comparison.

**Tests:** the new logic (model-id normalisation, agent picker, bulk correlation, export) relies on the `SocConsole` smoke test with a mocked `EntityGraph`; add a focused `Chat`/`FindingsTab` test for the areas above.

mattmorr1:
## Disposition after rebase attempt: substantially superseded by main

I attempted to rebase this branch onto current `main` to make it mergeable. It conflicts in 7 files, and the core conflict is not mechanical: `main` has since replaced the findings table this PR modifies. Walking through what landed after this branch forked:

- **Investigate-from-findings** — `main`'s `DashboardScreen` already wires `openChat(findingPrompt(f))` (an "Investigate with Vigil" affordance), and `FindingPopup` already takes `onConfigureAi`. This PR's per-finding investigate is already in `main`.
- **Findings table** — #379 replaced the hand-rolled `<table>` with a declarative `<DataTable>` + `ColumnPicker` + `findingsColumns` + `EmptyState`. This PR's agent-select checkboxes and row actions are built on the old `<table>`, which no longer exists.
- **Model picker** — #412/#432/#434 fixed the chat picker to list non-Anthropic providers and hide embedding models. This PR's `aiConfigApi.listModels()` reimplementation targets the same problem those already solved.
- **Resizable dock / seed plumbing** — #401 merged the resize feature into `SocConsole`/`Chat` (see also the follow-up in #438). This PR's `seedAgentId` plumbing conflicts with that and, without the agent-selection UI, has no caller.
- **Entity graph** — `main` renders the live `EntityGraph` component already.

**What remains genuinely novel:** per-agent *selection* (choosing which of the 13 agents investigates) and **bulk** correlated investigation across multiple selected findings. Those are worth having — but they need to be re-implemented on top of `main`'s `DataTable` (row-selection + a row/toolbar action column) rather than merged, because the table they were written against is gone.

### Recommendation
Close this PR and open a focused follow-up that adds, on `main`'s `DataTable`:
1. multi-row selection with a "Investigate selected (N)" toolbar action that calls `openChat` with a correlated prompt, and
2. a per-row / detail-modal agent picker that passes the chosen `agentId` through `openChat` — this PR's `openChat(prompt, { agentId })` seam (`ChatOpenOptions`, `Chat`'s `seedAgentId`) is the salvageable design here; it would need to be added to `main`'s `openChat`/`ScreenProps` (currently `(prompt?: string) => void`) and threaded to `Chat`.

I did not force the merge — resolving it would have dropped #379's DataTable/deep-links or injected dead `seedAgentId`/`onInvestigate` plumbing with no caller. Happy to build the follow-up on `main`'s DataTable if you want it.

mattmorr1:
done in #401

PR #387, opened 2026-07-09
Improve manual upload settings UX

ShmalexM:
## Summary
- rename the redesigned Settings ingestion tab to Manual Upload
- move local uploads into a first-class Manual Upload card above S3
- show supported local upload extensions and neutralize parquet ingestion copy

## Validation
- python3 -m py_compile backend/api/ingestion.py
- git diff --check
- npm run build

### DIFF
--- a/backend/api/ingestion.py
+++ b/backend/api/ingestion.py
@@ -5,7 +5,7 @@ Handles uploading and ingesting findings/cases from various file formats:
 - JSON files
 - CSV files
 - JSONL (JSON Lines) files
-- Parquet files (DeepTempo LogLM embeddings)
+- Parquet files
 - S3 sync
 """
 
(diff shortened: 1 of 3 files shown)


mattmorr1:
## Review — improve manual upload settings UX

The `.ndjson` accept filter is accurate (the backend already maps `.ndjson` → `jsonl` in `ingestion.py`), and the `ManualUploadPanel` card is a good first-class placement. One real gap:

1. **MEDIUM — the headline rename was never applied to the tab label.** The PR's stated goal is to rename the ingestion tab to "Manual Upload", and the file-header comment was updated, but `IntegrationsSection.tsx:33` still declares `['ingestion', 'Data Ingestion']`, which is what `TABS.map(...)` renders. The user-visible sub-tab still reads **"Data Ingestion"**. Change line 33's label to `'Manual Upload'`. This also matters for #388, whose new dashboard empty-state copy points users to a "Manual Upload" tab that won't exist until this rename actually lands.

Otherwise the copy/neutralized-parquet changes are fine. Pairs with #388 (disjoint files, no conflict) — land this first so #388's guidance text is accurate.

mattmorr1:
Pushed `a9ef296`: applied the tab rename the PR's title promised — `IntegrationsSection` now renders the ingestion sub-tab as **Manual Upload** (previously only the panel card had been renamed; the tab still read 'Data Ingestion'). This also makes #388's dashboard empty-state copy — which points users to 'Settings > Integrations > Manual Upload' — accurate. Frontend build passes.

PR #388, opened 2026-07-09
Add workspace data cleanup controls

ShmalexM:
## Summary
- group findings, investigations, cases, and workspace cleanup in Settings > General > Data Management
- add a backend endpoint and frontend API helper to clear cases and case-derived metrics
- improve empty states for findings, cases, entity graph, and case metrics

## Validation
- python3 -m py_compile backend/api/cases.py
- git diff --check
- npm run build


mattmorr1:
## Review — workspace data cleanup controls

The cascade logic is **correct and complete** — I verified all 18 `ForeignKey("cases.case_id")` relations in `database/models.py` are handled (NOT-NULL children deleted, `AttackLayer`/`Investigation` nulled, `case_findings` cleared, findings preserved), no orphans or `IntegrityError`. Route ordering is right (`/all` before `/{case_id}`), the transaction is atomic, and the `deleted ?? deleted_count` count fix is correct. But there's a security hole:

1. **HIGH — the bulk delete has no authorization, client or server.** #402 gated per-case delete behind `hasPermission('cases.delete')` in `CasesScreen`, and that client check is the *only* enforcement (the cases router carries auth-only `AUTH_DEPENDENCY`, no permission check). The new `DELETE /cases/all` inherits the same auth-only dependency, and the new "Clear Cases & Metrics" / "Clear Workspace Data" buttons in `GeneralSection.tsx` have **zero** permission gating. A read-only analyst who can't delete a single case can open Settings → General and wipe the entire case database. **Fix:** gate the new buttons behind `hasPermission('cases.delete')` (plus findings/investigations equivalents), and add a server-side permission dependency on `clear_all_cases` — client gating is not a security control.

2. **MEDIUM — dashboard copy references a tab that doesn't exist.** `DashboardScreen.tsx` empty state says "Settings > Integrations > Manual Upload", but that tab is still labeled "Data Ingestion" (see #387 finding — the rename was never applied). Land #387 first (with its rename fixed).

3. **LOW** — `clear_all_cases` runs ~15 synchronous ORM deletes in an async handler (blocking the event loop). Matches the existing pattern in this file, so not a regression, but a full wipe holds the block longer. **LOW** — no backend test for the destructive endpoint and no frontend test for the new `GeneralSection` branch; #402 established the pattern (`CasesScreen.test.tsx`). Add a delete-preserves-findings test and a permission-denied test once (1) lands.

Note: overlaps #379 (already merged), which added empty states/deep-links — make sure the improved empty states here don't regress that wiring.

mattmorr1:
Pushed `c0823f1` closing the authorization hole:
- **Server-side enforcement** — `DELETE /cases/all` now takes `current_user = Depends(get_current_user)` and returns **403** unless the caller has `cases.delete` (via `AuthService.check_permission`, which grants all in DEV_MODE and reads the role JSONB in production). This backs the client gate that #402 relied on for per-case delete.
- **Client gating** — each destructive button in `GeneralSection` is now hidden unless the user holds the matching permission: Clear Findings/Investigations → `findings.delete`, Clear Cases & Metrics → `cases.delete`, Clear Workspace Data → both. A read-only analyst sees an explanatory note instead.
- **Test** — added `test_cases_clear_permission.py` asserting the endpoint 403s without `cases.delete` and that it checks exactly that permission.

Backend + frontend build/tests pass; slop clean (health 57→57).

Two notes for merge: (1) this PR's dashboard empty-state copy points to 'Settings > Integrations > Manual Upload' — that tab label is fixed in #387, so land #387 first. (2) The branch predates #379 (merged empty-states/deep-links) and will need a rebase onto current main; the DashboardScreen overlap should be resolved keeping #379's `goSettings` wiring.

PR #390, opened 2026-07-11
fix(mcp): pin mcp-remote to >=0.1.16 to close CVE-2025-6514

jamessmoore:
## Summary
- Pins `mcp-remote` to `^0.1.16` for the `azure-sentinel`, `splunk`, and `jira` entries in `mcp-config.json`.
- `mcp-remote` <0.1.16 has a critical RCE (CVE-2025-6514): a malicious or compromised MCP server can return a crafted `authorization_endpoint` in its OAuth metadata, which the `open` package shells out to — full RCE via PowerShell subexpression injection on Windows, arbitrary executable execution on macOS/Linux. Fixed upstream in `mcp-remote@0.1.16`.
- These three entries invoked `npx -y mcp-remote <url>` with no version pin, relying on npx always resolving `latest` (patched) at runtime — true today, but fragile (stale npx cache, offline runs, registry issues all bypass that assumption). Given Vigil is a security product, its own MCP config referencing an unpinned package with a known critical CVE is worth closing explicitly.
- Added an inline `_security_note` next to each entry citing the CVE so the reasoning for the pin isn't silently reverted later.

## Test plan
- [x] `mcp-config.json` validated as well-formed JSON after edit
- [ ] Maintainers confirm `0.1.16`+ is compatible with the pinned Sentinel/Splunk/Jira MCP server versions in use


jamessmoore:
REF:https://jfrog.com/blog/2025-6514-critical-mcp-remote-rce-vulnerability/

nestor-deeptempo:
thanks @jamessmoore for you contribution!

PR #391, opened 2026-07-13
feat: add local Ollama recovery settings

ShmalexM:
## Summary

- add persisted AI Operations settings for host-run local Ollama recovery
- expose enable/disable, retry limit (0-3), and Bifrost restart controls in Settings > AI
- document matching environment fallbacks for unattended local development

## Safety boundary

These settings only configure the feature. The recovery implementation separately refuses to run unless `DEV_MODE=true` and `BIFROST_URL` resolves to a loopback host.

## Validation

- `/Users/alexmargaris/vigil/venv/bin/python3.12 -m pytest tests/unit/test_runtime_config.py -q`
- `git diff --check`

PR #392, opened 2026-07-13
feat: recover local Ollama enrichment

ShmalexM:
## Summary

- route finding enrichment through the configured reporting provider, including local Ollama
- preserve the full finding description, cached enrichment, and raw model output in the detail modal
- on a transient local Bifrost connection failure, keep the request open, health-check/restart the local gateway, then retry once (configurable)
- show staged progress during AI analysis; a local enrichment gets a 180-second request budget while ordinary API requests retain a 15-second fail-fast timeout
- skip optional MCP eager connections by default in `DEV_MODE`, so one broken local integration cannot prevent the API from starting

## Related configuration

Configuration controls are in #391. This implementation is independently mergeable and uses safe local defaults until those persisted controls land.

## Safety boundary

The restart path only runs when the backend is host-run with `DEV_MODE=true` and Bifrost is on a loopback address. It invokes `docker restart deeptempo-bifrost` without a shell, serializes recoveries, and does not retry cloud/API/model failures.

## Validation

- `/Users/alexmargaris/vigil/venv/bin/python3.12 -m py_compile backend/main.py backend/api/findings.py services/local_ai_recovery.py`
- `DEV_MODE=true /Users/alexmargaris/vigil/venv/bin/python3.12 -m pytest tests/unit/test_local_startup.py tests/unit/test_local_ai_recovery.py -q` (6 passed)
- `npm run build` in the local frontend (passed)
- local browser flow: regeneration stayed active past the old 15-second cutoff and completed without a manual retry
- `git diff --check`

### DIFF
--- a/backend/api/findings.py
+++ b/backend/api/findings.py
@@ -596,6 +596,11 @@ Respond ONLY with valid JSON. Be specific and actionable. Focus on helping a SOC
                 "analysis_notes": response[:1000],  # Store first 1000 chars as notes
                 "raw_response": response  # Store full response
             }
+
+        # Keep the provider's original response even when it parsed cleanly.
+        # Analysts can compare the rendered fields against the local model's
+        # exact output without having to regenerate the enrichment.
+        enrichment["raw_response"] = response
         
         # Add metadata
         enrichment['generated_at'] = datetime.utcnow().isoformat() + 'Z'
--- a/backend/api/findings.py
+++ b/backend/api/findings.py
@@ -6,7 +6,6 @@ from pydantic import BaseModel
 import logging
 
 from services.database_data_service import DatabaseDataService
-from services.defaults import DEFAULT_MODEL
 
 router = APIRouter()
 logger = logging.getLogger(__name__)
@@ -309,7 +308,8 @@ async def get_or_generate_enrichment(finding_id: str, force_regenerate: bool = Q
     
     This endpoint checks if AI enrichment already exists for the finding.
     If it exists, returns the cached enrichment immediately.
-    If not, generates new enrichment using Claude AI, caches it, and returns it.
+    If not, generates new enrichment using the configured reporting model,
+    caches it, and returns it.
     
     Args:
         finding_id: The finding ID to enrich
@@ -350,17 +350,34 @@ async def get_or_generate_enrichment(finding_id: str, force_regenerate: bool = Q
             "enrichment": existing_enrichment
         }
     
-    # Generate new enrichment using Claude
+    # Generate new enrichment using the configured reporting provider.
     try:
-        from services.claude_service import ClaudeService
-        
-        claude_service = ClaudeService(use_backend_tools=True, use_mcp_tools=False)
-        
-        # Check if API key is configured
-        if not claude_service.has_api_key():
+        from services.llm_router import LLMRouter, get_provider_spec
+        from services.model_registry import get_registry
+
+        resolved_model = get_registry().resolve_model_for_component("reporting")
+        if not resolved_model:
             from backend.api.claude import NO_PROVIDER_DETAIL
 
             raise HTTPException(status_code=503, detail=NO_PROVIDER_DETAIL)
+
+        provider_id, model_id = resolved_model
+        provider = get_provider_spec(provider_id)
+        if provider is None:
+            raise HTTPException(
+                status_code=503,
+                detail=f"Configured provider '{provider_id}' is unavailable",
+            )
+
+        claude_service = None
+        if provider.provider_type == "anthropic":
+            from services.claude_service import ClaudeService
+
+            claude_service = ClaudeService(use_backend_tools=True, use_mcp_tools=False)
+            if not claude_service.has_api_key():
+                from backend.api.claude import NO_PROVIDER_DETAIL
+
+                raise HTTPException(status_code=503, detail=NO_PROVIDER_DETAIL)
         
         # Extract finding details (use `or` to guard against keys present with None values)
         severity = finding.get('severity') or 'unknown'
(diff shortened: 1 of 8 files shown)


backend/api/findings.py
```diff
@@ -474,23 +491,79 @@ async def get_or_generate_enrichment(finding_id: str, force_regenerate: bool = Q
...
 
         # Generate enrichment
-        logger.info(f"Generating AI enrichment for {finding_id}")
+        logger.info(
+            "Generating AI enrichment for %s via %s/%s",
+            finding_id,
+            provider.provider_id,
+            model_id,
+        )
+        max_enrichment_tokens = 1400
```
mattmorr1: **Behavioral change to the existing Anthropic path, worth calling out.** This drops the enrichment budget from `max_tokens=4096` to `1400` for *all* providers — including the Claude path, which also switches from `DEFAULT_MODEL` to the registry-resolved `reporting` model. The prompt requests a detailed multi-field JSON analysis; 1400 output tokens (~1000 words) risks truncating enrichments that previously fit. If the intent is to bound local-model latency, consider keeping the prior budget on the anthropic branch and only tightening it for the local path — otherwise please justify the reduction for Claude explicitly, since it's a regression to existing behavior.

backend/api/findings.py
```diff
@@ -474,23 +491,79 @@ async def get_or_generate_enrichment(finding_id: str, force_regenerate: bool = Q
...
+            }
+            from services.local_ai_recovery import (
+                is_gateway_connection_error,
+                local_bifrost_recovery_enabled,
+                local_bifrost_recovery_retry_limit,
+                recover_local_bifrost,
+            )
+
+            retry_limit = local_bifrost_recovery_retry_limit()
+            for attempt in range(retry_limit + 1):
```
mattmorr1, line 533: This retry loop is the largest behavioral change in the PR and has no direct test. `test_local_ai_recovery.py` exercises the recovery helpers in isolation, but nothing drives the endpoint loop: the eligibility gate (`provider_type == 'ollama' and enabled and is_gateway_connection_error`), the re-raise when `recovery.ready` is false, and the exact retry count go uncovered. An async test that patches `LLMRouter.dispatch` to raise a connection error and asserts the recover-then-retry sequence would lock this in.

services/local_ai_recovery.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def is_gateway_connection_error(error: Exception) -> bool:
+    """Recognize network errors without retrying provider or model errors."""
+    if isinstance(
+        error,
+        (httpx.ConnectError, httpx.ConnectTimeout, httpx.ReadTimeout, httpx.NetworkError),
+    ):
+        return True
+    return error.__class__.__name__ in {"APIConnectionError", "APIConnectionTimeout"}
```
mattmorr1: `"APIConnectionTimeout"` isn't a real openai SDK class name — read timeouts raise `APITimeoutError` (a subclass of `APIConnectionError`), so this literal never matches. Excluding timeouts is arguably correct (a read-timeout means Bifrost is reachable but the model is slow, so a gateway restart wouldn't help), but as written it reads like an intended guard that silently does nothing. Either drop the dead string or replace it with a comment stating the exclusion is deliberate.

mattmorr1 (commented):
Traced the recovery chain end-to-end — the core feature is correctly wired. `LLMRouter().dispatch` kwargs match its keyword-only signature, the OpenAI-path return dict exposes `content`, and an `openai.APIConnectionError` passes through `_wrap_budget_errors` untouched (no budget `status_code`), so `is_gateway_connection_error` classifies it correctly. `recover_local_bifrost` re-checks health inside the `_RECOVERY_LOCK` before restarting (correct double-check), and the restart path is loopback + `DEV_MODE` gated. No correctness blockers, and the safety-boundary claims hold.

Three things worth addressing before merge — one behavioral regression and two lower-priority notes, inline.

mattmorr1:
@ShmalexM heads-up on one item from the review that needs a decision before merge — the `max_tokens` change in `backend/api/findings.py` is a **behavioral regression on the existing Claude enrichment path**, not just the new local-Ollama path.

The old code called Claude with `max_tokens=4096`. This PR replaces that with a shared `max_enrichment_tokens = 1400` used for **every** provider, and also switches the Anthropic branch from `DEFAULT_MODEL` to the registry-resolved `reporting` model. The enrichment prompt asks for a detailed multi-field JSON analysis, so 1400 output tokens (~1000 words) risks truncating enrichments that previously fit under 4096.

Two ways forward:
- If the 1400 cap is meant to bound *local-model* latency, keep 4096 on the anthropic branch and only apply the tighter budget to the local path.
- If the reduction is intentional for all providers, could you note the rationale in the PR description? As-is it silently changes existing Claude behavior.

Details on the inline thread: https://github.com/Vigil-SOC/vigil/pull/392#discussion_r3606850595 (see `findings.py` line 500).

mattmorr1:
## Review — recover local Ollama enrichment

The docker-restart/loopback safety boundary is **sound**: `_restart_bifrost` uses `create_subprocess_exec` (no shell, hardcoded container name, no user data in argv), and recovery is gated on `DEV_MODE` + a loopback `BIFROST_URL` host, so remote/prod deployments cannot trigger a restart. Good.

Verified defects to fix before merge:

1. **HIGH — global 15s axios timeout breaks every long LLM call.** `frontend/src/services/api.ts` sets `timeout: 15_000` on the shared instance. Only `getEnrichment` overrides it (180s). Every other synchronous LLM-backed POST (`cases/generate-report`, `claude/analyze-finding`, `claude/agent`, `agents/start-investigation`, `workflows/execute`, `workflows/generate`, `orchestrator/scan-findings`) now aborts at 15s with `ECONNABORTED`, and the SSE streaming endpoints inherit it too. Remove the timeout from the shared instance; add short per-request timeouts only to genuinely-fast calls.

2. **HIGH — new 503s are swallowed into 500.** The `raise HTTPException(503, ...)` paths for "no provider" / "provider unavailable" sit inside the `try` whose `except Exception as e:` re-wraps everything as 500 (`HTTPException` subclasses `Exception`). `FindingPopup.tsx` keys its "not configured" copy off a 503, so that message is unreachable — users always see the generic restore-failed error. Add `except HTTPException: raise` before the broad handler.

3. **MEDIUM — dead timeout branch in the connection classifier.** `local_ai_recovery.py` `is_gateway_connection_error` matches `__class__.__name__ in {"APIConnectionError", "APIConnectionTimeout"}`, but the OpenAI SDK has no `APIConnectionTimeout` — the real class is `APITimeoutError` (a subclass of `APIConnectionError`). A hung gateway raising `APITimeoutError` matches neither name, so recovery never fires on timeouts. Prefer `isinstance(error, openai.APIConnectionError)`.

4. **MEDIUM — `max_enrichment_tokens=1400` is applied to both providers.** That's a regression from 4096 on the Anthropic path and risks truncating the large enrichment JSON schema mid-object (which then falls back to the generic text branch). Apply the smaller budget only on the `ollama` branch.

5. **LOW** — `_wait_for_bifrost` worst case is ~135s (45 iterations × (2s health + 1s sleep)), not the advertised 45s / "up to a minute"; bound it with a monotonic deadline. `_RECOVERY_LOCK` serializes only within one process — worth a comment for multi-worker setups.

**Tests:** the retry/recovery orchestration in `findings.py`, `_restart_bifrost`, and `_wait_for_bifrost` are untested; the classifier test masks finding 3 (only covers `httpx.ConnectError`/`RuntimeError`, never the SDK-name branch). Add cases with real `openai.APIConnectionError` / `openai.APITimeoutError`.

Note: this PR's enrichment rewrite **directly overlaps #377**, which rewrites the same `get_or_generate_enrichment` to route through `services/configured_llm.py`. See that PR — the reconciliation is to route through the shared bridge and fold this PR's Ollama recovery into it, rather than ship two provider resolvers.

mattmorr1:
Pushed `d350e27` addressing the review findings:
- **axios timeouts** — the 15s bound stays on the shared instance for fast reads, but chat/analyze/agent/workflow/report/scan and enrichment POSTs now override to 180s, and the two SSE stream endpoints set `timeout: 0`. Fixes the regression where every long LLM call aborted at 15s.
- **503 masking** — `get_or_generate_enrichment` now re-raises `HTTPException` before the broad handler, so the no-provider/unavailable 503 reaches `FindingPopup`'s `not_configured` branch instead of collapsing to 500.
- **connection classifier** — recognises `openai.APITimeoutError` (via `isinstance(err, openai.APIConnectionError)`), dropping the non-existent `APIConnectionTimeout` name so a hung gateway actually triggers recovery.
- **token budget** — Anthropic enrichment restored to 4096; the tight 1400 cap now applies only to local models.
- **health wait** — `_wait_for_bifrost` uses a monotonic deadline (true ~45s ceiling), and the recovery lock's process-local scope is documented.
- **tests** — classifier test now covers the real openai timeout types (previously the dead branch was untested), plus retry-limit lower-bound/invalid-value cases.

Frontend build + backend unit tests pass. The remaining slop complexity warning on `get_or_generate_enrichment` is best resolved when this branch reconciles with #377 (both rewrite that function) — see the note on #377.

mattmorr1:
Combined with the other half of this feature into #435 (rebased onto current main, with the review fixes from both branches applied and conflicts resolved). #435 supersedes this PR — recommend merging #435 and closing this one.

mattmorr1:
Superseded by #435, which combined this PR with its other half (rebased onto main, with the review fixes applied) and has now been **merged**. The functionality here is in `main`. Closing as delivered via #435.

PR #394, opened 2026-07-17
chore: remove dead services, tools and stray files

mattmorr1:
Removes code nothing references, and fixes a live path that was already broken on the way.

## `tools/_legacy`

Thirteen modules. Twelve are referenced by nothing at all — no imports, no `mcp-config.json` entry.

The thirteenth was **live**: `tools/splunk.py` was a symlink to `_legacy/splunk.py`, and `mcp-config.json:108` launches that path as the Splunk MCP server. It is promoted to a real file at `tools/splunk.py` and `_legacy` is dropped entirely, so the configured path keeps resolving.

Its test asserted the *mechanism* (`tools/splunk.py` is a symlink into `_legacy`) rather than the *guarantee* `mcp-config.json` depends on (that path resolves), so it now checks the latter. The dotenv coverage around it is untouched.

## Also removed

- `services/cribl_service.py`, `services/investigation_workflow_service.py` — no importers
- `vigil-mcp-2026-audit.sh`, `vigil-mcp2-prep-current-state-20260612.txt` — one-off scratch output at the repo root

## Verification

- Every `tools/*.py` path in `mcp-config.json` resolves (checked all 17)
- `pytest tests/unit/test_splunk_dotenv.py` — 3 passed

Note: slop flags `handle_call_tool` complexity in `tools/splunk.py`. That is pre-existing code this PR only moves; the new path reads as new code.


mattmorr1:

PR #395, opened 2026-07-17
fix(ui): restore missing icons and clip rail labels

mattmorr1:
**The login identity icon never rendered**
-> `LoginScreen` asks for `<Icon name="name" />`, but no `name` icon exists — `Icon` does `ICON[name] || ""`, so it silently rendered an empty `<svg>`. Adds a `user` icon in the set’s existing style and points the field at it.

Why the compiler did not catch it

`ICON` was annotated `Record<string, string>`, which widens `IconName = keyof typeof ICON` to plain `string` — so **any** misspelled icon name type-checks and renders nothing. Dropping the annotation makes `IconName` a literal union, and the next typo fails the build instead of vanishing.

Swept every `<Icon name=... >` in the codebase against the icon set: this was the only broken reference (`SocConsole`’s `name={icon}` is dynamic and correctly typed).

## Rail labels clipped mid-glyph

`.rail .nav-label` sets `overflow: hidden` with no `text-overflow`, so a long account name (`Dev User (Full Admin)`) was cut rather than ellipsized.

A copy commit swapping the login brand panel to lead with the project’s values. Kept separate so the fix above stands alone.

### DIFF
--- a/frontend/src/redesign/screens/login/LoginScreen.tsx
+++ b/frontend/src/redesign/screens/login/LoginScreen.tsx
@@ -95,10 +95,10 @@ function LoginInner() {
           </div>
 
           <div className="brand-body">
-            <h1>Security operations, under one calm pane of glass.</h1>
+            <h1>Vigilant, Open, Trustworthy.</h1>
             <p>
-              Triage findings, correlate signals across your estate, and let Vigil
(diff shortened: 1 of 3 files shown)


mattmorr1:

PR #396, opened 2026-07-17
feat: seamless startup, live DB switching, adaptive tables

mattmorr1:
Prerequisite for the desktop app (#TBD), but standalone-useful on its own.

Brings the stack up without manually starting each service: `ensure_docker` starts Docker Desktop / the engine and waits for it, and `start_autostart_services` brings up the services marked autostart, so `./start.sh` works from cold rather than needing each piece up first.

Also: switch databases without taking the UI down, and render dashboard tables against schemas that do not match the expected shape.

Cherry-picked onto `main` directly — it has no dependency on the `ollama-agent` line it was originally written above.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -240,3 +240,6 @@ splunk_test_data.json
 *_test.json
 enrichment_result_*.json
 claude_analysis_*.txt
+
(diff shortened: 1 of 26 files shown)


mattmorr1:

PR #397, opened 2026-07-17
feat: Vigil desktop app + first-run bootstrap + provider-URL SSRF hardening

mattmorr1:
Single branch for the desktop work and everything it depends on (previously split across #394/#395/#396, now consolidated here at the team's request). Ten commits, oldest first:

1. **feat: seamless startup** — `ensure_docker` / `start_autostart_services` so the stack comes up cold; live DB switching; schema-adaptive dashboard tables.
2. **feat(desktop): native desktop app** — opens like an app, tears the stack down on quit. Window loads the backend origin (`127.0.0.1:6987`) which already serves the SPA, so cookies/CSRF/SSE are untouched. Two modes off one seam: a checkout drives `scripts/`+venv (dev); no checkout runs published images, Docker-only. Mac bundle ad-hoc signed (unsigned → TCC denies protected folders with no prompt).
3. **feat(auth): first-account bootstrap** — no signup + user creation needs an admin = an empty instance is unusable, and the old seed admin had a bcrypt hash matching no password. `GET/POST /api/auth/bootstrap` lets the operator create the first admin (zxcvbn-validated, advisory-locked, self-closing at 403 once a user exists); login screen shows it on an empty instance. Seed admin dropped from `06_auth_tables.sql`.
4. **feat(desktop): optional offline image** — bundle the backend image in the artifact and `docker load` it first-run, so a no-registry / offline machine still works.
5. **fix(desktop): resolve npm under the minimal GUI PATH** — `app_up.sh`/`app_setup.sh` resolved python/uvicorn by path but left `npm` bare; a GUI-spawned run hit `npm: command not found`. `ensure_npm_on_path` (beside `find_python`).
6. **fix(ui): restore missing icons** — login identity field asked for a `name` icon that does not exist (empty svg); add `user`, and drop the `Record<string,string>` annotation so a bad icon name fails the build instead of vanishing.
7. **copy: login wording** — brand panel leads with the project's values.
8. **chore: remove dead code** — 12 unused `_legacy` tools; `_legacy/splunk.py` was live (symlinked, run by `mcp-config.json`) so promoted to a real file; two dead services + two stray root files.
9. **fix(desktop): bootstrap fires in source mode too** — `app_up.sh` was seeding admin/admin123, so the account-creation screen never showed on a source run; dropped, so both paths start empty and behave identically.
10. **fix: close SSRF gaps and audit-log write** — resolves the final review findings on this PR (details below).

## Review-finding fixes (commit 10)

- **SSRF — IPv4-mapped IPv6 metadata bypass.** `url_safety._is_blocked_ip` now collapses a v4-mapped IPv6 literal (`::ffff:169.254.169.254`) to its embedded IPv4 before classification, so it re-classifies as *cloud metadata address* and stays blocked past the `allow_loopback` carve-out instead of slipping through as a plain private address.
- **SSRF — Ollama discovery skipped the IP gate.** `fetch_ollama_models` had a bespoke `allow_loopback` branch that only checked scheme/userinfo/fragment and never ran the RFC1918/metadata checks. It now routes through `validate_provider_url(..., allow_loopback=allow_loopback)` unconditionally; `allow_loopback` only opts past loopback/private, never metadata.
- **Audit row silently dropped.** `_audit_retarget` (from `POST /storage/reconnect`) wrote via a non-committing session, omitted the NOT-NULL `config_type`, and pushed f-strings into JSONB columns — so every retarget audit insert failed and was swallowed. It now delegates to `ConfigService.record_audit`, which sets `config_type="database"`, passes JSONB dicts, and commits via `session_scope`.

## Verification

- Bootstrap lifecycle against a live `DEV_MODE=false` backend: `required:true` → weak password rejected (zxcvbn) → create `201` → login `200` → `required:false` → second attempt `403`.
- Packaged app, standalone, no repo, empty volumes, **image removed to simulate a fresh machine and loaded from the bundle**: 4 containers healthy, SPA 200, create-account then sign-in works; quit removes containers, keeps data volumes.
- SPA carries all three UI fixes together (person icon, wording, account creation).
- No `admin123` hash in any commit of this branch's own changes (it remains in `scripts/init_default_user.py` on main, untouched here).
- SSRF regression suite green (21 cases), including new coverage for the IPv4-mapped metadata literal and the `allow_loopback` loopback-yes / metadata-no contract.

## Known gaps

- Standalone runs a **published image**; until one built from this branch is on GHCR, the DMG/AppImage must bundle the image (commit 4) — I build those locally.
- Untested: Linux AppImage on real hardware; macOS Finder double-click (a terminal launch inherits its TCC grant).
- The `admin123` hash in `scripts/init_default_user.py` on main is now unused by these paths and worth removing separately.
- **DNS rebinding against provider-URL validation** is a known residual: the validator resolves once but httpx re-resolves at connect time, so a short-TTL rebinding domain can still reach a blocked target. Admin-gated in practice; tracked for a pin-the-validated-IP fix in #405.


mattmorr1:
## Review pass + fixes (commits `1a6723f`..`75a95e0`)

Reviewed the branch and pushed fixes for every inline bot finding plus a CI blocker. All changes replayed onto the current branch tip after the recent `main` merge.

### Claude review-bot comments — addressed

| Comment | Resolution |
|---|---|
| `DashboardScreen.tsx` missing imports | Already imported on the current tip; `tsc --noEmit` clean. Note: CI's *Build Frontend* is `skipping`, so a real TS2304 here would not be caught by CI — worth enabling on this PR. |
| `connection.py` DSN allowlist bypass | `get_database_url` now percent-encodes host + database, so a decoded `?`/`@` can't re-open as libpq params. Verified `sslrootcert` no longer reaches the URL query; normal db names round-trip unchanged. Also fixes IPv6 host reconstruction. |
| `connection.py` `refresh_if_stale` stale creds | Added `DatabaseConfig.identity()` (host/port/db/user/password/ssl/extra_query); a password-only rotation now retargets instead of being silently skipped. |
| `seed_reference_data.py` `$$` splitting | Replaced `split(";\n")` with a single-quote / dollar-quote-aware splitter; all 15 init files split with zero orphaned PL/pgSQL fragments. |
| `service_manager.start(wait=...)` no-op | `--wait` now threaded into `docker compose up`; the flag blocks until healthy. |

### CI blocker (separate from the bot comments)

`tests/unit/test_splunk_secrets_migration.py` pointed at `tools/_legacy/splunk.py`, which this PR promoted to `tools/splunk.py` — the test errored at fixture setup and turned the backend unit job red. Repointed; suite passes.

### Additional findings from a deeper review (not all fixed)

Fixed here:
- **`retarget()` engine-swap race** — the swap ran with no lock while `refresh_if_stale` now calls it on every request; two concurrent retargets leaked the intermediate engine's pool. Now serialized under a manager-level lock. (`_RETARGET_LOCK` in `storage_status.py` only guarded `/reconnect`.)
- **Desktop quit can hang** — `before-quit` awaited `compose down` with no timeout; a wedged daemon left the app unquittable. Now raced against a 15s deadline.
- **Standalone never started Docker** — source mode calls `ensure_docker`, standalone went straight to `compose`. Added a TS `ensureDockerRunning` mirroring `scripts/lib.sh`.

Not addressed (flagging for follow-up):
- Sync `engine.connect()` on the async path during an auto-retarget (blocks the loop up to 5s; rare).
- `ollama_process.start` leaks the parent log fd; zombie not reaped after the readiness window.
- `list_services` issues redundant blocking `docker info` calls per service.
- `prepare-standalone.js` `99_drop_seed_admin.sql` is now a no-op (seed admin already removed upstream).

### Note on the ~2.5k-line deletions

These are a **refactor, not dead-code removal**: `openai_agent_service` → `LLMRouter.dispatch_openai_stream`, `tool_manager` → `claude_service`/`agent_runner`, `llm_format` → `llm_router`, `telemetry_config` → `telemetry.py`. Only `cribl_service` and `investigation_workflow_service` were genuinely unused. No dangling references remain. Two caveats worth a look before merge: behavior-equivalence of the relocated tool-safety-tier logic, and the ~640 lines of deleted `test_openai_agent_service`/`test_llm_format` coverage that `test_llm_router` only partly replaces.

nestor-deeptempo (approved):

PR #398, opened 2026-07-17
feat(loglm): page-extension host, authenticated MCP, and pgvector embeddings

nestor-deeptempo:
## Summary

Adds LogLM as a first-class connector: a page-extension host embedded in the SOC console, LogLM wired in as an authenticated MCP integration, and pgvector-backed finding embeddings — plus the trust/CSP hardening that makes cross-origin extensions safe to load.

## Highlights

- **Page-extension host** — the connector serves an ES-module bundle imported into the Vigil origin; Vigil mints short-lived session tokens server-to-server against the connector BFF (`mint_secret` HMAC).
- **Trust boundary** — a single runtime `EXTENSION_CONNECTOR_ALLOWLIST` drives the SSRF guard, the CSP `script-src`/`connect-src`, and the frontend trust gate; deprecates the build-time `VITE_EXTENSION_ORIGIN_ALLOWLIST`.
- **LogLM MCP** — static remote MCP entry with a `connectorUrl`-derived URL, bearer-authenticated; `mcp_token` and `mint_secret` are first-class wizard secrets, and the ingest webhook now requires a bearer (fail-closed).
- **Findings** — `pgvector(768)` embedding column with an HNSW ANN index (cosine distance); the daemon runs triage/enrich in a background task, decoupled from the store. Null severity is guarded (LogLM findings arrive without one).

## Database

New init file `17_loglm_setup.sql` bundles the vector extension, embedding column, and `loglm.view` RBAC grant. It's a **new filename** on purpose — Helm upgrades and fresh docker-compose both apply it, whereas edits to already-shipped `01`/`06` never reach existing deployments. Helm copy and `values.yaml` are updated in lockstep.

## Ops note

Set `EXTENSION_CONNECTOR_ALLOWLIST` to the connector origin(s) in production or extensions stay CSP-blocked (dev localhost is allowed by default).


daemon/poller.py
```diff
@@ -28,6 +29,12 @@
 logger = logging.getLogger(__name__)
 
 
+def _blocked_ingest_sources(findings: List[Dict[str, Any]], disabled_ids: Set[str]) -> Set[str]:
```
mattmorr1: i feel like this is not necessary, can call function directly?

nestor-deeptempo: good call out 👍

PR #401, opened 2026-07-22
feat(redesign): make the Vigil Assistant dock resizable

ShmalexM:
## Summary

- add pointer and keyboard resizing to the left edge of the Vigil Assistant dock
- persist the preferred desktop width while clamping the effective width to the viewport
- expose an accessible separator and a reachable, explicitly labelled close action
- disable resizing and use the full viewport on narrow screens

## Why

Analysts need more horizontal room for long investigation output without turning the assistant into a popover or covering the entire workspace. The dock now supports a 360–720px desktop range, defaults to 420px, and remembers the user's preference under soc.chat.width.v1.

## Compatibility

This change does not alter chat execution, messages, model selection, streaming, or scroll behavior. It overlaps Chat.tsx and SocConsole.tsx with #378 and should be rebased after that PR settles.

## Verification

- focused Chat and SocConsole tests — 17 passed
- focused ESLint — passed
- production frontend build — passed
- git diff --check — passed


mattmorr1:
<will commit>  -> pushing out a related change with desktop mode anyways, don't want to overstep on this PR.

mattmorr1:
## Review — make the Vigil Assistant dock resizable

Well-implemented, and the risky parts hold up under scrutiny:
- **localStorage handling is correct** — `readChatWidth` handles a missing key, empty/non-numeric values (`Number.isFinite` guard), and out-of-range widths (clamped to [360,720]); no crash paths.
- **Pointer/listener cleanup is sound** — pointer capture uses optional chaining (jsdom-safe), `finishResize` releases capture and commits the width, and the close/unmount effects reset the resize refs. No leaks.
- **Accessibility** — the separator is a valid focusable slider pattern (`role="separator"`, `aria-orientation`, live `aria-valuenow` during drag), and the narrow-screen path forces full-width, hides the handle, and preserves the stored preference. Tests cover keyboard, pointer drag, restore/persist, and the narrow-screen disable path.

Minor:
- **NIT** — `onResizeKeyDown` lacks the `maxWidth <= minWidth` guard that `onResizePointerDown` has. Harmless today (clampWidth collapses to a single value and the handle is `tabIndex=-1` on narrow screens), but the two entry points should be consistent.

Overlap note: this touches `SocConsole.tsx` and `Chat.tsx`, which #378 also modifies (the `<Chat>` props/JSX and the prop destructuring/types). Land after #378 rebases onto main; expect a small manual resolution there. Shares no files with #403.

PR #403, opened 2026-07-22
feat(cases): create linked cases from findings

ShmalexM:
## Summary
- extract redesign case creation into a reusable dialog that accepts initial findings
- preserve standalone Cases -> New Case creation while gating creation with cases.write
- create a linked case directly from finding detail, with read-only deduplicated finding IDs
- default the title, open status, and highest selected severity; keep failures non-destructive
- navigate directly to the created case detail

## Verification
- npm run test:run -- src/redesign/screens/cases/CaseCreateDialog.test.tsx src/redesign/screens/dashboard/FindingPopup.test.tsx src/redesign/screens/dashboard/DashboardScreen.test.tsx (8 passed)
- focused ESLint for all changed implementation and test files
- npm run build
- git diff --check

## Dependency note
PR #378 remains open and owns the finding-selection UI. This branch intentionally does not duplicate that toolbar. The reusable dialog already accepts and tests multiple findings, so selected-finding wiring should be added while rebasing this draft after #378 settles.

No backend route, demo-only review-case API, or AI field generation is included.

### DIFF
--- a/frontend/src/redesign/screens/cases/CaseCreateDialog.test.tsx
+++ b/frontend/src/redesign/screens/cases/CaseCreateDialog.test.tsx
@@ -1,0 +1,110 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { fireEvent, render, screen, waitFor, within } from '@testing-library/react'
+import CaseCreateDialog from './CaseCreateDialog'
+import { casesApi } from '../../../services/api'
+
+vi.mock('../../../services/api', () => ({
+  casesApi: { create: vi.fn() },
+}))
+
+const createdResponse = (caseId: string) => ({ data: { case_id: caseId } }) as never
+
+beforeEach(() => {
+  vi.mocked(casesApi.create).mockReset()
+})
+
+describe('CaseCreateDialog', () => {
+  it('creates a standalone case without linked findings', async () => {
+    vi.mocked(casesApi.create).mockResolvedValueOnce(createdResponse('case-standalone'))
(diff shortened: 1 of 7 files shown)


mattmorr1:
## Review — create linked cases from findings

Clean extraction of a reusable `CaseCreateDialog`. Error handling is good: the dialog stays open on backend failure and surfaces `response.data.detail` via `role="alert"`, guards against a missing `case_id`, and the dedup + highest-severity default logic is correct. Test coverage is thorough (standalone, dedup, single-finding title, cancel, backend-failure, permission gate, navigation encoding).

Minor:
- **LOW** — `CaseCreateDialog` does `useMemo(() => ..., [initialFindings])`, but callers pass a fresh array literal each render (`caseFinding ? [caseFinding] : []`, or the default `[]`), so the memo recomputes every render and provides no benefit. No correctness impact (`findingKey` is value-stable). Memoise the array at the call site or key the memo on a stable derived string.

Overlap note: touches `FindingPopup.tsx`, `DashboardScreen.tsx`, and `CasesScreen.tsx`. #378 also edits `FindingPopup.tsx`/`DashboardScreen.tsx` (both insert a primary button as the first child of `.fp-actions`, both extend the `FindingPopup` prop object). Land after #378 rebases onto main; expect manual resolution in those two files. Shares no files with #401. Per the PR's own note, the selected-finding wiring from #378's toolbar should be connected once #378 settles.

mattmorr1:
already merged in #401

PR #410, opened 2026-07-23
fix(db): enable pgvector extension before create_all()

### LINKED ISSUE #406: Bug: pgvector issues on start with recent llm_provider_config changes
recent commits introduced a pgvector flow that is evident on service start. happening on current origin/main deploy

sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) relation "llm_provider_configs" does not exist
LINE 2: FROM llm_provider_configs

johnvanlowe:
## Summary

- `database/connection.py`'s `create_tables()` now runs `CREATE EXTENSION IF NOT EXISTS vector` before calling `Base.metadata.create_all()`
- Fixes the startup failure introduced by #398 (loglm pgvector embeddings), which added a `VECTOR(768)` column to the `Finding` ORM model but did not update `create_tables()` to ensure the extension exists first
- The Docker init path (`17_loglm_setup.sql`) already enabled the extension for fresh container first-runs, leaving existing volumes and non-Docker startups broken on every subsequent launch

## Root cause

`Finding` has `embedding = Column(Vector(768), nullable=False)`. SQLAlchemy's `create_all()` emits `CREATE TABLE findings (... embedding VECTOR(768) ...)` before any extension setup, causing:

```
psycopg2.errors.UndefinedObject: type "vector" does not exist
```

This aborted the entire `create_all()` run, leaving the DB with only 4 of 43 expected tables and making `llm_provider_configs` missing — the symptom reported in #406.

## Test plan

- [ ] Fresh checkout with existing Postgres volume: `./start.sh` completes without the `UndefinedObject` error
- [ ] `python3 scripts/init_schema.py` exits 0 and logs "Database schema ready"
- [ ] `SELECT count(*) FROM information_schema.tables WHERE table_schema='public'` returns 45
- [ ] Re-running `init_schema.py` is idempotent (`IF NOT EXISTS` on both extension and tables)
- [ ] Deployments without pgvector log a warning and continue (non-fatal path)

Fixes #406

### DIFF
--- a/backend/api/analytics.py
+++ b/backend/api/analytics.py
@@ -985,7 +985,14 @@ class EstimateCostRequest(BaseModel):
     in the heuristic path.
     """
 
-    provider_type: str = Field(..., description="anthropic | openai | ollama")
+    provider_type: Optional[str] = Field(
+        default=None,
+        description=(
+            "anthropic | openai | ollama. Advisory only — the endpoint "
+            "resolves the real provider from model_id (registry, then name "
+            "heuristic) and uses this value only as a last-resort fallback."
+        ),
+    )
     model_id: str
     messages: List[Dict[str, Any]] = Field(default_factory=list)
     system_prompt: Optional[str] = None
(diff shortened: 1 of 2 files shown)


johnvanlowe:
close without accept flow

Issue #325, opened 2026-06-10
feat: add get_active_provider_spec() and dispatch_stream() to LLMRouter


mattmorr1:
## What to build

Two additions to `services/llm_router.py` that are prerequisites for routing chat endpoints to non-Anthropic providers:

**`get_active_provider_spec()`** — a module-level helper that queries `llm_provider_configs` for the default active provider of any type (not just Anthropic). The existing `get_provider_spec()` only looks for the default Anthropic row; this generalises it. Returns `None` if the DB is unreachable so callers can fall back gracefully.

**`LLMRouter.dispatch_stream()`** — an async generator method that streams responses from non-Anthropic providers via Bifrost's OpenAI-compatible `/v1` SSE surface. Uses `AsyncOpenAI(base_url=bifrost_url/v1)` with `stream=True` and yields `{"type": "text", "content": "..."}` dicts — the same chunk format `ClaudeService.chat_stream()` already produces, so callers need no format conversion.

## Acceptance criteria

- [ ] `get_active_provider_spec()` returns a `ProviderSpec` for an active Ollama row when one is configured as default
- [ ] `get_active_provider_spec()` returns the Anthropic `ProviderSpec` when Anthropic is the only active provider
- [ ] `get_active_provider_spec()` returns `None` when the DB is unreachable (no exception raised)
- [ ] `LLMRouter.dispatch_stream()` yields `{"type": "text", "content": str}` chunks for a valid Ollama/OpenAI provider
- [ ] Unit tests cover both helpers with a mock Bifrost/DB

## Blocked by

None — can start immediately

craig-dt:
This looks **already implemented** (under different names) and can likely be closed as completed — flagging so nobody re-does it.

Both requested capabilities exist in `services/llm_router.py` today, wired into the live chat path:

| This issue asked for | Exists now | Delivered by |
|---|---|---|
| `get_active_provider_spec()` — generalized active-provider lookup | `get_default_provider_spec()` — `services/llm_router.py:655` | #341 |
| `LLMRouter.dispatch_stream()` — SSE stream non-Anthropic via Bifrost `/v1` | `LLMRouter.dispatch_openai_stream()` — `services/llm_router.py:493` (+ `stream_openai_raw()` at `:430`) | #348, #357 |

### Acceptance criteria mapping
[...]

PR #437, opened 2026-07-23
test(llm): cover LLMRouter non-Anthropic streaming path (#436)

### LINKED ISSUE #436: test: add unit coverage for LLMRouter streaming (dispatch_openai_stream / stream_openai_raw)
Follow-up from #325. The non-Anthropic streaming path (`LLMRouter.dispatch_openai_stream` / `stream_openai_raw` in `services/llm_router.py`) is wired into the chat SSE path (`backend/api/claude.py:901`) but has no dedicated by-name unit test with a mocked Bifrost SSE stream. Coverage today is indirect (routing decision only). Add unit tests asserting: text chunks are yielded as `{'type':'text','content':...}`, empty/no-content chunks are skipped, `AsyncOpenAI` is built with `base_url=.../v1` and `stream=True`, `include_usage` sets `stream_options`, and the client is closed on normal completion and on early consumer disconnect (GeneratorExit).

craig-dt:
## What

Adds dedicated unit coverage for `LLMRouter`'s non-Anthropic streaming path — the one gap flagged when reviewing #325. Closes #436.

`LLMRouter.dispatch_openai_stream()` / `stream_openai_raw()` (`services/llm_router.py`) stream non-Anthropic providers through Bifrost's OpenAI-compatible `/v1` SSE endpoint and are wired into the chat SSE path at `backend/api/claude.py:901`. Until now they had only indirect coverage (the routing *decision* was tested; the stream generators themselves weren't).

## Tests added (`tests/unit/test_llm_router.py`)

- **`test_dispatch_openai_stream_yields_text_and_skips_empty`** — text deltas surface as `{"type": "text", "content": ...}`; frames with no `choices` or no `content` are skipped; `AsyncOpenAI` is constructed with `base_url=".../v1"`, `create()` is called with `stream=True` and a provider-prefixed model (`ollama/llama3.1:8b`); client closed on normal completion.
- **`test_stream_openai_raw_include_usage_sets_stream_options`** — `include_usage=True` sets `stream_options={"include_usage": True}`.
- **`test_stream_openai_raw_closes_client_on_early_disconnect`** — `aclose()` mid-stream propagates `GeneratorExit` into the `yield`, and the `finally` still closes the client (no httpx pool leak).

## Notes

- Mocks Bifrost via `patch("openai.AsyncOpenAI")` + `AsyncMock`, matching the existing dispatch tests in this file. No network/DB.
- Deliberately asserts the disconnect/close behavior on `stream_openai_raw` directly rather than through `dispatch_openai_stream`: a plain `async for` isn't generator delegation, so closing the outer generator wouldn't synchronously run the inner `finally` — testing the inner generator's `aclose()` is the deterministic check.
- Full file: **30 passed** (3 new, 0 regressions).

### DIFF
--- a/tests/unit/test_llm_router.py
+++ b/tests/unit/test_llm_router.py
@@ -863,3 +863,112 @@ def test_discover_anthropic_api_key_returns_none_when_db_unavailable():
 
     with patch.object(builtins, "__import__", side_effect=boom_import):
         assert llm_router.discover_anthropic_api_key() is None
+
+
+# ---------------------------------------------------------------------------
+# Streaming — dispatch_openai_stream / stream_openai_raw (GH #325, #436)
+# ---------------------------------------------------------------------------
+
+
+def _delta_chunk(content):
+    """A minimal OpenAI streaming chunk carrying a content delta."""
+    return SimpleNamespace(
+        choices=[SimpleNamespace(delta=SimpleNamespace(content=content))]
+    )
+
+
+async def _achunks(chunks):
+    for c in chunks:
+        yield c
+
+
+@pytest.mark.asyncio
+async def test_dispatch_openai_stream_yields_text_and_skips_empty():
+    """The chat SSE path streams non-Anthropic providers through Bifrost's
+    OpenAI-compatible /v1 endpoint. Text deltas must surface as
+    {"type": "text", "content": ...}; frames with no choices or no content
+    (e.g. role/usage-only frames) must be skipped."""
+    router = LLMRouter(bifrost_url="http://test-bifrost:8080")
+    chunks = [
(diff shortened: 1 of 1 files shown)


mattmorr1:
Reviewed — solid test-only addition. Verified all three cases pass against current main (`pytest tests/unit/test_llm_router.py -k stream` → 3 passed). Good coverage of the non-Anthropic SSE path: text-delta surfacing, empty/role-only frame skipping, the Bifrost `/v1` base_url and provider-prefixed model, `stream_options.include_usage`, and — the valuable one — client cleanup on early consumer disconnect (`GeneratorExit` → `finally` closes the client, no httpx pool leak). No changes requested.

PR #438, opened 2026-07-23
fix(redesign): define missing chat-dock width helpers (repairs broken main)

mattmorr1:
## Hotfix: define the chat-dock width helpers merged undefined in #401

`main` does not compile and the SOC console crashes at render. #401 (`e555120`) merged `SocConsole.tsx` code that references `readChatWidth`, `clampChatPreference`, `chatMaxForViewport`, `CHAT_MIN_WIDTH`, `CHAT_MAX_WIDTH`, `CHAT_WIDTH_STORAGE_KEY` and uses the `CSSProperties` type without defining or importing any of them. `tsc --noEmit` reports 7 errors; at runtime `<SocConsoleInner>` throws `ReferenceError: readChatWidth is not defined`.

It slipped through because the `Build Frontend` CI job is SKIPPED for these PRs and the vite/esbuild build does not fail on undefined identifiers or type errors (only `tsc` does, and nothing gates on it).

### Fix
- Define the module-level constants (`CHAT_MIN_WIDTH=360`, `CHAT_MAX_WIDTH=720`, `CHAT_DEFAULT_WIDTH=420`, `CHAT_WIDTH_STORAGE_KEY='soc.chat.width.v1'`) and helpers (`clampChatPreference`, `chatMaxForViewport`, `readChatWidth`) matching #401's documented 360–720px range, 420px default, and `soc.chat.width.v1` persistence key. `readChatWidth` handles a missing key, non-numeric/empty values, and out-of-range widths.
- Import `CSSProperties` from react.
- Apply `consoleStyle` (which carries `--chat-w`) to the console wrapper — it was defined but never wired, so the dock never received its width and resizing had no visual effect.
- Add the `maxWidth <= minWidth` guard to `onResizeKeyDown` in `Chat.tsx` for parity with the pointer handler.

### Validation
- `tsc --noEmit`: clean (was 7 errors)
- `vitest` SocConsole + Chat: 16/17 pass. The one failure — the pre-existing "renders the 404 screen" test — is unrelated: it renders the extension-loading branch because `useExtensions()` reports `loading: true` in the test env, so `NotFoundScreen` never mounts. That is orthogonal to this compile fix and left untouched.
- `npm run build`: passes.

### Follow-up (not in this PR)
- Add a `tsc --noEmit` gate to CI frontend lint, or stop skipping `Build Frontend`, so an undefined-identifier regression like this cannot merge again.

### DIFF
--- a/frontend/src/redesign/SocConsole.tsx
+++ b/frontend/src/redesign/SocConsole.tsx
@@ -3,7 +3,7 @@
    dock, floating "Ask Vigil" FAB, and the theme tweaks panel.
    Ported from the design's index HTML + main.js.
    ============================================================ */
-import { useCallback, useEffect, useMemo, useState } from 'react'
+import { useCallback, useEffect, useMemo, useState, type CSSProperties } from 'react'
 import { useNavigate, useParams } from 'react-router-dom'
 import './styles.css'
 import { useAuth } from '../contexts/AuthContext'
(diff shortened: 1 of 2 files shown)


nestor-deeptempo (approved):
👍

PR #443, opened 2026-07-24
refactor(#413): introduce provider-agnostic LoopController + TurnEngine (PR3a)

craig-dt:
## What

Behaviour-preserving refactor that extracts the multi-turn agentic tool loop out of `OpenAIAgentService.stream` into a new, provider-agnostic seam in `services/agent_loop.py`:

- **`LoopController`** — the provider-agnostic loop skeleton: iteration/wall-clock guards, inter-iteration backoff, canonicalized loop-detection, the tool-execution + human-approval flow, per-turn telemetry, and the iteration-limit message.
- **`TurnEngine`** — the interface a provider adapter implements, plus **`OpenAITurnEngine`**, which speaks the OpenAI-compatible dialect (delta reassembly over `LLMRouter.stream_openai_raw`, tool-call normalization, usage accounting, malformed-tool-JSON heuristic).
- `OpenAIAgentService.stream()` now builds an engine + controller and delegates, passing itself as the controller's tool-exec/approval/telemetry **runtime**. Net **−321 lines** in that file.

Each engine owns its own message-history dialect, so the controller only ever handles provider-agnostic `TurnResult` / `NormalizedToolCall` values. A future `AnthropicTurnEngine` can keep native Anthropic-block history with no conversion at the boundary.

## Why

This is **PR3a** of the phased work to make `LLMRouter` the single LLM entry point (#413). It establishes the provider-agnostic loop seam that later sub-PRs build on (Anthropic engine, unified persistence/budget, `run_agent_task` passthrough) without landing one giant diff. This PR is purely structural — **no behaviour change**.

## How it was tested

- **Parity guard, unmodified:** the existing `tests/unit/test_openai_agent_service.py` (27 tests) passes **without edits** — it exercises the real controller + engine after the refactor and is the behaviour-preservation proof.
- **New focused unit tests:** `tests/unit/test_agent_loop.py` (22 tests) cover the controller against a fake engine/runtime (tool loop, loop-detection, wall-clock guard, iteration limit, streaming-error path, full approval flow) and the `OpenAITurnEngine` delta reassembly (text, tool-call fragments across chunks, empty-name skip, id synthesis, usage, malformed heuristic).
- **49/49 pass.** `black` / `isort` / `flake8` clean; `mypy` clean on `services/agent_loop.py`.
[...]

Issue #413, opened 2026-07-23
Complete the LLM provider boundary and enforce it with an import contract


epowell101:
**Track: suggested cleanup.**

Consolidates #325 (LLMRouter: `get_active_provider_spec()`, `dispatch_stream()`), #326, #327, #328 (endpoints and model resolution hardcode ClaudeService), and #393 (unify tool dispatch). Fourteen non-test modules import `claude_service` directly today, trapping the module's depth and blocking non-Anthropic providers. Future hunt agents must run on any configured provider.

**Acceptance criteria**
- The five referenced issues closed
- An import-linter contract in CI forbidding `from services.claude_service import` outside the router and its tests
- `docs/AGENTS.md` names `LLMRouter` as the sole entry point for LLM calls

mattmorr1:
done

PR #444, opened 2026-07-24
feat(#413): AnthropicTurnEngine behind the shared LoopController (PR3b)

### LINKED ISSUE #442: test: tests/unit/test_agent_runner_preflight.py fails on event-loop teardown when run after async test files (Py3.13)
## Summary

`tests/unit/test_agent_runner_preflight.py` fails with an event-loop `RuntimeError` when it runs **after** an async test file (e.g. `tests/unit/test_openai_agent_service.py`) in the same `pytest` process. Run on its own it passes; run in the full `tests/unit` suite it fails. This is a **test-isolation bug**, not a product bug, and it reproduces on a clean `main` (no relation to any in-flight feature branch).

## Reproduce

```bash
# Fails (5 failed):
python -m pytest tests/unit/test_openai_agent_service.py tests/unit/test_agent_runner_preflight.py -q

# Passes on its own (5 passed):
python -m pytest tests/unit/test_agent_runner_preflight.py -q
```

## Symptom

```
RuntimeError: There is no current event loop in thread 'MainThread'.
  .../python3.13/asyncio/events.py:716: RuntimeError
RuntimeWarning: coroutine 'AgentRunner._preflight_budget_blocked' was never awaited
```

## Likely cause

On Python 3.13, `asyncio.get_event_loop()` no longer creates/returns a loop when none is set for the thread — it raises. A preceding async test file closes the loop it created, leaving no current loop; the preflight test (or a helper/fixture it uses) then calls `asyncio.get_event_loop()` and hits the `RuntimeError`. Because the coroutine is never scheduled, the follow-on `RuntimeWarning: coroutine ... was never awaited` appears.

## Suggested fix

Have the preflight test drive its coroutines via `asyncio.run(...)` or the `pytest-asyncio` event-loop fixture / `@pytest.mark.asyncio` rather than `asyncio.get_event_loop()`. Alternatively, set a project-wide `asyncio_mode`/event-loop policy in the pytest config so async tests don't leak loop state across files.

## Impact

CI ordering may mask or expose this depending on collection order; locally it makes the full `tests/unit` run red. No production code is affected.

_Filed while working on #413 (PR3a). Surfaced by, but unrelated to, that refactor._

craig-dt:
## What

**PR3b** of #413. Introduces `AnthropicTurnEngine` so the Anthropic chat path runs behind the same provider-agnostic `LoopController` that PR3a established for the OpenAI path. Two structural commits + one review-driven fix:

1. **`refactor(agent-loop): thin the controller`** — moves the tool-execution phase and per-iteration telemetry *out* of `LoopController` and *into* the `TurnEngine` (`execute_tools()` + telemetry at the end of `stream_turn()`). The controller now owns only the cross-cutting guards every provider shares: iteration cap, wall-clock timeout, backoff, canonicalized loop-detection, and the stop/limit messages. It no longer holds a runtime. Behaviour-preserving for the OpenAI path — `test_openai_agent_service.py` passes **unmodified**.
2. **`feat(agent-loop): add AnthropicTurnEngine; route chat_stream through it`** — an Anthropic-native `TurnEngine` that reproduces `ClaudeService.chat_stream`'s loop body: native `messages.stream` with `thinking_start`/`thinking`/`thinking_end` + `text` events, `get_final_message()` usage, per-iteration reasoning-trace persistence, and batch tool execution via `_process_mixed_tool_use`. It wraps a `ClaudeService` and reuses its proven helpers, so canonical history stays native Anthropic blocks (no conversion at the boundary). `ClaudeService.chat_stream` keeps its setup (message/tool/thinking assembly, `_prepare_context_async` + `context_windowed`) and delegates the multi-turn loop.
3. **`fix(chat): surface {"type":"error"} stream events in the chat UIs`** — see delta #2 below.

## Why

Making `LLMRouter` the single LLM entry point (#413) requires the Anthropic loop to run behind the same controller as the OpenAI loop. Their turn bodies diverge in every particular (thinking events, `context_windowed`, batch-vs-per-tool execution, block-vs-OpenAI history, telemetry shape), so the controller had to become a thin sequencer and each engine own its own dialect. This sets up the later sub-PRs (unified persistence/budget in 3c, `run_agent_task` in 3d, caller migration + uniform gate in 3e).

## Deliberate behaviour deltas on the Anthropic chat path

All three follow from unifying the two loops behind one controller, and are intentional:

1. **Loop-detection** now uses the canonicalized `frozenset` signature (the unification chosen in review) instead of `str(input)` list-containment. It requires 3 *consecutive* identical tool-sets; the previous check tripped on 2 repeats and caught `A,B,A,B` oscillation. Bounded by the 30-iteration / 300s guards either way.
2. **Per-turn stream errors** now yield `{"type":"error"}` and stop gracefully (matching the OpenAI path) instead of raising. Both chat consumers only handled a top-level `error` key, so commit 3 adds `type === 'error'` handling to `Chat.tsx` and `ClaudeDrawer.tsx` — which also repairs the pre-existing OpenAI/Ollama path (it already emitted `{"type":"error"}` and dropped it). Setup-phase errors still raise.
3. **Assistant message** is appended just before loop-detection rather than after — unobservable (a loop-detect trip discards engine-local history).

The tier/approval gate remains **absent** on the Anthropic path (unchanged from `chat_stream`); unifying it across providers is deferred to PR3e, alongside the frontend `approval_required` handler.

## How it was tested

- `test_openai_agent_service.py` (OpenAI end-to-end parity) passes **unmodified** — the behaviour-preservation proof for commit 1.
- `test_agent_loop.py` rewritten for the thin-controller split + new `AnthropicTurnEngine` characterization tests (thinking events, tool-use normalization, `_persist_interaction` args, usage, batch tool exec).
- The existing `chat_stream` test and `test_claude_service` / `test_claude_chat_routing` suites pass.
- `black` / `flake8` / `mypy` clean on all new/changed backend files; `eslint` clean on the two frontend files.
- A fresh-eyes reviewer confirmed both commits are faithful lifts; the one HIGH finding (errors silently dropped by the UIs) is fixed in commit 3.

## Notes for reviewers

- **Stacked on #443 (PR3a).** Until #443 merges, this PR's diff includes 3a's commits plus the `test(preflight)` fix (`627ca42`, "Fixes #442"). The 3b-specific changes are the top three commits.
- `services/claude_service.py` has pre-existing `flake8`/`black` findings unrelated to this change; only the `chat_stream` region was touched.

Part of #413.

### DIFF
--- a/frontend/src/components/claude/ClaudeDrawer.tsx
+++ b/frontend/src/components/claude/ClaudeDrawer.tsx
@@ -378,8 +378,8 @@ export default function ClaudeDrawer({ open, onClose, initialMessages, initialAg
                   if (!data) continue
                   let event: any
                   try { event = JSON.parse(data) } catch { continue }
-                  if (event.error) {
-                    throw new Error(event.error)
+                  if (event.error || event.type === 'error') {
+                    throw new Error(event.error || event.content || 'stream error')
                   } else if (event.type === 'thinking_start') {
                     setIsThinking(true)
                     currentThinking = ''
@@ -598,8 +598,8 @@ export default function ClaudeDrawer({ open, onClose, initialMessages, initialAg
                     continue
                   }
                   
-                  if (event.error) {
-                    throw new Error(event.error)
+                  if (event.error || event.type === 'error') {
+                    throw new Error(event.error || event.content || 'stream error')
                   } else if (event.type === 'context_windowed') {
                     logger.info(`Context compressed: ${event.windowed_messages} older messages condensed, ${event.remaining_messages} recent messages kept`)
                     currentText += `[Context compressed: ${event.windowed_messages} older messages were condensed to preserve context within the model's limits. Recent messages and all key details are preserved.]\n\n`
--- a/frontend/src/redesign/shell/Chat.tsx
+++ b/frontend/src/redesign/shell/Chat.tsx
@@ -622,7 +622,8 @@ export default function Chat({
             } catch {
               continue
             }
-            if (ev.error) throw new Error(ev.error)
+            if (ev.error || ev.type === 'error')
+              throw new Error(ev.error || ev.content || 'stream error')
             if (ev.type === 'thinking_start') {
               setIsThinking(true)
               curThinking = ''
--- a/services/agent_loop.py
+++ b/services/agent_loop.py
@@ -3,21 +3,22 @@
 This module houses the two halves of the multi-turn tool loop that used to live
 entirely inside ``OpenAIAgentService``:
 
-  - :class:`LoopController` — the **provider-agnostic** loop skeleton: iteration
-    and wall-clock guards, inter-iteration backoff, canonicalized loop-detection,
-    the tool-execution + human-approval flow, per-turn telemetry, and the
-    iteration-limit message. It knows nothing about any specific provider; it
-    drives a :class:`TurnEngine` for model I/O and calls back into a ``runtime``
-    handle for tool execution, approval, and logging.
+  - :class:`LoopController` — the **provider-agnostic** loop *sequencer*: only
+    the cross-cutting guards every provider shares — the iteration cap, the
+    wall-clock timeout, inter-iteration backoff, canonicalized loop-detection,
+    and the stop/limit messages. It knows nothing about any specific provider; it
+    drives a :class:`TurnEngine` and forwards its events.
   - :class:`TurnEngine` — the interface a provider adapter implements, plus the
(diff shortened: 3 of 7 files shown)


craig-dt:
Closing to consolidate. We're rolling all #413 (LLM provider boundary) work into a single PR, to be opened once the full boundary is complete and reviewable as one coherent change — loop controller + both TurnEngines, dispatch persistence, the Agent-SDK passthrough, the uniform approval gate, and the import-linter contract. No work is lost: the branch is preserved and nothing is merged. Withdrawing this partial/stacked PR for now.

Issue #450, opened 2026-07-25
Route llm_worker.py raw multi-turn loop off ClaudeService internals (complete the #413 import boundary)


craig-dt:
Follow-up from the #413 LLM-provider-boundary work.

**Context.** #413 makes `services/llm_router.py` (LLMRouter) the sole entry point to LLM dispatch, enforced by an import-linter contract. The contract ships with a **temporary, documented module-level exemption for `services/llm_worker.py`**; this issue tracks removing it.

**Why llm_worker is exempted.** `llm_worker` is the ARQ execution substrate behind `LLMGateway`. It is already **router-first**: `llm_call` / `llm_call_raw` call `_maybe_dispatch_via_router(...)`, which routes any non-default-Anthropic provider through `LLMRouter`. It falls back to `ClaudeService` only for the **default-Anthropic path**:

- `_sync_claude_call` → `ClaudeService.chat()` (a plain chat; could route through `LLMRouter.chat`, but that does not remove the import — see below)
- `_sync_claude_raw` → reaches into `claude_service.client.messages.create(...)` **and** the private `claude_service._persist_interaction(...)` for the **raw multi-turn tool loop** used by the daemon AgentRunner. There is no `LLMRouter` equivalent for a raw multi-turn Anthropic entry today.

Because `on_startup` constructs the shared `ClaudeService` for `_sync_claude_raw`, migrating `_sync_claude_call` alone would **not** remove the import — so the whole module is exempted rather than churning the hot worker path for no boundary gain.

**Proposed fix.** Add a raw multi-turn entry point to `LLMRouter` (Anthropic `messages.create` loop + interaction persistence) that `llm_worker` can call, so `_sync_claude_raw` no longer touches `ClaudeService.client` / `._persist_interaction`. Then drop the import-linter exemption for `services/llm_worker.py`.

**Done when:** `services/llm_worker.py` imports no `ClaudeService`, and the import-contract exemption for it is removed (CI still green).

mattmorr1:
Consolidated into #449, which now tracks removing **both** temporary #413 import-contract exemptions (`daemon/agent_runner.py` and `services/llm_worker.py`) as two separately-shippable sub-tasks. The full `llm_worker` analysis and proposed fix from this issue were carried over verbatim, including the note that #471/#497 may relocate the raw multi-turn loop first. Closing as a duplicate — no scope lost.

Issue #449, opened 2026-07-25
Remove the two temporary #413 import-contract exemptions (agent_runner + llm_worker)


craig-dt:
Follow-up from the #413 LLM-provider-boundary work. Consolidates #450.

**Context.** #413 makes `services/llm_router.py` (LLMRouter) the sole entry point to LLM dispatch, enforced by an import-linter contract (forbidding `services.claude_service` imports outside `services.llm_router` + tests). The contract ships with **two temporary, documented exemptions** — `daemon/agent_runner.py` and `services/llm_worker.py`. This issue tracks removing both, so the exemption list reaches zero and the boundary is real rather than declared.

The two are independent and land as separate PRs; they differ sharply in effort, so don't treat this as one unit of work.

---

## Sub-task 1 — `daemon/agent_runner.py` (Coding: Low)

**Why it's exempted (letter-of-the-rule, not spirit).** agent_runner's real LLM entry is already `LLMGateway` (→ ARQ → `llm_worker`), and #393/PR2 moved its tool execution to `tool_manager.execute_backend_tool`. It imports `ClaudeService` only for **4 non-dispatch helper borrowings**, not to dispatch an LLM call:

- `backend_tools` (the tool schema list)
- `_truncate_tool_response`
- `enable_thinking` / `thinking_budget` (config values)
- `_skill_tool_index`

**Proposed fix.** Relocate these 4 helpers to a neutral module (e.g. a small `services/agent_tooling.py` or onto `tool_manager`) that both `ClaudeService` and `agent_runner` import, so `agent_runner` no longer imports `ClaudeService`. Then drop the import-linter exemption for `daemon/agent_runner.py`.

## Sub-task 2 — `services/llm_worker.py` (Coding: High)

**Why it's exempted.** `llm_worker` is the ARQ execution substrate behind `LLMGateway`. It is already **router-first**: `llm_call` / `llm_call_raw` call `_maybe_dispatch_via_router(...)`, which routes any non-default-Anthropic provider through `LLMRouter`. It falls back to `ClaudeService` only for the **default-Anthropic path**:

- `_sync_claude_call` → `ClaudeService.chat()` (a plain chat; could route through `LLMRouter.chat`, but that does not remove the import — see below)
[...]

PR #448, opened 2026-07-24
fix(redesign): define missing chat-dock resize helpers (#401 regression)

craig-dt:
## What

Restores the redesign SOC console, which white-screens on `main`. `SocConsole.tsx` (from #401) referenced five module helpers — `readChatWidth`, `clampChatPreference`, `CHAT_WIDTH_STORAGE_KEY`, `chatMaxForViewport`, `CHAT_MIN_WIDTH` — that were never defined or imported, so `useState(readChatWidth)` throws `ReferenceError` in `SocConsoleInner` and the whole console renders blank.

## Why

#401 ("make assistant dock resizable") appears to have lost these helper definitions (and the `consoleStyle`/`--chat-w` wiring) when it merged. All 15 `SocConsole.test.tsx` cases fail on `main` as a result.

## Changes

- Define the five helpers at module scope in `SocConsole.tsx`: width band **[360, 720]**, default **420**, localStorage key **`soc.chat.width.v1`** (the key the tests already assert), `clampChatPreference` + a per-viewport max (never more than half the screen), and `readChatWidth` restoring the persisted preference with a safe default.
- Wire the dropped `consoleStyle` onto the root element (folding in `bgVars`) so the `--chat-w` dock-width CSS variable is actually applied — it was computed but never used — and import the `CSSProperties` type it needs.

## How tested

- `npx vitest run src/redesign/SocConsole.test.tsx`: **0/15 → 13/15**.
- Loaded the app in a headless browser: the console mounts, nav + dashboard render, and the resizable chat dock opens at its 420px default with **0 render errors**.
- `eslint` + `tsc` clean on `SocConsole.tsx`.

## Out of scope (noted)

The 2 still-failing `SocConsole.test.tsx` cases are a **separate, pre-existing** test-mock gap — the shared-api mock lacks `caseMetricsApi` and a `default` export that the Metrics/Analytics screens now require. Unrelated to the resize dock (it was masked by this crash) and left for a follow-up.

Closes #447

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #393, opened 2026-07-16
Consolidate backend tool dispatch: route ClaudeService + daemon through tool_manager.execute_backend_tool


mattmorr1:
Follow-up from PR #357 review (finding #8).

PR #357 introduced `services/tool_manager.py::execute_backend_tool` as a shared backend-tool dispatch path, but only `OpenAIAgentService` routes through it. `ClaudeService` and the daemon still use their own table:

- `services/claude_service.py` — `_execute_backend_tool`
- `daemon/agent_runner.py` — calls `self._claude_service._execute_backend_tool` (~lines 1256, 1517)

So there are currently two dispatch tables. The `tool_manager` docstring was corrected in #357 to stop claiming both loops route through it; this issue tracks the actual consolidation.

**Scope**
- Migrate `ClaudeService._execute_backend_tool` callers (including the daemon's two call sites) to `tool_manager.execute_backend_tool`.
- Ensure tool-name mapping, `wrap_tool_result` wrapping, and approval/tier gating remain consistent across chat, daemon, and workflow paths.
- Remove the now-redundant table once all callers are migrated.

**Why deferred**: kept out of #357 to limit blast radius on the Anthropic/daemon paths; #357 shipped the OpenAI path on the shared dispatcher only.

PR #451, opened 2026-07-27
feat(#413): make LLMRouter the sole LLM entry point + enforce with an import contract

### LINKED ISSUE #413: Complete the LLM provider boundary and enforce it with an import contract

craig-dt:
## Summary

Makes **`LLMRouter` (`services/llm_router.py`) the sole entry point for every LLM call** in Vigil, and enforces it with an import-linter contract so the boundary can't erode. This lets every agent/workflow/endpoint run on any configured provider (Anthropic, OpenAI, or a local Ollama via Bifrost) instead of being implicitly Anthropic-only, and removes the direct `ClaudeService` coupling from the whole codebase.

Closes #413. (Also completes the intent of the child issues #325 and #327.)

## Why

Previously ~14 non-test modules imported `services.claude_service.ClaudeService` directly and called `ClaudeService.chat(...)` — a single call that couples LLM dispatch with the multi-turn agentic tool loop and MCP tool loading. Because `ClaudeService` is the Anthropic SDK path, every one of those call sites was implicitly Anthropic-only. There was no enforced boundary, so new code kept reaching for `ClaudeService`. See [ADR 0001](docs/adr/0001-llmrouter-sole-llm-entry-point.md).

## What changed (phased; each phase was fresh-reviewed before the next)

- **Docs (PR1):** ADR 0001 + `CONTEXT.md` glossary establishing the boundary decision.
- **Tool dispatch (PR2, #393):** consolidated the daemon's tool execution onto `tool_manager.execute_backend_tool` (one tool table).
- **Router internals (PR3):** a provider-agnostic `LoopController` + per-provider `TurnEngine`s (Anthropic native-stream + OpenAI over Bifrost); unified streaming; dispatch-layer interaction persistence; per-tool approval gating on the Anthropic path; the Anthropic Agent SDK folded behind `LLMRouter.run_agent_task`; streaming budget errors surfaced as a typed banner.
- **Caller migration (PR4):** redirected every feature caller off `ClaudeService` onto `LLMRouter` — chat / streaming / Agent-SDK endpoints (`backend/api/claude.py`, `agents.py`), `findings.py`, `timeline.py` (relocated `generate_event_analysis` into a new provider-agnostic `services/timeline_analysis_service.py`), the text-generation services, `workflows_service.py` (via a new `LLMRouter.run_agent_chat`), the daemon scheduler health probe, and the demo script. New router entry points: `chat`, `dispatch_stream`, `run_agent_chat`, plus availability helpers `anthropic_api_key_available()` / `agent_sdk_available()`.
- **Enforcement (PR5):** an import-linter `forbidden` contract ([`.importlinter`](.importlinter)) forbidding any import of `services.claude_service` outside `services.llm_router`, run by a **blocking** `import-contract` CI job and a `lint-imports` pre-commit hook.

## Exemptions (temporary, tracked)

Two LLM-plumbing modules keep the import under a documented exemption — they borrow `ClaudeService` internals rather than being feature callers, and are already router-first for their main paths:

- `daemon/agent_runner.py` — 4 non-dispatch helper borrowings — #449
- `services/llm_worker.py` — the raw multi-turn loop (`.client` + `_persist_interaction`) — #450

## How it was tested

- Every slice was reviewed by a fresh agent (author never grades its own homework); all verdicts FAITHFUL.
- Unit suites green for the router, agent-loop, OpenAI-agent, chat-routing, workflow, generator, and timeline paths; new tests added per slice.
- `lint-imports` = **1 kept, 0 broken** (233 files, 572 deps), and proven non-vacuous (removing an exemption breaks the contract). The contract needs no runtime deps (grimp static analysis), verified in a clean venv.

## Reviewer notes

- **`import-contract` should be added as a required status check** in branch protection — the job hard-fails on a boundary violation (no `continue-on-error`), but "required" is a GitHub setting.
- **Rebased onto current `main` — no conflicts.** `main` had gained #438 (an equivalent fix for the #401 chat-dock regression) and #437 (a LLMRouter streaming test). This branch has been rebased so it sits directly on top of `main`: the branch's own now-redundant #401 fix was **dropped** (superseded by #438 — so **PR #448 is now moot**), and its `tests/unit/test_llm_router.py` additions were **merged with #437's** (both test sets kept). The PR is mergeable.
- The 17 known local-only unit failures (`test_pr_d_mechanical` / `test_prompt_caching` / `test_workflow_*`) are environment artifacts (no local DB/Bifrost) and pass on CI.


mattmorr1:
@craig-dt @johnvanlowe I think we need to have a bit of a longer conversation before we go ahead with this

mattmorr1:
if we keep work on this, please gitignore the .jsons that lead to 288k+ lines committed.

craig-dt:
Merged current `main` into this branch to clear the conflict, and removed the
generated artifacts that were drowning the diff. It is mergeable again, and now 42
files instead of 535.

## The resolution

One conflict, `frontend/src/components/claude/ClaudeDrawer.tsx` — deleted by #502,
modified here. **Taken as a deletion.**

Nothing is lost by that. This branch's only change to the drawer was two identical
hunks broadening SSE error detection:

```diff
-  if (event.error) {
-    throw new Error(event.error)
+  if (event.error || event.type === 'error') {
+    throw new Error(event.error || event.content || 'stream error')
```

The same broadening already exists on the surviving chat surface,
`frontend/src/redesign/shell/Chat.tsx`, which this branch also modifies:

```js
if (ev.error || ev.type === 'error')
  throw new Error(ev.error || ev.content || 'stream error')
```

No import of `ClaudeDrawer` remains in the tree — the only surviving mention is a
comment in `Chat.tsx`. Everything else auto-merged, including `backend/api/findings.py`
and `.github/workflows/ci-cd.yml`.

**Merged rather than rebased on purpose.** A rebase needs a force-push, which would
rewrite this branch's 36 commits and detach the review context already anchored to
them. The merge is append-only.

## Verification

Full suite locally, diffed **by test name** against clean `main` at `311790e`
(Python 3.13.9):

- **No new failures.** Every test that fails on this branch also fails on `main`.
- 15 remain red on both: `test_pr_d_mechanical` ×6, `test_prompt_caching` ×4,
  `test_workflow_phase_approval` ×4, `test_unauth_endpoints` ×1. All pre-existing.
- 14 tests red on `main` are green here — **but that is not a product fix and should
  not be read as one.** All 14 live in test files this branch modifies, so the two
  runs execute different test code. Specifically: the 5 `test_agent_runner_preflight`
  failures are `RuntimeError: There is no current event loop in thread 'MainThread'`,
  because `main`'s version calls `asyncio.get_event_loop().run_until_complete()`,
  which raises on 3.13 once any earlier async test has run; this branch uses
  `asyncio.run()`. The `test_claude_api` and `test_workflow_skill_wiring` ones are
  rewritten here to target the router seam.

## Also in this push: the diff is reviewable again

This branch was showing **535 changed files against 41 real ones** — ~92% noise. The
cause was on this branch, not in history: `3e5eec2` (the previous
`Merge remote-tracking branch 'origin/main'`) committed 499 files, of which **494 were
`graphify-out/`** — generated knowledge-graph output swept in while resolving that
merge. 485 are machine-generated AST cache
(`graphify-out/cache/ast/v0.8.40/<sha256>.json`); the rest are a rendered graph, an
HTML call-flow, and a cost report. ~20MB, no source, all regenerable.

`682065e` untracks them with `git rm -r --cached` (the directory stays on disk) and
adds the ignore rule. **The branch is now 42 files against `main`.**

The `.gitignore` line is verbatim the one #505 proposed for `main`, so when this
branch lands, `main` picks up the same rule and the artifacts cannot be recommitted
from anywhere. I have closed #505 as superseded rather than leave two PRs adding the
same three lines — if #451 is going to be rejected or deferred, say so and I will
reopen it, since `main` should get that rule either way.

I scoped this to #451 on the basis that these artifacts were introduced here, by the
commit named above, so removing them undoes this branch's own damage rather than
riding an unrelated fix along. Say the word if you would still rather see it split
out.

## One thing reviewers should know

**#522 overlaps this branch on 17 files.** `core/llm/` consolidation and this
branch's `LLMRouter` boundary are working the same ground:

```
.github/workflows/ci-cd.yml   backend/api/{agents,claude,findings,timeline}.py
daemon/{agent_runner,scheduler}.py
services/{agent_ai_generator,custom_integration_service,skill_service}.py
services/{workflow_ai_generator,workflows_service}.py
tests/integration/test_claude_api.py
tests/unit/{test_agent_runner_preflight,test_workflow_phase_approval,test_workflow_skill_wiring}.py
```

#522 currently reports `CONFLICTING` against `main` on its own account
(`core/secrets.py` and `services/splunk_enrichment_service.py`, both modify/delete),
so I have not tried to predict the combined resolution. I am happy to go second and
rebase onto #522 once it lands — just say which order you want.

---

*Analysis by Claude (Claude Code), reviewed by @craig-dt before posting. The merge
resolution and the by-name test comparison were reproduced locally against `main` at
`311790e`; the file overlap was taken from `refs/pull/522/head`.*

craig-dt:
Rebased onto `main` at `d31c8fa`. The `backend/api/findings.py` conflict is resolved —
#537 extracted that logic into `services/findings/enrichment/`, so I took main's
version wholesale; the code moved rather than changed hands.

That leaves one question I'd rather ask than answer unilaterally, because it touches
@thetosy's module rather than mine.

## The import contract is now broken by a path that did not exist when this PR opened

```
$ lint-imports
ClaudeService is reachable only through LLMRouter (#413) BROKEN

backend is not allowed to import services.claude_service:
-   backend.api.findings -> services.findings.enrichment (l.10)
    services.findings.enrichment -> services.findings.enrichment.service (l.24)
    services.findings.enrichment.service -> services.claude_service (l.89)
```

`services/findings/enrichment/service.py:89` constructs `ClaudeService` directly for
the Anthropic branch and dispatches through `loop.run_in_executor`, while the
non-Anthropic branch already goes through `LLMRouter().dispatch()` at `:142`.

This is not a criticism of #537 — it faithfully carried the old handler's behaviour
across, and said so. The module docstring lists it as one of two "known-imperfect
behaviours … carried over from the old inline handler rather than fixed here, so any
regression stays bisectable", and `tests/unit/test_findings_enrichment.py:462-466`
goes further:

> Anthropic goes through the sync ClaudeService on a threadpool with NO retry; every
> other provider goes through LLMRouter with a retry loop and local Bifrost recovery.
> That asymmetry is existing behaviour carried over from the handler, so pin it — **if
> it's ever unified, these are the tests to change.**

## The ask

That signpost is exactly the decision in front of me, so: **@thetosy, is now the time
to unify it, and are you happy for it to happen inside this PR?**

Concretely it would mean, in `services/findings/enrichment/service.py`:

- the resolver stops returning `claude_service` in its tuple
- `_dispatch(*, provider, model_id, prompt, claude_service, finding_id)` loses the
  `claude_service` parameter
- the Anthropic key precheck becomes `anthropic_api_key_available()` instead of
  constructing `ClaudeService` to call `.has_api_key()`
- the Anthropic dispatch becomes `await LLMRouter().chat(..., service_config={...})`,
  which runs the sync call off the event loop internally, so the manual executor goes

and in `tests/unit/test_findings_enrichment.py`, the ones you flagged — `_FakeClaudeService`,
`test_anthropic_dispatch_uses_claude_service_with_the_larger_cap`, and the
`fake_dispatch` signature.

**One behavioural note worth your call:** unifying puts the Anthropic path behind
`LLMRouter`'s retry loop, which it has never had. That is arguably the point, but it is
a real change in behaviour on the cloud path, not a pure refactor — so I'd rather you
chose it than discover it.

## The alternatives, and why I like them less

- **A third `ignore_imports` exemption.** There is precedent — `daemon.agent_runner`
  (#449) and `services.llm_worker` (#450) are both exempted and tracked. But those
  borrow *non-dispatch* helpers; this one is a pure dispatch call, which is precisely
  what the contract exists to stop. Exempting it would hollow the contract out at the
  moment of establishing it.
- **Dropping `.importlinter` from this PR.** Then the boundary is documented but not
  enforced, and the enforcement is most of the value.

Happy to do the conversion here, split it into its own PR against `main` that lands
first, or leave it to you — whichever you prefer. This PR stays red on the *Import
Contract (LLM boundary)* job until it's settled, and I'd rather it stayed visibly red
than be quietly exempted.

---

*Analysis by Claude (Claude Code), reviewed by @craig-dt before posting. The violation
was reproduced with `lint-imports` against this branch merged with `main` at `d31c8fa`;
the quoted docstring and test comment are verbatim from `services/findings/enrichment/service.py`
and `tests/unit/test_findings_enrichment.py`.*

mattmorr1:
@craig-dt solved this by force routing via bifrost now. Thanks for the suggestion!

PR #505, opened 2026-07-30
chore: gitignore graphify-out/

craig-dt:
## What

Adds `graphify-out/` to `.gitignore`.

## Why

The graphify tooling writes ~20MB of generated output to `graphify-out/` at the repo root — `graph.json` (9MB), `graph.html`, `vigil-callflow.html`, `GRAPH_REPORT.md`, `manifest.json`, and a `cache/` directory. None of it belongs in the repo, but it showed up as untracked noise in every `git status`, which makes it easy to miss real untracked files.

`.gcloudignore` already excludes the directory; this just brings `.gitignore` in line.

## How tested

`git status` no longer lists `graphify-out/`. No tracked file is newly ignored — `git ls-files graphify-out/` was already empty, so nothing is removed from the index.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -244,3 +244,6 @@ claude_analysis_*.txt
 
 # Autostart list, written by Settings -> Services (see services/autostart_config.py)
 .vigil-autostart
+
+# Generated knowledge-graph output (graphify) — ~20MB, never committed
+graphify-out/


mattmorr1:
This is only an issue if someone is pushing their active vigil session to github, which should never be the case.  this should be appended to #451.

craig-dt:
Closing as superseded by #451.

This PR added a `.gitignore` rule for `graphify-out/`. It turned out the artifacts
were not just unignored — 494 of them were **committed**, on the `#451` branch, by
`3e5eec2` (a previous `Merge remote-tracking branch 'origin/main'` that swept in 499
files, 494 of them generated output).

So the ignore rule alone was treating the symptom. `682065e` on #451 now does both:
untracks the 494 files with `git rm -r --cached` and adds this exact three-line rule.
When #451 lands, `main` gets the same rule this PR proposed, and the files are gone
rather than merely ignored.

Keeping both open would leave two PRs adding identical lines to `.gitignore`, which is
how merge collisions start. If #451 ends up rejected or deferred I will reopen this —
`main` should carry that rule either way.

PR #506, opened 2026-07-30
docs: add airgapped deployment guide

craig-dt:
## What

Adds `docs/AIRGAPPED_DEPLOYMENT.md` — a customer-facing guide for deploying Vigil into networks with no route to the public internet.

Covers both Docker Compose and Helm/Kubernetes, both internal-mirror (Harbor/Nexus/Artifactory) and sneakernet artifact transfer, and a firewall/ACL reference for restricted-egress networks that aren't truly airgapped.

## Why

Airgapped/classified networks are a recurring ask, and "what does airgapping mean for an AI-native SOC platform?" has a non-obvious answer that isn't captured anywhere in the repo today. The existing `DEPLOYMENT_GUIDE.md` assumes internet access throughout.

## The three findings that shape the guide

**1. Vigil does not phone home.** No licensing callback, no usage telemetry, no update check, no analytics SDK (no Sentry/PostHog/Segment/Datadog in the frontend). Observability is self-hosted. Every outbound connection is one the operator configured. This is the first question every airgap security reviewer asks, and it's a clean answer.

**2. The LLM is the airgap problem.** All 13 agents are inference-driven, so local serving is mandatory — this is a capability decision, not a packaging one. The guide presents Ollama / vLLM / customer-supplied endpoint, recommends Ollama for single-site and vLLM for real volume, and states plainly that local open-weight inference degrades tool-calling quality (which matters most for the agents that chain many tool calls). Sizing is flagged as needing an engineer's sign-off rather than guessed at.

**3. Most of the integration catalog doesn't survive.** Of 40 MCP servers:
- **18 fetch their own code at every process launch** via `npx -y`, `uvx`, `uv`, or `docker run` — two straight from `git+https://github.com/…`. These fail at *agent-invocation* time, not deploy time, so a naive airgapped install passes every startup health check and looks fine until an analyst asks for a CrowdStrike lookup.
- ~10 more ship in-repo but target internet services with no on-prem edition (VirusTotal, Shodan, ANY.RUN…).
- Roughly a quarter of the catalog is viable fully airgapped. The guide gives substitutions (MISP for intel lookup, CAPE for detonation) and is explicit that Shodan-class external exposure data has none.

## Also documented

The offline gotchas that cost real time — Bifrost's default RFC1918 block (effectively mandatory to disable, since airgapped inference hosts are almost always on private addresses, and the key-level form is silently dropped), its `config.db` reseeding only on first boot (`--force-recreate`, not `restart`), and the model-catalog refresh loop failing every 5 minutes against unreachable providers.

## Scope notes

- **Hardware sizing is explicitly out of scope** this revision, per the request. Where a decision depends on hardware, the guide states the dependency and stops.
- **Section 10 (firewall/ACL) is marked DRAFT.** The hostname inventory is built statically from the codebase and dependency manifests; it has not been validated against live traffic capture, and per-service port/protocol detail is outstanding. Open questions are listed inline in §10.5.

## How tested

Documentation only — no code changes. Verified mechanically:
- All 17 referenced repo paths exist.
- All 23 internal anchor links resolve (checked with GitHub's slug rules).
- Image names, MCP server classification (18 fetched / 22 in-repo), dependency counts, and container image list were all derived from the current tree rather than assumed.

Two claims were corrected during drafting after failing verification: MITRE technique data is code-resident in `services/mitre_lookup.py` rather than a bundled refreshable dataset, and the Bifrost RFC1918 behaviour is stated without a version claim the repo doesn't substantiate.

### DIFF
--- a/docs/AIRGAPPED_DEPLOYMENT.md
+++ b/docs/AIRGAPPED_DEPLOYMENT.md
@@ -1,0 +1,648 @@
+# Airgapped Deployment Guide
+
+Deploying Vigil into networks with no route to the public internet.
+
+**Audience:** the platform engineer who will install and operate Vigil inside a
+restricted network. Assumes working knowledge of Docker or Kubernetes; assumes
+no prior knowledge of Vigil.
+
+**Out of scope (this revision):** hardware sizing and capacity planning. Where a
+decision depends on hardware — notably GPU selection for local inference — this
+guide states the dependency and stops.
+
+---
+
+## Table of Contents
+
+- [1. Read this first](#1-read-this-first)
+- [2. Deployment tiers](#2-deployment-tiers)
+- [3. The inference decision](#3-the-inference-decision)
+- [4. Integration inventory: what survives](#4-integration-inventory-what-survives)
+- [5. Software bill of materials](#5-software-bill-of-materials)
+- [6. Getting artifacts across the boundary](#6-getting-artifacts-across-the-boundary)
+- [7. Install: Docker Compose](#7-install-docker-compose)
+- [8. Install: Helm / Kubernetes](#8-install-helm--kubernetes)
+- [9. Offline configuration and known gotchas](#9-offline-configuration-and-known-gotchas)
+- [10. Restricted egress: firewall and ACL reference](#10-restricted-egress-firewall-and-acl-reference)
+- [11. Staying current](#11-staying-current)
+- [12. Acceptance checklist](#12-acceptance-checklist)
+- [13. Known limitations](#13-known-limitations)
+
+---
+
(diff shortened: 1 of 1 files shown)


mattmorr1:
@craig-dt I think long term plan for air-gapped deployment is a full docs, just existing as a series of md files. 
This would look like: instead of vigilsoc.org/docs, it would be a series of documentation inside the repo.

Not to say this document is not helpful, but I think that is more of a priority - increasing transparency across use cases for vigil.

craig-dt:
Closing — opened prematurely, before the author had reviewed the content. Will reopen once the draft has been reviewed internally.

Issue #463, opened 2026-07-29
Case repository seam


nestor-deeptempo:
**Coding:** High · **Vigil context:** High · **Lane:** backend · **Good for:** expert; headline data-layer work

**Context.** Case persistence is the messiest corner of the data layer. Every `case_*` service hand-rolls the same session boilerplate — the `should_close_session = session is None` dance appears **57 times** — and re-implements the same `get_cases`/`search_cases` query builders. The API layer also writes fields the ORM doesn't map: `backend/api/cases.py:390-439` sets `finding_ids`, which `update_case`'s `hasattr` guard silently drops, so case↔finding links only persist against the JSON fallback and are **lost on Postgres** (a live data-loss bug). A single repository seam that owns session/backend selection and returns typed Case aggregates collapses the duplication and fixes that bug as a side effect. Importantly, the deep `DatabaseDataService` already hides demo-mode, auto-reconnect, and the JSON fallback — it's just bypassed by the `case_*` services — so this is **adoption, not a rewrite**.

**Acceptance criteria**
- A Case repository + unit-of-work seam; all `case_*` services route reads/writes through it.
- The 57 duplicated session dances and the duplicate query builders removed.
- Case↔finding association persists on Postgres, with a regression test proving the current silent drop is fixed.

**Files:** `services/case_*_service.py`, `services/database_data_service.py`, `backend/api/cases.py`.

**Blocked by:** #461 (execution-model ADR).

Issue #462, opened 2026-07-29
Swap the 8 sync requests integrations to httpx


nestor-deeptempo:
**Coding:** Med · **Vigil context:** Med · **Lane:** backend · **Good for:** mid-level dev

**Context.** Eight integration clients (SIEM/EDR/enrichment vendors) still call out with the synchronous `requests` library. On the async event loop, each of those calls blocks the entire worker while it waits on a remote API — the same head-of-line problem behind #461. Once the execution model is decided there, this is the mechanical follow-through: convert those eight clients to `httpx` in the chosen style so no integration call stalls the server.

**Acceptance criteria**
- All 8 integrations use `httpx`; no `requests` left in integration clients.
- Timeouts and retry behavior preserved.
- Behavior unchanged (spot-checked against a live or mocked call per integration).

**Blocked by:** #461 (execution-model ADR).

thetosy:
@nestor-deeptempo @mattmorr1 Ill take this

Issue #461, opened 2026-07-29
ADR: choose the execution model — sync-def + threadpool vs. fully async


nestor-deeptempo:
**Coding:** High · **Vigil context:** High · **Lane:** shared · **Good for:** expert, with maintainer sign-off

**Context.** FastAPI is async, but Vigil's data layer is synchronous — plain `create_engine` + `sessionmaker`, zero `AsyncSession` — and a few endpoints do blocking I/O (e.g. S3 at `backend/api/findings.py:30-64`) directly on the event loop. The effect is that one slow query or upload blocks *every* concurrent request on that worker. There are two legitimate fixes — stay sync-`def` and push blocking work to a threadpool, or go fully async with SQLAlchemy 2.0 + `asyncpg` + `httpx` — but they pull the codebase in opposite directions and the choice is expensive to reverse. So it needs a decision (ADR) and a proven reference before the data-layer refactor commits to one; the Case repository (#463) is the first thing that depends on it.

**Acceptance criteria**
- ADR under `docs/adr/` deciding sync-`def` + threadpool **or** async SQLAlchemy 2.0 + `asyncpg` + `httpx`, with rationale.
- One vertical slice (one endpoint end-to-end) lands as the reference implementation of the chosen model.
- The blocking-S3-on-event-loop case is fixed as part of that reference slice.
- Decision recorded as accepted before repository work begins.

The one thing already correct — heavy LLM work is offloaded to the ARQ worker — should be preserved. **Gates #463.**

nestor-deeptempo:
**Resolved via #518.**

The execution-model question is decided and the reference implementation has shipped, so closing this out.

**Decision:** sync-`def` + threadpool (not fully-async SQLAlchemy/`asyncpg`/`httpx`). Fully-synchronous route handlers become plain `def` so FastAPI runs them in the threadpool; the one handler that must also `await` the LLM stays `async` with its DB calls wrapped in `asyncio.to_thread`. Convention: plain `def` is the default, `to_thread` is the exception. Heavy LLM work stays offloaded to the ARQ worker as before.

**Reference implementation:** #518 converted six route modules (`findings`, `attack`, `ingestion`, `jira_export`, `config`, `vstrike`) off the event loop — more than the single vertical slice this issue asked for.

**Blocking-S3-on-event-loop:** fixed — `get_findings` in `backend/api/findings.py` is now plain `def`, so the S3 sync no longer runs on the loop.

[...]

Issue #415, opened 2026-07-23
Typed settings object and lint rules against ambient state


epowell101:
**Track: suggested cleanup.**

Roughly 220 direct `os.getenv`/`os.environ` reads sit inside backend, services, and daemon business logic, and ten-plus services use the `global _service` singleton idiom. Code (and any AI agent editing it) sees the consumer of configuration, never the setter. Goal is to stop the bleeding, not mass-migrate.

**Acceptance criteria**
- A `pydantic-settings` Settings object introduced; daemon config path adopts it first
- Lint rules (ruff/flake8 or import-linter) failing CI on new `os.getenv` in `services/` and `backend/api/`, and on new module-level service instantiation
- Existing violations inventoried in a tracking issue

PR #518, opened 2026-07-30
fix(api): offload blocking I/O off the event loop

nestor-deeptempo:
## What & why

Adopts the sync-`def` + threadpool execution model decided in #461. Async route handlers were calling the synchronous data layer — S3/`boto3`, `requests`, sync SQLAlchemy — directly on the event loop, so one slow call (an S3 sync, a Jira export) blocked every concurrent request on the worker.

## Change

Fully-synchronous handlers become plain `def` so FastAPI runs them in the threadpool; the one handler that also `await`s the LLM (`get_or_generate_enrichment`) stays `async` with its two DB calls wrapped in `asyncio.to_thread`.

| File | Converted |
|---|---|
| `findings.py` | `get_findings` + 5 others → `def`; enrich handler → `to_thread` |
| `attack.py` | 4 handlers |
| `ingestion.py` | 5 S3 handlers |
| `jira_export.py` | 2 (blocking `requests`) |
| `config.py` | `test_s3_connection` |
| `vstrike.py` | `ingest_findings` |

**Convention:** plain `def` is the default; `asyncio.to_thread` is the exception, only for handlers that must also `await`. No behavior change beyond moving blocking work off the loop.

## Out of scope

- `cases.py` (41 handlers) → deferred to #463 (the case-repository refactor reworks that code).
- The `requests` integration clients → #462 (`httpx` swap).
- Automated lint guardrail → left for #415's lint home rather than standing up new infra.

Refs #461

### DIFF
--- a/backend/api/attack.py
+++ b/backend/api/attack.py
@@ -34,7 +34,7 @@ def _parse_finding_timestamp(finding: dict) -> Optional[datetime]:
 
 
 @router.get("/layer")
-async def get_attack_layer():
+def get_attack_layer():
     """
     Get ATT&CK Navigator layer data.
 
@@ -90,7 +90,7 @@ async def get_attack_layer():
 
 
 @router.get("/techniques/rollup")
-async def get_technique_rollup(
+def get_technique_rollup(
     min_confidence: float = 0.0,
     time_range: str = Query("all", pattern="^(24h|7d|30d|all)$"),
 ):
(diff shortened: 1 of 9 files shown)


mattmorr1:
@nestor-deeptempo do you need to remove async imports from all files?

nestor-deeptempo:
@mattmorr1 
a bunch of our route handlers are async but are calling synch blocking I/O i.e. boto3/S3, requests (Jira), sync SQLAlchemy

in an async handler a blocking call never yields to the event loop, so one slow S3 fetch or DB query stalls every other request on that worker

we could move all of them to async but it's kinda a lot of work and touches a lot of files, so i'm just changing the async handlers for now

mattmorr1:
makes sense! we should probably pick one, add to claude.md going forward but for this scope, seems good.

Issue #497, opened 2026-07-29
Unify the ClaudeService and OpenAIAgentService agent loops into one provider-neutral engine


nestor-deeptempo:
> ⏸️ **On hold** — do not start. Folded into the upcoming agent overhaul; picking it up now risks duplicating that work.

**Coding:** High · **Vigil context:** High · **Lane:** shared · **Good for:** expert

**Context.** Two full agent loops exist and do the same job twice — the raw-Anthropic loop in `services/claude_service.py` and the OpenAI-format loop in `services/openai_agent_service.py` (turn loop, tool execution, approval gating, loop detection, backoff) — and they're already drifting. Write the loop **once** as a provider-neutral engine; each provider becomes a thin adapter ("plug") that advertises capabilities, and the loop uses a capability only when the active adapter supports it.

The Anthropic-only quality levers (extended thinking, `cache_control` prompt caching) live *only* in the `ClaudeService` copy today — which is why it can't simply be deleted. In the unified engine, Anthropic's adapter keeps calling the `anthropic` SDK (via the Bifrost `/anthropic` passthrough), so those survive as **negotiated capabilities**, not hardcoded branches. Non-Anthropic adapters just don't advertise them — no regression, since they never had them.

**Acceptance criteria**
- One provider-neutral agent loop; the duplicate loops in `claude_service.py` and `openai_agent_service.py` retired.
- Each provider is a thin adapter that advertises its capabilities; the loop consumes a capability only if the active adapter supports it.
- Anthropic adapter retains extended thinking + prompt caching as negotiated capabilities; non-Anthropic providers unaffected.
- Every safety/tool fix now lands once, and adding a provider is a small adapter — not a loop rewrite.

**Preconditions**
- Revive the test net (#460) — merging two loops on the 3,670-line `ClaudeService` without it is a blind rewrite.
- Converge the two loops' behavior *before* merging code.
- Design the neutral message/event types to carry thinking (blocks, `signature`, `thinking_delta` streaming) + cache hints up front — the thinking seam is the most likely to leak provider specifics back into the "neutral" loop.

**Relates to:** #413 (provider boundary + import contract), #414 (decompose `ClaudeService`), #471 (drop `claude-agent-sdk` — the *third* loop), #485 (`core/llm` package). This is the two-loops → one-loop step those set up.

Issue #471, opened 2026-07-29
Drop claude-agent-sdk; run agents on the neutral loop


nestor-deeptempo:
> ⏸️ **On hold** — do not start. Folded into the upcoming agent overhaul; picking it up now risks duplicating that work.

**Coding:** Med · **Vigil context:** High · **Lane:** shared (coordinate) · **Good for:** someone who knows the LLM harness

**Context.** `claude-agent-sdk` is a third, Anthropic-locked agent loop reached only by the `/claude` and `/agents` endpoints (via the `use_agent_sdk` flag + `_get_agent_sdk_mcp_servers` + `mcp_registry.get_agent_sdk_configs`). Every internal caller — daemon, worker, workflows — already passes `use_agent_sdk=False`, so the daemon's raw tool-use loop is the migration target. This removes **only** `claude-agent-sdk`; the raw `anthropic` SDK and its quality levers (extended thinking, prompt caching) stay.

**Acceptance criteria**
- Grep agent configs for the SDK's built-in Read/Write/Bash tools first (expected: none).
- Delete the SDK branch in `claude_service.py`; repoint `/claude` + `/agents` at the neutral loop via `LLMRouter`.
- Drop `claude-agent-sdk` from `requirements.txt`; remove the `integration_compatibility_service` metadata row.

**Files:** `services/claude_service.py`, `backend/api/claude.py` + `agents.py`, `requirements.txt`.

**Relationship to existing issues:** complements the provider-boundary work (#413, #449, #450, #325–328, #445), which routes callers off `ClaudeService` internals but does **not** remove the `claude-agent-sdk` dependency — that removal is this issue.

mattmorr1:
done

Issue #414, opened 2026-07-23
Decompose claude_service.py along its three responsibilities


epowell101:
**Track: suggested cleanup. Depends on the LLM provider boundary issue.**

At 3,670 lines, `services/claude_service.py` holds three responsibilities in one body: context management (summarization, folding, history windows), tool dispatch (backend, MCP, mixed), and transport (chat, streaming, image blocks). Its public interface is admirably narrow; the body has outgrown coherence.

**Acceptance criteria**
- Three modules with the existing public interface preserved behind the router
- No resulting file exceeds 800 lines
- Existing `test_claude_service.py` coverage migrated and passing without weakened assertions
- Mock density reduced where tests pinned internals that no longer exist

PR #522, opened 2026-07-31
refactor(llm): consolidate the LLM layer into core/llm/

### LINKED ISSUE #485: Reorg R4: core/llm/ package (consolidate ~13 files)
**Coding:** High · **Vigil context:** High · **Lane:** shared · **Good for:** expert

Part of #481.

**Context.** Consolidate the ~13 LLM files into a `core/llm/` package and establish the harness boundary: `claude_service`, `llm_router`, `llm_clients`, `llm_format`, `llm_gateway`, `llm_worker`, model registry, `openai_agent_service`, ollama, bifrost, cost, `prompt_security`.

**Acceptance criteria**
- LLM code lives under `core/llm/` (router/harness/providers/cost/bifrost sub-packages).
- The harness boundary is defined; `git mv` + update all callers in the same PR (**no shims**; fail closed — a stale import errors in CI); no behavior change.

**Pairs with:** #414 (ClaudeService decompose).

**Reorg-only note:** the agent-loop *rewrite* (#471 / #497) is on hold — this issue is a pure relocation (`git mv` + in-PR caller updates, no shims), not a rewrite.

mattmorr1:
Closes #485. reorg r4

`services/` held 16 flat LLM modules (~11k lines) with no stated layering: two
agent harnesses sat next to the single-call router, the Bifrost admin client,
the model registry, the ARQ worker and the cost math — all peers in one
directory. Nothing said which layer could call which.

This is a **relocation, not a rewrite**. `git mv` + shims, no behaviour change.
The agent-loop rewrite (#471 / #497) stays on hold; the `ClaudeService`
decompose is still #414.

## The harness boundary

The one design deliverable, in `core/llm/README.md` and enforced by
`tests/llm/test_boundary.py`. The distinction is **loop vs. call**: a router
call is one stateless completion; a harness runs many, executing tools between
turns.

| Sub-package | Owns |
|---|---|
| `router/` | One stateless completion: provider selection, wire-format translation, pre-dispatch sanitization |
| `harness/` | The multi-turn agent loop: tool execution via MCP, approval gating, streaming |
| `providers/` | SDK clients, model registry, live discovery, local Ollama supervision |
| `cost/` | Pricing math, pre-call estimation, virtual-key budget enforcement |
| `bifrost/` | The only place that speaks Bifrost's admin + logging REST APIs |
| `gateway/` | ARQ enqueue/worker transport in front of the router |

`gateway/` is a sixth sub-package not named in the issue — `llm_gateway` and
`llm_worker` are queue transport, neither router nor harness, and forcing them
into either would erase the distinction the package exists to make.

The ratchet enforces two rules: the router may not import the harness at **any**
scope, and nothing under `core/llm/` may import `backend`/`daemon` at module
scope. Verified by injecting both violations and confirming failure.

| Commit | Scope |
|---|---|
| `fcdd9ba` | leaves: clients, format, security, discovery |
| `311dcba` | providers + Bifrost + cost |
| `8a2373b` | router |
| `e9f8992` | gateway + worker + 4 deploy entrypoints |
| `7d02a13` | harness |
| `64c6d97` | de-shim + boundary ratchet |

All 16 compatibility shims are deleted — no consumer still routes through them,
and zero LLM modules remain in `services/`.

fixed issues associated with the following:

1. **`core/secrets.py` put `backend/` on `sys.path[0]` process-wide**, so
   `backend/tools/` shadowed the root `tools/` package. `claude_service`
   happened to undo it by re-fronting the repo root at import time. It only
   ever needed the repo root.
2. **`backend/tools/` → `backend/maintenance/`**, so the collision cannot
   recur. It is the only hard collision (`backend/services/` merges harmlessly
   as a namespace package) and had no Python importers — it is invoked via
   `python -m`.
3. **Repo-root paths computed by directory depth**, silently wrong the moment a
   module moves. `core.config.REPO_ROOT` is now public; the MCP tools-cache
   lookup and the SPA mount use it.

Five cache tests drove the old `Path(__file__).parent.parent` expression through
a mock chain, so they now patch `REPO_ROOT` — they had to move regardless of
which fix was chosen.

## Scope calls worth a second opinion

- **`compute_call_cost` extracted** from `daemon/agent_runner.py` into
  `core/llm/cost/calls.py`. A 14th file, outside the issue's list — but it is
  the only runtime reason both harnesses import the daemon, so leaving it made
  the boundary decorative. Easy to revert to a re-export if you'd rather keep
  #485 strictly to the named files.
- **One grandfathered violation**: `harness/claude.py` reads
  `backend.schemas.tool_schemas` at module scope behind a `try/ImportError`.
  Making it lazy is a behaviour change, so it is named in the ratchet allowlist
  and in the README, and belongs to #414.


nestor-deeptempo (approved):
👍

Issue #483, opened 2026-07-29
Reorg R2: one vendor slice → core/integrations/splunk/ + _base


nestor-deeptempo:
**Coding:** Med · **Vigil context:** High · **Lane:** backend · **Good for:** Vigil-savvy dev — establishes the pattern the other vendors copy

Part of #481.

**Context.** Move one vendor end-to-end into a self-contained vertical slice at `core/integrations/splunk/` — bundling client·ingestion·enrichment·tool·adapter·descriptor·router — and stand up `core/integrations/_base/` plus the integration descriptor that later vendors derive their registries from.

**Acceptance criteria**
- `core/integrations/splunk/` holds the full Splunk slice; `_base/` + descriptor created.
- Splunk's registry entries derive from its descriptor.
- `git mv` + update all callers in the same PR (**no shims**; fail closed — a stale import errors in CI); no behavior change.

**Note (no-shim decision):** the slice originally shipped with two re-export shims — `services/splunk_service.py` (removed in #527) and `daemon/federation/adapters/splunk.py` (removed in #528); both callers are repointed to `core.integrations.splunk.*`.

Issue #481, opened 2026-07-29
[Epic] File-structure reorg → clients/ + services/ + core/ + infra/


nestor-deeptempo:
**Context.** Umbrella for the target physical structure: two front-end clients (`clients/web`, `clients/desktop`), three backend services (`services/{api,daemon,worker}`), and the shared `core/` library, with deploy machinery under `infra/`. Each child below is **independently shippable** via `git mv` + in-PR caller updates — **no shims**; git history survives via the clean `git mv` — with **no behavior change**. They're listed in dependency order.

Relates to #418 (vertical-slice convention ADR) and #225 (`docs/adr/`).

### Children (dependency order)
- [ ] #482 — agents → `core/agents/` (best PoC)
- [ ] #507 — `clients/` split (`frontend/` → `clients/web/`, `desktop/` → `clients/desktop/`) — independent; parallelizable
- [ ] #483 — one vendor slice → `core/integrations/splunk/` + `_base`/descriptor
- [ ] #484 — remaining vendors → `core/integrations/<vendor>/` (template, ~11)
- [ ] #485 — `core/llm/` package (consolidate ~13 files)
- [ ] #486 — mechanical domain regroupings
- [ ] #487 — `infra/` consolidation (docker + helm + database/init)
- [ ] #488 — app-shell split — thin `services/{api,daemon}` + router auto-discovery
- [ ] #508 — `services/worker/` — extract the ARQ llm-worker into its own thin service (blocked by #485; pairs with #488)
- [ ] #489 — final audit: verify a single `core/` namespace with no stale imports or leftover shims

**Guiding constraints:** every step independently shippable; moves are `git mv` + in-PR caller updates (**no shims**; history survives via the clean `git mv`); a moved module's callers are repointed in the same PR so a stale old-path import fails closed (loud CI error); runtime behavior unchanged except explicit bug fixes.

Goal file structure

<img width="563" height="961" alt="Image" src="https://github.com/user-attachments/assets/fcdad312-9ae9-4442-8a96-40823f47e0ec" />

PR #527, opened 2026-08-01
reorg(splunk): drop the services.splunk_service shim

nestor-deeptempo:
Removes the compatibility shim left by the Splunk slice (#483) and points the sole caller (`daemon/poller.py`) at `core.integrations.splunk.client` directly.

Part of the decision to drop import shims across the reorg (#481): update callers in the same PR and **fail closed** — a stale old-path import now errors in CI instead of silently resolving through a re-export.

**Changes**
- Delete `services/splunk_service.py` (6-line re-export shim)
- Repoint `daemon/poller.py` (the only importer of the old path) to `core.integrations.splunk.client`

**Verification**
- No `services.splunk_service` references remain anywhere in the tree
- `daemon/poller.py` compiles

Ref #481. Complements the no-shim policy update applied to the reorg issues.

### DIFF
--- a/daemon/poller.py
+++ b/daemon/poller.py
@@ -110,7 +110,7 @@ class DataPoller:
             # Initialize Splunk service if configured
             if is_integration_enabled('splunk'):
                 try:
-                    from services.splunk_service import SplunkService
+                    from core.integrations.splunk.client import SplunkService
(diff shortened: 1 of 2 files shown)


mattmorr1:
lgtm

Issue #464, opened 2026-07-29
Unit-of-work transaction boundary (one per request)


nestor-deeptempo:
**Coding:** High · **Vigil context:** High · **Lane:** backend · **Good for:** expert; pairs with the Case repository

**Context.** There's no transaction boundary per request. Services scatter **74 `.commit()`** and **57 `.rollback()`** calls wherever convenient, so a multi-step operation that fails halfway can leave the database partially written — some rows committed, others not. This issue establishes one unit-of-work boundary per request/operation, built on the seam from #463, so multi-step writes are all-or-nothing.

**Acceptance criteria**
- One transaction boundary per request/operation; scattered `commit`/`rollback` removed in favor of the UoW.
- A representative multi-step operation is demonstrably atomic (rolls back fully on a mid-operation failure) with a test.

**Files:** service methods across `services/`, wired to the UoW seam.

**Blocked by / pairs with:** #463.

PR #529, opened 2026-08-03
Add Case repository seam and unit-of-work (#463)

### LINKED ISSUE #463: Case repository seam

G-r-ay:
## What & why

Closes #463. Case persistence was the messiest corner of the data layer: every `case_*` service hand-rolled the same session boilerplate (the `should_close_session = session is None` dance appeared **57 times**), the case query builder was implemented twice, and a live data-loss bug silently dropped case↔finding links on Postgres.

This introduces a repository + unit-of-work seam and routes the case services through it.

## Changes

- **`database/case_repository.py`** — a single case query builder that consolidates the two divergent implementations (`DatabaseService.get_cases` and `CaseSearchService.search_cases`), plus the case↔finding link logic. Operates on a caller-provided session.
- **`services/unit_of_work.py`** — one context manager owning session lifecycle with a real commit/rollback boundary. When a caller passes a session it joins that transaction (no commit/close); otherwise it owns one. Replaces the 57-dance, which only ever *closed*.
- **`database/service.py`** — `update_case` now resolves `finding_ids` into the `findings` relationship (mirroring `create_case`) instead of the `hasattr` loop silently dropping it. **Fixes the case↔finding data-loss bug on Postgres.**
- **Migrated all 57 `should_close_session` dances** across the 8 `case_*` services to the seam, folding scattered inline `commit()`/`rollback()` into the transaction boundary. Net **−517 lines**.

## Tests

20 new unit tests (run in normal CI, no Postgres needed):
- `test_case_finding_link.py` — regression test proving `finding_ids` now persists (was silently dropped).
- `test_unit_of_work.py` — commit/rollback/close ownership vs. joining a passed session.
- `test_case_repository.py` — consolidated builder + finding-link.
- `test_case_collaboration_seam.py` — a migrated service threads the seam correctly.

## Notes / follow-ups (out of scope here)

- **Pre-existing latent bug:** `Case.tags` is a base `ARRAY(String)`, whose `.contains()` is unimplemented — so `search_cases` (old *and* new) raises `NotImplementedError` on any `tags`/`mitre_techniques` filter. Preserved behavior; worth its own issue.
- **`case_automation_service.py`** uses direct `get_db_session()` in background-task loops (not one of the 57 dances, a different pattern) — left for the unit-of-work-per-request work in #464.


craig-dt:
Heads-up before this lands. The `case_notification_service.py` rewrite carries one
pre-existing bug forward, and it collides with two small open PRs of mine. Both are
merge-order points rather than objections to the seam — `unit_of_work` reads well and
killing the 57-copy session dance was worth doing.

## 1. The rewrite preserves the #559 data-loss bug

`services/case_notification_service.py:65` on this branch, inside the rewritten
`create_notification`:

```python
notification = CaseNotification(
    ...
    priority=priority,
    metadata=metadata or {},
```

`CaseNotification` has no `metadata` column — it was deliberately renamed
(`database/models.py:2279-2280`):

```python
# Metadata (renamed from 'metadata' to avoid SQLAlchemy reserved word conflict)
notification_metadata: Mapped[Optional[dict]] = mapped_column(JSONB, nullable=True)
```

Passing `metadata=` to a declarative constructor neither raises nor writes a column:
it shadows SQLAlchemy's `MetaData` on the instance and is discarded. The insert
commits, returns a row, and logs success with the payload gone. Since `to_dict()`
reads `self.notification_metadata` (`database/models.py:2312`), the API then reports
`"metadata": null` for every notification this service creates — including the
`assigned_by` and comment-mention payloads passed at `:115`, `:161` and `:209` of this
branch's version.

**This is not introduced here.** It is pre-existing on `main`
(`services/case_notification_service.py:68`); this PR faithfully preserved the line it
found. #559 tracks it and #563 fixes it with `notification_metadata=metadata or {}`.

I'm raising it on this PR because the rewrite *re-types* that line inside a
restructured block. Resolving the #563 conflict by "take the newer refactor" silently
reinstates the `NULL`, and nothing fails to signal it.

## 2. Conflicts if this lands first

Both reproduce from a clean checkout:

```
git fetch origin main refs/pull/529/head:pr529 \
    refs/pull/563/head:pr563 refs/pull/558/head:pr558
T=$(git merge-tree --write-tree origin/main pr529)
C=$(git commit-tree $T -p origin/main -p pr529 -m trial)
git merge-tree --write-tree --name-only $C pr563
# CONFLICT (content): Merge conflict in services/case_notification_service.py
git merge-tree --write-tree --name-only $C pr558
# CONFLICT (content): Merge conflict in services/case_notification_service.py
```

| PR of mine | Overlap |
|---|---|
| #563 | `create_notification` — the kwarg above |
| #558 | `notify_watchers` — this PR rewrites its session handling at `@@ -282,15 +250,11 @@`; #558 inserts a vocabulary guard at `@@ -282,6 +299,18 @@`, directly above the `should_close_session` block this PR removes |

**I'd rather this PR go first.** A repository/unit-of-work seam shouldn't queue behind
a one-line kwarg fix — I'll rebase #558 and #563 on top and keep both fixes intact.
Say the word if you'd prefer the opposite order for either.

## 3. One thing that is *not* a problem here

Noting this explicitly because I flagged the opposite on #566: this PR's
`MERGEABLE / CLEAN` status is **accurate**. Its merge base is `d854e70`, which
predates both #518 and #502, but it doesn't touch `backend/api/findings.py`, so the
stale-base conflict I hit on #566 doesn't apply here — `git merge-tree origin/main
pr529` comes back clean. #566 sits one commit on top of this branch, which is why the
notification flag appears on both; this is the PR that merges to `main`, so the
resolution decision lands here.

---

*Analysis by Claude (Claude Code), reviewed by @craig-dt before posting. All
file:line citations were verified against `main` at `311790e` and
`refs/pull/529/head`; both conflicts were reproduced locally with the commands above.*

Issue #532, opened 2026-08-03
test_route_auth_coverage.py is vacuous: examines 1 route of 359, always passes


craig-dt:
**Coding:** Med · **Vigil context:** Med · **Lane:** backend / testing · **Good for:** someone comfortable with FastAPI internals

**Context.** `tests/security/test_route_auth_coverage.py` is the route-inventory check that
locks in the deny-by-default contract at `/api/*` — added after the 2026-05 disclosure. It is
**vacuous**: it passes unconditionally, regardless of how any router is mounted.

The test iterates `app.routes` and keeps entries whose `path` starts with `/api/`. Under
FastAPI **0.137.1** / Starlette **1.3.1**, `include_router` no longer flattens child routes
into `app.routes` — it stores lazy `_IncludedRouter` objects, which have **no `.path`
attribute at all**. `app.routes` now contains 40 of those plus 2 real routes, so the filter
discards everything and the loop examines exactly **one** route, `/api/health`, which is on
`PUBLIC_API_PATHS` and therefore skipped.

`missing` is always empty. The assertion can never fire.

```
routes the test actually examines: 1
    /api/health
_IncludedRouter count: 40    has .path? False

effective /api/* routes that SHOULD be examined: 359
```

`requirements.txt` pins `fastapi>=0.109.0` with no upper bound, so a routine transitive
upgrade disarmed the gate silently. (Observed broken at 0.137.1; the exact version that
introduced lazy inclusion has not been bisected.)

**The auth boundary itself is currently intact.** A corrected audit found 359 effective
`/api/*` routes: 351 require auth, 7 are un-authed and all 7 are on `PUBLIC_API_PATHS`
(login, refresh, bootstrap, both password-reset legs, health, vstrike `/findings`).
**0 unprotected.** This issue is about the loss of regression protection, not a live
vulnerability — the next change that drops auth from a router would ship green.

**Proposed fix.**

*1. Repair the traversal* — recurse through `_IncludedRouter`, using the full effective path
and inheriting mount-site dependencies. Verified working against the live app:

```python
def visit(obj, inherited_auth, acc):
    if type(obj).__name__ == "_IncludedRouter":
        ictx = obj.include_context
        mount_auth = any(
            d.dependency in AUTH_DEPS
            for d in (getattr(ictx, "dependencies", None) or [])
        )
        for cand in obj.effective_candidates():
            visit(cand, inherited_auth or mount_auth, acc)
        return
    path = getattr(obj, "path", None)
[...]

PR #533, opened 2026-08-03
ci: run tests/security/ on pull requests

### LINKED ISSUE #531: tests/security/ never runs in PR CI (84 security tests outside the merge gate)
**Coding:** Low · **Vigil context:** Low · **Lane:** CI · **Good for:** small, high-value first PR

**Context.** `tests/security/` holds 4 files and 84 tests — including the deny-by-default
assertions added after the 2026-05 disclosure — and **no pull-request CI job runs any of
them.** `ci-cd.yml` runs `tests/unit/` and `tests/integration/` only, both path-scoped, so
the whole directory is invisible to the merge gate. The only job that executes it is
`nightly.yml` (`pytest tests/`, 02:00 UTC) — i.e. after merge, when a regression has already
landed on `main`.

What is currently unguarded on every PR:

| File | Tests | Asserts |
|---|---|---|
| `test_unauth_endpoints.py` | 40 | unauthenticated requests to the disclosure's routes return 401/403; authenticated non-admins get 403 |
| `test_path_traversal.py` | 22 | integration IDs reject `../` traversal |
| `test_ssrf.py` | 21 | provider URLs reject private/link-local targets |
| `test_route_auth_coverage.py` | 1 | every `/api/*` route requires auth or is on `PUBLIC_API_PATHS` |

All 84 pass today in ~2s, so this is a pure gate-coverage gap — there is nothing to repair
first and no failures to triage.

**Proposed fix.** One step in the existing `test-unit-backend` job in `ci-cd.yml` (deps are
already installed there; a 2s suite doesn't justify a separate job):

```yaml
- name: Run security tests
  run: pytest tests/security/ -v --no-cov
  env:
    TESTING: "true"
    DEV_MODE: "false"
    JWT_SECRET_KEY: "ci-only-not-a-real-secret"
```

The explicit `env:` block is **required, not defensive.** `backend/middleware/auth.py:24`
defaults `DEV_MODE` to `false`, and with no `.env` in CI `JWT_SECRET_KEY` is unset, so
`_load_jwt_secret()` raises. `test_route_auth_coverage.py` never sets it — that file only
works today because `test_unauth_endpoints.py` calls `os.environ.setdefault("JWT_SECRET_KEY", ...)`
at *collection* time while the coverage test imports `backend.main` lazily inside its test
body. CI must not depend on that accidental cross-file ordering.

Add it as its own step rather than widening the `pytest tests/unit/` path, so a security
failure is legible in the job log instead of buried among ~1000 unit tests.

**Acceptance criteria**
- `tests/security/` runs on every pull request to `main`/`develop`.
- The step sets `DEV_MODE` and `JWT_SECRET_KEY` explicitly and passes in a clean CI
  environment (no `.env`).
- Security failures are attributable to their own named step in the job log.
- No test *content* changes in this PR; nothing else in CI changes.

**Out of scope.** `test_route_auth_coverage.py` is currently **vacuous** — FastAPI 0.137.1
changed `include_router` to store lazy `_IncludedRouter` entries with no `.path`, so its
`path.startswith("/api/")` filter examines 1 route out of 359 and passes unconditionally.
That is a separate defect getting its own issue and PR; this issue only makes the directory
run. The other 83 assertions become real immediately.

For the record, the auth boundary itself is **currently intact** — an out-of-band audit found
359 effective `/api/*` routes, 351 requiring auth, 7 un-authed and all 7 on
`PUBLIC_API_PATHS`, 0 unprotected. What is missing is regression protection, not security.

**Note.** Gates [#478](https://github.com/Vigil-SOC/vigil/issues/478) (router
auto-discovery), which re-mounts all 42 routers and can only be shown safe by the
route-inventory check running and working.

craig-dt:
## What

Adds a `Run security tests` step to the existing `test-unit-backend` job so `tests/security/` runs on every pull request.

## Why

`tests/security/` holds **84 tests across 4 files** — including the deny-by-default assertions added after the 2026-05 disclosure — and **no pull-request job ran any of them**. `ci-cd.yml` runs `tests/unit/` and `tests/integration/` only, both path-scoped, so the whole directory was visible solely to `nightly.yml` (`pytest tests/`, 02:00 UTC) — after a regression had already merged.

| File | Tests | Asserts |
|---|---|---|
| `test_unauth_endpoints.py` | 40 | unauthenticated requests to the disclosure's routes return 401/403; authenticated non-admins get 403 |
| `test_path_traversal.py` | 22 | integration IDs reject `../` traversal |
| `test_ssrf.py` | 21 | provider URLs reject private/link-local targets |
| `test_route_auth_coverage.py` | 1 | every `/api/*` route requires auth or is on `PUBLIC_API_PATHS` |

## Decisions worth reviewing

**Its own step, not a widened path.** Running `pytest tests/security/` separately rather than extending `pytest tests/unit/` keeps a security failure legible in the job log instead of buried among ~1000 unit tests.

**`DEV_MODE` and `JWT_SECRET_KEY` are explicit, not defensive.** `backend/middleware/auth.py:24` defaults `DEV_MODE` to `false`, and with no `.env` in CI an unset `JWT_SECRET_KEY` makes `_load_jwt_secret()` raise. `test_route_auth_coverage.py` never sets it — that file currently only works because `test_unauth_endpoints.py` calls `os.environ.setdefault("JWT_SECRET_KEY", ...)` at *collection* time while the coverage test imports `backend.main` lazily inside its test body. CI must not depend on that accidental cross-file ordering.

**`--no-cov`.** Stops the step from overwriting the `coverage.xml` that the Codecov step immediately below uploads.

## Scope

Test *content* is unchanged; this only makes existing tests run. Nothing else in CI changes.

Note that `test_route_auth_coverage.py` is itself **vacuous** under FastAPI 0.137.1 — it examines 1 route out of 359 and passes unconditionally — so it will pass without asserting anything until #532 lands. The other 83 assertions become real immediately, which is why this is worth landing on its own.

For the record, the auth boundary is **currently intact**: 359 effective `/api/*` routes, 351 requiring auth, 7 un-authed and all 7 on `PUBLIC_API_PATHS`, 0 unprotected. This restores regression protection; it is not a live-vulnerability fix.

## Testing

```
$ env TESTING=true DEV_MODE=false JWT_SECRET_KEY=ci-only-not-a-real-secret \
    pytest tests/security/ -v --no-cov
84 passed, 5 warnings in 7.64s
```

Also verified: the workflow YAML parses, the new step lands between `Run unit tests` and `Upload coverage to Codecov`, and `--no-cov` leaves `coverage.xml` untouched.

Clean-environment behaviour (no `.env`) is proven by this PR's own CI run — that could not be verified locally without moving the developer's `.env`.

One incidental change: the blank line above the new step previously held trailing whitespace and is now genuinely blank.

Closes #531

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -126,7 +126,23 @@ jobs:
           TESTING: "true"
           CLAUDE_API_KEY: ${{ secrets.CLAUDEAPI }}
           ANTHROPIC_API_KEY: ${{ secrets.CLAUDEAPI }}
-      
+
+      # tests/security/ is path-scoped out of the unit and integration jobs, so
+      # until now nothing ran it on a pull request — only nightly.yml, i.e. after
+      # merge. Its own step (rather than widening the path above) keeps a security
+      # failure legible instead of buried in ~1000 unit tests.
+      #
+      # DEV_MODE and JWT_SECRET_KEY must be explicit: backend/middleware/auth.py
+      # defaults DEV_MODE to false, and with no .env in CI an unset JWT_SECRET_KEY
+      # makes _load_jwt_secret() raise. --no-cov keeps this step from overwriting
+      # the coverage.xml the Codecov step below uploads.
+      - name: Run security tests
+        run: pytest tests/security/ -v --no-cov
+        env:
+          TESTING: "true"
+          DEV_MODE: "false"
+          JWT_SECRET_KEY: "ci-only-not-a-real-secret"
+
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v5
         with:


mattmorr1:
Superseded: `tests/security/` already runs on every pull request. Main's `ci-cd.yml` widened the unit job's path rather than adding a step —

```yaml
# tests/security/ used to run only in nightly, where its one red
# test was mistaken for a real auth bypass; it was the fixture not
# forcing DEV_MODE off for routers that read settings live. Fixed,
# so all 84 gate here.
pytest tests/unit/ tests/security/ -m "not external_service" -v
```

so merging this would add a second step running the same 84 tests twice. The rationale in the body has also aged out: it reasons about `backend/middleware/auth.py` and `backend.main`, which the `services/api/` + `core/` reorganisation removed.

Closing rather than rebasing. Reopen if the duplicate-step framing is still wanted for log legibility.

PR #535, opened 2026-08-03
refactor(api): auto-discover routers via ROUTER_META

### LINKED ISSUE #478: Router auto-discovery (replace 42 manual include_router calls)
**Coding:** Med · **Vigil context:** Med · **Lane:** backend · **Good for:** a strong coder new to Vigil; self-contained

**Context.** 42 manual `include_router` calls in `main.py` plus a hand-maintained re-export list — adding a route means editing central files everyone touches. Replace with a `pkgutil` scan (+ optional `ROUTER_META`) that mounts every router automatically.

**Acceptance criteria**
- Routers auto-discovered and mounted; the manual `include_router` list removed.
- Adding a router needs no `main.py` edit.

**Files:** `backend/main.py`, `backend/api/__init__.py`.

**Note:** lands independently of the folder move; #488 (app-shell split) then wires it into the new `services/api` layout.

craig-dt:
## What

Replaces the 42 hand-written `include_router` calls in `backend/main.py`, plus the partial re-export list in `backend/api/__init__.py`, with a `pkgutil` scan. Adding a router now means adding a module — no edit to `main.py`.

`main.py` loses **314 lines**: ~70 router imports and the entire mount block become one call.

```python
mount_routers(app, context_path=_CONTEXT_PATH, auth_dependency=AUTH_DEPENDENCY)
```

Each module declares its own mount metadata:

```python
ROUTER_META = RouterMeta(
    prefix="/api/findings",
    tags=["findings"],
    auth=Auth.REQUIRED,
)
```

New files: `backend/api/_meta.py` (the `Auth` enum + `RouterMeta`, a leaf module so the 42 routers can import it without a cycle back to `main`), `backend/api/_discovery.py` (the scan), `tests/unit/test_router_discovery.py`.

## Verification: exact route parity

The risk in this change is a route quietly moving, losing auth, or losing a feature gate. So parity is measured, not argued. A harness dumped path + method + auth + rate-limit + tags + allowlist status for every effective route before and after:

| | before | after |
|---|---|---|
| total routes | 368 | 368 |
| `/api/*` routes | 359 | 359 |
| requiring auth | 351 | 351 |
| rate-limited | 10 | 10 |

**Zero diff** across all 368 routes. Feature gates confirmed off by default, and confirmed to open under `DARKTRACE_ENABLED=true`.

> ⚠️ **`tests/security/test_route_auth_coverage.py` cannot corroborate the auth half of this.** It is vacuous under FastAPI 0.137.1 — it examines 1 route out of 359 and passes unconditionally (#532). The route-inventory diff above is what establishes auth parity here. Please don't read that test's green tick as evidence.

## Design decisions

**`ROUTER_META` is mandatory, with no convention fallback.** Filename-inferred prefixes are wrong for **21 of 42** modules — `ai_config`→`/api/ai`, `case_metrics`→`/api/cases/metrics`, `llm_providers`→`/api/llm/providers`, `ingestion`→`/api/ingest`, and 17 more. Not a tunable failure rate; exactly half, and the failure mode is silent. A module without `ROUTER_META` is a startup error.

**`auth` is a 3-value enum, not a bool**, and is stated explicitly in all 42 modules with no default — so any module's posture is readable locally, and `grep -c "auth=Auth."` is a complete audit.

| Value | Count | Meaning |
|---|---|---|
| `REQUIRED` | 37 | attach the shared authenticated-user dependency |
| `ROUTER_MANAGED` | 3 | router enforces auth internally — don't attach |
| `PUBLIC_WEBHOOK` | 2 | inbound machine caller; HMAC/key at the endpoint |

**Deviating from `REQUIRED` requires a written `reason`,** enforced by `__post_init__` in both directions: non-`REQUIRED` without a reason raises, and `REQUIRED` *with* one also raises so the field can't drift into a notes box. Weakening auth is the one change here that should be impossible to make quietly. `PUBLIC_WEBHOOK` additionally cannot be constructed without an `enabled` gate — an inbound receiver can't be exposed by default even by accident.

**Mount order is alphabetical, with no `order` field.** Measured rather than assumed: of 30 param-vs-literal path shadows in the tree, all 30 sit *within* a single router (`/api/findings/{finding_id}` vs `/api/findings/all`), where order comes from decorator order and discovery never touches it. **Zero cross-router shadows**, so order is genuinely irrelevant and an `order` field would be hand-maintained bookkeeping — the exact thing this removes. `test_no_cross_router_path_shadowing` fails the day that changes, naming both routers. Its detector was checked against 6 cases including an arity mismatch, so it discriminates rather than trivially passing.

## Also in here

- **Two feature gates move out of `main.py`** to sit with the receivers they guard. `darktrace_enabled()` is new in `api/darktrace_webhook.py`; cloudflare already had `cloudy_ingestion_enabled()`.
- **`api/kafka.py` declared prefix/tags in its `APIRouter()` constructor** — the sole exception, and the reason its mount passed no `tags`. Now in `ROUTER_META` like everything else, so there is one home for mount metadata.
- **`api/__init__.py`'s re-export list is gone.** It covered only 19 of 42 modules (`main.py` imported the other 23 directly), so it was a second inconsistent convention kept in sync by hand — and its eager imports meant one bad module broke the whole package. No consumers outside `main.py`.
- **Six comments carrying non-obvious rationale move with their routers** rather than being deleted with the mount block: budgets deliberately sharing analytics' prefix, claude needing rate limiting *on top of* auth, webhook management vs inbound receivers, and the auth router's deliberate public/private mix.
- `main.py`'s `E402`/`F401` count drops from 39 to 14, purely from deleting those imports.

## Testing

```
pytest tests/unit/ -m "not external_service"   -> 1072 passed, 13 failed
pytest tests/security/                          -> 84 passed
pytest tests/unit/test_router_discovery.py      -> 13 passed
```

The 13 failures are **pre-existing** — `test_pr_d_mechanical`, `test_prompt_caching`, `test_workflow_skill_wiring`, all unrelated to routing. I confirmed the identical 13 fail on untouched `main` in a clean worktree at `311790e`. Unit count rises 1059 → 1072 from the new tests.

Black and flake8 clean on everything authored here. All 42 `ROUTER_META` blocks were verified byte-identical to Black's output rather than reformatting whole files — an early pass of `black backend/api/` rewrote 40 files / ~2000 lines of pre-existing code, which was reverted so the diff stays reviewable. Each module's change is 8 lines.

## Merge order

Independent of, but best landed after, #533 (runs `tests/security/` in PR CI) and #534 (repairs the vacuous route-auth test). Once #534 is in, that test can corroborate the auth parity claimed above instead of merely passing. This PR's parity evidence stands either way.

## Assignment

#478 was assigned to @thetosy, who offered to own it on 2026-07-29. No code was pushed in the interim and we picked it up to unblock next sprint — see the note on the issue. @thetosy, apologies for the overlap; happy to hand over anything useful here, and the design rationale is all in the commit bodies.

Closes #478


nestor-deeptempo:
sorry to take over, this was blocking a pr i wanted to get out before the morning and i wasn't sure what the status of this was, your work was good!

Issue #411, opened 2026-07-23
Adopt Alembic migrations; retire create_all as the schema mechanism


epowell101:
**Track: suggested cleanup (gates Hunt Objects).**

Schema evolution today is `Base.metadata.create_all(checkfirst=True)` (`database/connection.py`, `backend/main.py`). It creates missing tables and never alters existing ones, so any change to an existing table silently diverges across deployments: fresh installs work, upgrades break at runtime. The planned Hunt Object schema will iterate, making a migration framework a hard prerequisite.

**Acceptance criteria**
- Alembic initialized, autogenerate configured against `database/models.py`
- Baseline migration capturing the current schema
- CI job failing when models and migrations diverge
- `create_all` retained only for ephemeral test databases
- Upgrade path documented in `docs/`

nestor-deeptempo:
Heads-up on scope the current acceptance criteria miss — the schema isn't created by `create_all` alone; there are **two layered mechanisms**, and Alembic has to reconcile both:

1. **The 16 `database/init/*.sql` files** are the primary schema, run at Postgres initdb: docker-compose bind-mounts `../database/init` → `/docker-entrypoint-initdb.d` (**lexicographic** order), while the Helm `db-init-job.yaml` runs them in `values.yaml`'s `dbInit.sqlFiles` order behind the `_vigil_schema_versions` marker table. The two deploy paths order the files **differently** — a live trap.
2. **`Base.metadata.create_all(checkfirst=True)`** *also* runs at app startup (`backend/main.py:749`, `database/connection.py:634`). It only creates missing tables, never alters, and can report success on a stale schema (see `backend/api/storage_status.py`).

Two additions I'd suggest to the acceptance criteria:

- **Don't baseline Alembic from `models.py` alone** — the deployed schema comes from the SQL files, so autogenerating from the ORM will silently bake in any drift. Baseline against what the SQL files actually produce (or reconcile them first).
- **Cut over against *both* deploy paths** — a migration clean under docker-compose can still break the Helm dbInit Job (or vice-versa) due to the ordering difference. Verify a fresh `docker compose up` and `helm template` both land on the same schema post-Alembic.

nestor-deeptempo:
Worth covering in the upgrade-path docs: reference data seeded through `database/init/` has no repricing path today, and autogenerate won't produce one, since it diffs schema and never values.

[...]

PR #537, opened 2026-08-03
refactor(findings): extract AI enrichment into a module (#470)

### LINKED ISSUE #470: Finding-Enrichment module
**Coding:** Med · **Vigil context:** Med · **Lane:** backend · **Good for:** mid-level dev

**Context.** The ~330-line `findings.py` handler inlines the whole AI-enrichment flow — provider resolution, prompt building, and response parsing — so "enrich a finding with AI analysis" can only happen from that one HTTP path; ingestion and agents can't reuse it. Lift it into a standalone module with one seam so all three callers share it.

**Acceptance criteria**
- Enrichment extracted into a module with a single `enrich(finding)` seam.
- The `findings.py` handler, ingestion, and agents call it instead of duplicating the flow.

**Files:** `backend/api/findings.py` → a `findings/enrichment` module.

**Relationship to existing issues:** this is *AI/LLM* enrichment of a finding — distinct from the *threat-intel IOC* enrichment pipeline stage in the `tdd:data-layer-v1` track (#253 "Stage 5: VT/Shodan/OTX fanout", epic #211). The two are kept separate on purpose.

thetosy:
Closes #470.

Extracts the AI-enrichment flow out of `backend/api/findings.py` into
`services/findings/enrichment/` so it's reachable from more than one HTTP path.
Behaviour-preserving refactor — no functional change to
`POST /api/findings/{id}/enrich`.

## Why

`get_or_generate_enrichment` was a **333-line handler** that inlined the entire
flow: provider resolution, input shaping, prompt construction, dispatch,
response parsing, metadata stamping and persistence. "Enrich a finding with AI
analysis" could therefore only happen from that one HTTP path — ingestion, the
daemon and agents had no way to reuse it. None of the pure logic (prompt
building, JSON extraction) was unit-testable without a live provider and an
HTTP round-trip.

## What changed

New module, one seam:

```python
from services.findings.enrichment import enrich, EnrichmentError

payload = await enrich(finding)                 # persists by default
payload = await enrich(finding, persist=False)  # compose your own write
```

| File | Contents |
|---|---|
| `services/findings/enrichment/errors.py` | Domain exceptions + their status-code mapping |
| `services/findings/enrichment/prompt.py` | Input shaping + prompt build — pure functions |
| `services/findings/enrichment/parse.py` | JSON extraction + text fallback — pure functions |
| `services/findings/enrichment/service.py` | `enrich()` — resolve → dispatch → parse → stamp → persist |
| `services/findings/enrichment/__init__.py` | Public surface |

`backend/api/findings.py` (`+31 / −281`) keeps only what is genuinely HTTP
policy — fetch, cache check, and the error translation:

```python
finding = data_service.get_finding(finding_id)   # 404 stays here
...                                              # cache check stays here
enrichment = await enrich(finding, data_service=data_service)
```

The handler function drops from **333 lines to 75** (45 of them body; the rest
is its pre-existing docstring).

### No `HTTPException` in `services/`

The old code raised `HTTPException` from inside the business logic, which is
precisely what made it unusable from the daemon and from ingestion — neither
speaks HTTP. The module now raises domain errors and the handler owns the
single translation point:

| Exception | Status | Detail |
|---|---|---|
| `FindingNotFound` | 404 | `"Finding not found"` |
| `NoProviderConfigured` | 503 | `claude.NO_PROVIDER_DETAIL` (structured) |
| `ProviderUnavailable` | 503 | `str(exc)` |
| `EmptyProviderResponse` | 500 | `"Failed to generate enrichment: {exc}"` |

`NO_PROVIDER_DETAIL` deliberately stays in `backend/api/claude.py` and is
re-applied at the HTTP boundary, rather than following the module into
`services/` — three backend modules import it, and a `backend → backend` import
must not leak into the service layer. The 503 keeps carrying the structured
payload (not a bare string) so the chat drawer still renders its "Configure a
provider" CTA.

## Design choices

- **`enrich()` takes a finding `dict`, not a `finding_id`.** The fetch — and its
  404 — stays with the caller, so the module needs no database access for reads.
  The daemon already holds finding dicts.
- **No cache check inside `enrich()`.** Whether an existing `ai_enrichment`
  value is a usable cache hit is caller policy (`force_regenerate` is a query
  param), and the current check has a known bug (see below).
- **`persist=False` returns the payload without writing.** So a caller can
  compose its own write instead of the module hard-coding a destructive
  full-replace deep inside shared code.
- **`component` is a parameter** (default `"reporting"`, as the handler
  hardcoded) — other callers may want a different `ai_model_configs` component.
- **`data_service` is injectable.** Without it the module would construct a
  second `DatabaseDataService` (whose `__init__` runs `init_database`) alongside
  the API module's singleton. The handler passes its own; there's a test for it.

## Preserved as-is

**The dispatch asymmetry.** Anthropic goes through the sync `ClaudeService` on a
threadpool with `max_tokens=4096` and **no retry**; every other provider goes
through `LLMRouter.dispatch` with `max_tokens=1400`, a retry loop, and
ollama-only local-Bifrost recovery. That is existing behaviour, not a design
position — unifying it while moving code would make any regression impossible to
bisect. Seven tests now pin it so it can't drift silently. Whether the cloud path
*should* retry is a separate question.

The `run_in_executor` hop is left exactly where it is — it's the sync-in-async
seam #461 covers. `enrich()` is `async`; nothing here entrenches or resolves it.

## Known issue documented, not fixed: two writers on `ai_enrichment`

`daemon/processor.py:441` writes triage keys (`ai_triage`, `enrichment`,
`enriched_at`, `triage_confidence`, `category`, `recommended_action`,
`triage_reasoning`) into the same JSONB column this module writes analysis keys
to (`threat_summary`, `potential_impact`, `recommended_actions`, …). **The two
key sets do not overlap.** Three consequences, all live on `main` today:

1. **Cache poisoning across writers.** If the daemon triaged first,
   `POST /{id}/enrich` returns the daemon's payload labelled as `enrichment`;
   the frontend reads `undefined` off `enrichment.threat_summary`.
2. **API writes destroy daemon triage.** `update_finding(..., ai_enrichment=…)`
   is a whole-value replace, not a merge.
3. **API-enriched findings fall out of backfill permanently.** The daemon
   re-queues on `ai_enrichment IS NULL`; a non-NULL write excludes the finding
   forever.

Fixing this is a JSONB schema change needing a read-side compatibility shim for
existing rows (there's no Alembic yet — #411), so bundling it here would make
the diff unreviewable. It's spelled out in `service.py`'s module docstring, and
`persist=False` is the seam it needs. **Follow-up issue to be filed.**

## Testing

**55 new tests**, all passing:

- `tests/unit/test_findings_enrichment.py` (41) — entity singular/plural/alias
  keys, the `predicted_techniques` vs `mitre_predictions` fallback and its sort
  order, fenced/raw/malformed/empty response parsing, metadata stamping,
  `persist` on/off, and the dispatch asymmetry.
- `tests/unit/test_findings_enrich_endpoint.py` (14) — 404, cache hit,
  `force_regenerate`, both 503 shapes, 500, the success envelope, and
  `_resolve_provider`'s error mapping.

**Byte-identity proof.** The pre-refactor inline code was extracted verbatim
from `git show b0f8a94:backend/api/findings.py` and its output diffed against
the new module across **7 finding shapes** (plural keys, singular keys,
`dest_ips`/`usernames` aliases, all-absent, all-`None`-valued, empty dicts,
plural-empty-falls-back-to-singular) and **6 response shapes × 3 severities** —
prompts and parsed payloads identical, **0 mismatches**.

Full suite: `pytest tests/unit/ -m "not external_service"` → **1084 passed, 51
skipped, 0 failed**. `black` clean; `mypy` clean on the new package.

## Scope / sequencing

**not** in this PR:

- **The `ai_enrichment` collision fix** — see above; separate PR, separate review.
- **Daemon / ingestion adoption.** #470's acceptance criteria mention them, but
  no parallel AI-enrichment implementation exists in either
  (`grep -rn "ai_enrichment"` hits only `backend/api/findings.py`,
  `daemon/processor.py` and `services/database_data_service.py:210`)

⚠️ **Conflicts with #451.** That PR also modifies `backend/api/findings.py`
(`+16/−14`) and makes `LLMRouter` the sole LLM entry point, removing direct
`ClaudeService` imports. Its changes to the enrichment flow now land in
`service.py::_resolve_provider` and `_dispatch` instead of in the handler —
two small functions rather than a 333-line body. If #451 lands first, both
collapse further, likely to a single `dispatch()` call.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,12 +6,16 @@ This file provides guidance for AI assistants (Claude Code and similar tools) wo
 
 ## Project Overview
 
-**Vigil** is an open-source, AI-native Security Operations Center (SOC) platform. It orchestrates 13 specialized AI agents via Claude to perform triage, investigation, threat hunting, forensics, and automated response across 30+ security integrations.
+**Vigil** is an open-source, AI-native Security Operations Center (SOC) platform. It orchestrates 13 specialized AI agents via Claude to perform triage, investigation, threat hunting, forensics, and automated response across 40 security integrations.
 
 **Core pillars:**
-- **Agents** — 13 specialized AI agents (Triage, Investigator, Threat Hunter, Correlator, Responder, Reporter, MITRE Analyst, Forensics, Threat Intel, Compliance, Malware Analyst, Network Analyst)
+- **Agents** — 13 agents defined in `AGENT_CONFIGS`, which is the authoritative
+  list: triage, investigator, threat_hunter, correlator, responder, reporter,
+  mitre_analyst, forensics, threat_intel, compliance, malware_analyst,
+  network_analyst, auto_responder. (The README says "12" — it omits
+  `auto_responder`.)
 - **Workflows** — Multi-agent orchestrated playbooks (Incident Response, Full Investigation, Threat Hunt, Forensic Analysis)
-- **Integrations** — 30+ tools via MCP protocol (Splunk, CrowdStrike, VirusTotal, Shodan, Timesketch, Jira, Slack, etc.)
+- **Integrations** — 40 MCP servers in `mcp-config.json` (Splunk, CrowdStrike, VirusTotal, Shodan, Timesketch, Jira, Slack, etc.). Count only dict-valued keys: the `mcpServers` object also holds 7 `_comment_*` string keys used as section separators.
 
 **Ports:**
 - Backend API: `http://localhost:6987`
@@ -64,9 +68,16 @@ vigil/
 ├── docs/                 # Detailed documentation
 ├── docker/               # docker-compose.yml + Dockerfiles
 ├── scripts/              # Init and utility shell scripts
-├── mcp-config.json       # 30+ MCP server definitions
+├── mcp-config.json       # 40 MCP server definitions (+ `_comment_*` separator keys)
 └── env.example           # Template for all 220+ environment variables
 ```
+
+> **sys.path quirk:** `backend/main.py` puts `backend/` on `sys.path`, so modules
+> there are imported *bare* — `from secrets_manager import get_secret`, not
+> `from backend.secrets_manager import …`. `setup.cfg` (`mypy_path = backend`) and
+> `pyrightconfig.json` (`extraPaths`) mirror this so static analysis resolves them.
+> Note the resulting name collisions: `backend/services/` and `backend/tools/` are
+> **not** the top-level `services/` and `tools/` packages.
 
 ---
 
@@ -96,6 +107,18 @@ cd frontend && npm run dev
 # 4. (Optional) Daemon
 ./start.sh --daemon
 ```
+
+### Desktop (Electron)
+
+```bash
+cd desktop && npm run dev    # builds TS, launches Electron
+cd desktop && npm run dist   # packages a .dmg via electron-builder
+```
+
+The desktop app drives the stack through `scripts/app_up.sh` / `app_down.sh`
+rather than `start.sh`: no `--reload`, no Vite (the built SPA is served by the
+backend at :6987), and it **forces `DEV_MODE=false`** because its login and
+first-run bootstrap are the point.
 
 ### Fresh Environment
 
@@ -138,23 +161,29 @@ Default dev login: **admin / admin123** (when `DEV_MODE=false`)
 
 ### Python (pytest)
 
-```bash
-# All tests
-pytest
+The pytest config lives at **`tests/pytest.ini`** — there is no root-level
+`pytest.ini` and no `[tool:pytest]` in `setup.cfg`. A bare `pytest` from the repo
+root therefore finds no ini file: markers go unregistered, `--asyncio-mode=auto`
+and the coverage flags are skipped, and collection starts from the CWD instead of
+`testpaths = tests`. Always scope to `tests/` (or pass `-c tests/pytest.ini`).
 
-# With coverage
-pytest --cov=. --cov-report=html
+```bash
+# All tests, with the intended config
+pytest tests/
 
 # By marker
-pytest -m unit
-pytest -m integration   # requires running PostgreSQL
-pytest -m "not slow"
+pytest tests/ -m unit
+pytest tests/ -m integration   # requires running PostgreSQL
+pytest tests/ -m "not external_service"   # what CI's main unit job runs
 
 # Specific file
-pytest tests/test_backend_tools.py -v
+pytest tests/unit/test_backend_tools.py -v
 ```
 
-Available markers: `unit`, `integration`, `slow`, `auth`, `siem`, `claude`, `database`, `api`, `daemon`, `performance`
+Available markers: `unit`, `integration`, `slow`, `auth`, `siem`, `claude`,
+`database`, `api`, `daemon`, `performance`, `external_service`.
+`--strict-markers` is on, so an unregistered marker is an error — add new ones to
+`tests/pytest.ini`.
 
 ### Frontend (vitest)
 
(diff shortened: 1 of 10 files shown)


mattmorr1:
@thetosy please fix conflicts and push to refactor branch 🙏🏼

CLAUDE.md
```diff
@@ -6,16 +6,22 @@ This file provides guidance for AI assistants (Claude Code and similar tools) wo
 
 ## Project Overview
```
nestor-deeptempo, line 7: when fixing the merge conflicts on this file, there's some worth keeping/some worth dropping

i.e.

Fix/drop — stale or regressive vs current main:

The "Two UIs coexist / legacy MUI at /old/*, MUI v5 + theme/" section. [#502](https://github.com/Vigil-SOC/vigil/issues/502) deleted the legacy MUI UI and already rewrote these lines to "redesign is the console… there is no component library." Re-adding it documents frontend/src/pages/ and theme/ that no longer exist — take main's current version here.
The "⚠️ frontend tests don't run in CI" caveat — no longer true on main; ci-cd.yml runs npm run test:run.

Rephrase — about to collide with the reorg:

The agents section pins services/soc_agents.py as authoritative right as #530 moves it to core/agents/. Keep the insight (auto_responder is the 13th agent, README says 12, AGENT_CONFIGS is the source of truth) but phrase it path-agnostically.
Drop the churny size/count bumps (~124KB→~164KB, "70+"→"83"); keep the genuinely new modules (llm_clients / llm_router / model_registry).

Correct — worth keeping (don't lose these while resolving the conflict):

API-route example AsyncSession → sync Session — main's current example tells contributors to await a sync session, which is broken (there is no AsyncSession in the repo), so this is a real fix.
services/models.py → database/models.py — main currently cites a file that no longer exists.
MCP count 30+ → 40 with the _comment_* caveat, the tests/pytest.ini location note, the helm-chart.yml + Desktop sections, the sys.path quirk, and the Bifrost section.

services/findings/enrichment/service.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    if not finding:
+        raise FindingNotFound("Finding not found")
+
+    # Resolve before shaping input, so an unconfigured provider still reports
+    # NoProviderConfigured rather than whatever a malformed finding raises.
+    provider, model_id, claude_service = _resolve_provider(component)
+
+    summary = summarize_finding(finding)
+    finding_id = summary.finding_id
```
nestor-deeptempo: The old handler persisted with the path param; enrich() now takes the write target from the finding dict here, flowing into _persist (L278 → L199). When finding_id is missing/falsy, summarize_finding falls back to "", so the enrichment is written to update_finding("", ...) — a silent lost write. Masked over HTTP (get_finding(id) returns finding_id == id), but the daemon/agents seam this is built for passes its own dicts, which is where it'd bite. Only the happy path is tested. Fix: pass the caller's authoritative id into enrich() (or raise on empty)

nestor-deeptempo (changes requested):
i think it's ok to keep this to main as we haven't touched these files yet in the refactor, i haven't at least

services/findings/enrichment/prompt.py
mattmorr1, line 1: in general, try to keep docstrings to a minimum. also specific references to issues can be confusing

services/findings/enrichment/service.py
mattmorr1, line 1: same here

services/findings/enrichment/errors.py
mattmorr1, line 1: these classes don't do anything

services/findings/enrichment/prompt.py
mattmorr1, line 1: i dont think you can hard code temperature in here without passing directly into the api itself.

nestor-deeptempo: not sure what temperature is exactly

services/findings/enrichment/errors.py
nestor-deeptempo: probably don't need all of them but they're error wrappers so errors raise something more structured instead of a string or something

nestor-deeptempo (approved):
LGTM

Issue #549, opened 2026-08-03
case_evidence.chain_of_custody: move the custody log out of JSONB into a table


craig-dt:
**Coding:** Med · **Vigil context:** Med · **Lane:** backend / database · **Type:** correctness

**This is the highest-priority promotion in #468, and it is a correctness argument rather than a
tidiness one.** `case_evidence.chain_of_custody` (`Mapped[List[dict]]`, `nullable=False`) holds an
append-only legal record as a JSON array. That means:

- it **cannot be constrained** append-only — any writer can rewrite or truncate history;
- it **cannot be indexed** by actor or time, so "what did this custodian touch?" is a scan;
- every append is a read-modify-write of the whole array, so **two concurrent custody events can
  silently drop one**.

That last failure mode is not hypothetical for JSONB arrays in this codebase — #543 documents live
instances of in-place appends being discarded entirely.

## Acceptance criteria

- `case_evidence_custody(entry_id, evidence_id FK, occurred_at, actor, action, notes, created_at)`,
  or similar, with an index on `(evidence_id, occurred_at)`.
- Backfill preserving array order; entries without a timestamp get a deterministic, documented
  ordering rather than being dropped.
- Inserts replace array mutation at every writer.
- No `UPDATE` or `DELETE` path on custody rows — append-only is the point. Enforce it at least by
  convention plus a test, ideally by a trigger or a revoked grant.
- The API keeps returning the same custody list shape.

## Migration mechanism, and the compose caveat

Deliverable today via a numbered file in `database/init/` — this does **not** need #411 — but the
two deploy paths differ and the PR must state which it verified:

- **Helm:** `helm/vigil/templates/db-init-job.yaml:68-99` maintains
  `_vigil_schema_versions(filename PRIMARY KEY)` and skips already-applied files, so a new file
  applies on `helm upgrade`.
- **docker-compose:** `database/init/` is mounted at `/docker-entrypoint-initdb.d`, which Postgres
  runs **only when initialising an empty data directory**. An existing local database will not pick
  the file up — a developer must recreate the volume or apply the SQL by hand. Say so in the PR.

Per `CLAUDE.md`, a new init file also needs copying into `helm/vigil/files/database-init/` and adding
to `helm/vigil/values.yaml` `dbInit.sqlFiles` **in execution order**, verified with
`helm template ... | grep -E '^[[:space:]]*apply "NEWFILE\.sql"'`. CI's `diff -r` catches a missing
[...]

Issue #545, opened 2026-08-03
Collapse the case-event API surface to a single stream; drop the legacy JSONB columns


craig-dt:
> **Sequence: 3 of 3 — do not start before the other two land.**
> **Blocked by #544** (unify into `case_events`), which is itself blocked by #543 (silent append loss).
> #543 → #544 → this. This is the only one of the three that changes API output, which is why it is last
> and separate: if it needs reverting, nothing structural comes back with it.

**Coding:** Low–Med · **Vigil context:** Med · **Lane:** backend / frontend

**Context.** #544 unifies `cases.timeline`, `cases.activities` and `cases.notes` into a `case_events` table while **deliberately preserving** the old API shape: `Case.to_dict()` keeps deriving the three legacy keys so no reader has to change. That is the right way to land the storage change safely, but it leaves the API describing a split that no longer exists in the database.

This issue removes that scaffolding.

**Why it is separate rather than folded into #544.** #544 is provable by construction — a before/after comparison shows byte-equivalent output, so it cannot break a consumer. This one *intends* to change output, so it needs its own review, its own frontend work, and its own revert. Bundling them would mean a storage migration could only be rolled back together with a UI change.

## Scope

**Backend**

- `Case.to_dict()` (`database/models.py:262-263`) stops emitting derived `timeline` / `activities` / `notes` keys and emits a single ordered `events` list instead.
- `backend/api/timeline.py:95-119` stops fanning out over three collections; it reads `case_events` directly and maps `event_type` onto the `TimelineEvent.type` it already uses (`"activity"` / `"note"` / `"status"`).
- `backend/api/cases.py:333-345` (add-activity) and `:1112-1113`, `:1137-1139` (case merge) drop their read-modify-write handling.
- The three legacy columns are dropped from `cases` — the point at which #544's decision to leave them in place is finally settled.

**Consumers that must move in step**

- `frontend/src/redesign/screens/cases/useCases.ts:109,114` — currently `setActivities(data.activities || [])` and a separate `data.timeline || []`. Becomes one stream.
- `frontend/src/redesign/screens/cases/CaseSections.tsx:821` — `ActivityCard({ activities })` takes the unified type.
- `services/report_service.py:308,345` — currently renders two separate PDF tables ("Timestamp/Event" and "Timestamp/Type/Description"). Decide whether the report keeps two sections or becomes one chronological table; a merged table is arguably a better report, but it is a **user-visible output change** and should be an explicit choice, not a side effect.
[...]

Issue #544, opened 2026-08-03
Unify cases.timeline / activities / notes into a case_events table (no API change)


craig-dt:
> **Sequence: 2 of 3.**
> **Blocked by #543** (in-place JSONB appends are silently discarded) — land that first, or this
> migration will backfill from columns that are already missing events.
> Followed by #545 (collapse the API surface), which assumes this landed.
> This PR deliberately makes **no** API-visible change.

**Coding:** Med · **Vigil context:** High · **Lane:** backend / database

**Context.** `cases` carries three JSONB arrays that are the same concept stored three ways: `timeline`, `activities` and `notes`. The code already treats them as one — `backend/api/timeline.py:95-119` flattens all three into a single `TimelineEvent` stream, and gives `timeline` and `activities` **the same** `type="activity"`:

```python
for timeline_event in case.get('timeline', []):
    events.append(TimelineEvent(..., content=timeline_event.get('event'), type="activity", ...))
for activity in case.get('activities', []):
    events.append(TimelineEvent(..., content=activity.get('description'), type="activity", ...))
for note in case.get('notes', []):
    events.append(TimelineEvent(..., content=f"Note: ...", type="note", ...))
```

So the split is an artifact, not a distinction. And the shapes confirm which one is the better model — `activities` is a strict superset:

| | `timeline` | `activities` |
|---|---|---|
| fields | `timestamp`, `event` | `timestamp`, `activity_type`, `description`, `details` |
| written by | `services/case_workflow_service.py` (lifecycle events) | `backend/api/cases.py` (API + case merge) |

`event` ≈ `description`; `activities` adds a discriminator and a structured payload. Unifying is therefore **adopting the `activities` shape as canonical**, not reconciling two peers.

**Why a table rather than one JSONB column.** These are records with their own identity, ordering and authorship. As JSON arrays they cannot be indexed, filtered or paginated; there is no FK to attribute an entry to a user; and every append is a read-modify-write of the whole array, so two concurrent events can silently drop one. #543 is one instance of that exposure; a child table makes the whole class impossible, because an `INSERT` cannot be silently dropped.

**`notes` is in scope on purpose.** It is the third member of the same stream (`type="note"`). Unifying two of three would leave the same problem in place with a smaller surface, and would mean designing the discriminator twice.

## Proposed shape

```sql
CREATE TABLE IF NOT EXISTS case_events (
    event_id     TEXT PRIMARY KEY,
[...]

PR #556, opened 2026-08-03
fix(db): track in-place JSONB mutations so case timeline events persist

### LINKED ISSUE #468: Audit the 74 JSONB columns (keep vs. promote to typed)
**Coding:** Low · **Vigil context:** High · **Lane:** backend · **Good for:** someone who knows the data model; more analysis than code

**Context.** The schema has 74 JSONB columns. Some are legitimately schemaless (arbitrary vendor payloads, evidence blobs); others are typed data stashed in JSON to dodge a migration, which hides structure from the ORM, the API contract, and any query. This is a classification pass, not a big code change.

**Acceptance criteria**
- Each of the 74 columns classified keep-as-JSONB vs. promote-to-typed, with a one-line rationale.
- Follow-up migration issues filed for the "promote" set.

**Files:** `database/models.py` + the migration path.

**Note:** best done after Alembic (#411) lands, so promotions have a real migration mechanism.

### LINKED ISSUE #543: Case timeline events silently discarded: in-place JSONB appends are never persisted
> **Sequence: 1 of 3.** Ships first and stands alone — this is a data-loss bug fix, not a refactor.
> Not blocked by anything. Followed by the unification (issue B) and then the API collapse (issue C),
> both of which assume this landed. Do not merge B before this.

**Coding:** Low · **Vigil context:** Med · **Lane:** backend · **Type:** bug (silent data loss)

**Context.** Case timeline events are being silently discarded. `case.timeline.append(...)` mutates a `JSONB` column in place, no JSONB column in `database/models.py` is wrapped in `MutableList`/`MutableDict`, and `flag_modified` appears **nowhere** in the codebase — so SQLAlchemy never marks the attribute dirty and emits no `UPDATE`. The commit succeeds, nothing errors, the event is gone.

Three append sites in `services/case_workflow_service.py`:

| Site | Event recorded | Persists? |
|---|---|---|
| L310 | playbook step applied | ✅ — but only incidentally, because it calls `session.merge(case)` |
| L403 | **case auto-assigned to an analyst** | ❌ **silently discarded** |
| L470 | **case escalated** (`Escalated to {x}: {reason}`) | ❌ **silently discarded** |

So auto-assignment and escalation never appear in a case's timeline. Both are exactly the kind of event an analyst or auditor would go looking for after the fact.

**Why the sibling column is unaffected.** `cases.activities` writes are safe because every one goes through explicit assignment rather than in-place mutation — `data_service.update_case(case_id, activities=activities)` at `backend/api/cases.py:345`, and the merge path rebuilds the list with `+` at `backend/api/cases.py:1112-1113` (which is also why the subsequent `target.activities.append(merge_activity)` at L1138 *does* persist — the attribute is already dirty with a new list object). The bug is specific to the in-place `.append()` pattern.

## Fix

Wrap the mutable JSONB collection columns so SQLAlchemy tracks in-place changes:

```python
from sqlalchemy.ext.mutable import MutableList

timeline: Mapped[List[dict]] = mapped_column(
    MutableList.as_mutable(JSONB), nullable=False, default=list
)
```

This is preferable to sprinkling `flag_modified(case, "timeline")` at each call site, because the next `.append()` anyone writes is then correct by default rather than correct-if-remembered. Apply it to the JSONB collection columns that are appended to in place — at minimum `cases.timeline`; audit the other list-valued JSONB columns for the same pattern while in there.

**Also fix the mutable-default footgun in the same PR.** These columns use `default=[]`, a single shared list constant, where the correct form is `default=list` (a factory). `users.mfa_recovery_codes` already gets this right; `cases.timeline`, `cases.notes`, `cases.activities` and `cases.resolution_steps` do not. Same file, same class of defect, and fixing it separately would mean touching these lines twice.

## Acceptance criteria

- A test that appends to `cases.timeline` via the ORM, commits, re-reads in a **new session**, and asserts the event is present. This must fail before the fix.
- Regression tests covering the two broken paths specifically: auto-assignment (`case_workflow_service.py:403`) and escalation (`:470`) both leave a timeline entry.
- `default=[]` replaced with `default=list` on the affected columns.
- No schema change, so no migration and no init-SQL/Helm work — this is ORM-layer only.
- No API contract change: `Case.to_dict()` keeps emitting `timeline` and `activities` unchanged.

## Notes

- Found while auditing the JSONB columns for #468; the audit doc records it as the reason `cases.timeline` and `case_evidence.chain_of_custody` are promote-not-keep. Any array-valued JSONB column appended in place has the same exposure, which is a general argument for child tables — but that is issue B's job, not this one.
- This fix is worth landing **even if B and C are never approved**.

---

craig-dt:
## The bug

Case timeline events are being **silently discarded**. `case.timeline.append(...)` mutates a `JSONB` column in place; no JSONB column was wrapped in `MutableList`, and `flag_modified` appears **nowhere** in the codebase — so SQLAlchemy never marks the attribute dirty and emits no `UPDATE`. The commit succeeds, nothing errors, the event is gone.

Reproduced against real Postgres before fixing — three appends, one commit, re-read in a fresh session:

```
E  AssertionError: expected all three in order, got []
```

Every event lost. Two of the three append sites in `services/case_workflow_service.py` were affected:

| Site | Event | Before |
|---|---|---|
| L310 | playbook step applied | ✅ worked — but only incidentally, because it happens to call `session.merge(case)` |
| L403 | **case auto-assigned to an analyst** | ❌ **silently discarded** |
| L470 | **case escalated** (`Escalated to {x}: {reason}`) | ❌ **silently discarded** |

So auto-assignment and escalation never appeared in a case timeline. Both are exactly what an analyst or auditor would go looking for after the fact.

**`cases.activities` was unaffected**, and it's worth saying why rather than leaving it ambiguous: every write there goes through explicit assignment rather than in-place mutation — `update_case(case_id, activities=...)` at `backend/api/cases.py:345`, and the merge path rebuilds the list with `+` at `:1112-1113`, which is also why the later `.append()` at `:1138` *does* persist. The bug is specific to the in-place pattern.

## The fix

Declare the appendable collection columns with `MutableList.as_mutable(JSONB)`, via a named `MutableJSONBList` alias carrying the explanation. Preferred over sprinkling `flag_modified()` at each call site, because the next `.append()` anyone writes is then correct **by default** rather than correct-if-remembered.

Applied to the four `cases` collections plus `case_evidence.chain_of_custody` — a vanished append there would be a hole in an evidence audit trail. (#549 moves that column to its own table, removing the exposure entirely; this is the interim guard.)

**Also fixes every shared-mutable default in the file:** 8 × `default=[]` and 4 × `default={}` become `default=list` / `default=dict`. SQLAlchemy treats a non-callable default as a scalar constant, so all rows shared one object. The file already used the factory form in 23 places — this removes the second convention, and a test pins it so the two can't drift apart again.

## Tests — both verified to fail before the fix

**`tests/unit/test_jsonb_mutation_tracking.py`** — 11 tests, **no database**, so they run in the main PR gate. Asserts each column coerces to `MutableList` and that an in-place append fires a change event, plus the no-shared-defaults guard.

```
pre-fix:   11 failed
post-fix:  11 passed
```

**`tests/integration/test_case_timeline_persistence.py`** — the end-to-end symptom. Appends, commits, re-reads in a **fresh session** with its own identity map, so a stale cached object can't make it pass spuriously. Also covers repeated appends in one transaction, which catches tracking that fires only on the first mutation.

```
pre-fix:   expected all three in order, got []
post-fix:  2 passed
```

## Regressions

| Suite | Result |
|---|---|
| unit | **1070 passed**, 13 failed — the same 13 pre-existing failures |
| security | **84 passed** |
| integration | 6 failed — **identical to the 6 on clean `main`** |

The integration baseline was taken in a throwaway worktree at `311790e` rather than assumed: **zero new failures**.

## Scope

No schema change, so no migration, no init-SQL and no Helm work. No API change — `Case.to_dict()` output is unchanged. This is ORM-layer only, and it's worth landing regardless of whether #544 and #545 are ever approved.

Found while auditing the JSONB columns for #468 (#546); it's the reason `cases.timeline` and `case_evidence.chain_of_custody` are classified promote rather than keep.

Closes #543
Relates to #468, #544, #549

### DIFF
--- a/database/models.py
+++ b/database/models.py
@@ -24,9 +24,19 @@ from sqlalchemy import (
     text,
 )
 from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, relationship
+from sqlalchemy.ext.mutable import MutableList
 from sqlalchemy.dialects.postgresql import JSONB, UUID
 from pgvector.sqlalchemy import Vector
 import uuid
+
+# Collection columns that are appended to in place (``obj.col.append(...)``) must
+# be declared with ``MutableList.as_mutable(JSONB)``. A plain JSONB column gives
+# SQLAlchemy no way to observe an in-place mutation, so the attribute never
+# becomes dirty, no UPDATE is emitted, and the append is silently discarded on
+# commit — see issue #543, where case auto-assignment and escalation events were
+# being lost exactly this way. Wrapping makes the next ``.append()`` anyone
+# writes correct by default rather than correct-if-they-remembered.
+MutableJSONBList = MutableList.as_mutable(JSONB)
 
 # Fixed width for the findings vector column; sources of other dimensions
 # (LogLM 512) are zero-padded/truncated to this before storage.
@@ -199,22 +209,26 @@ class Case(Base):
     assignee: Mapped[Optional[str]] = mapped_column(String(100), nullable=True)
 
     # Tags (array of strings)
-    tags: Mapped[List[str]] = mapped_column(ARRAY(String), nullable=True, default=[])
+    tags: Mapped[List[str]] = mapped_column(ARRAY(String), nullable=True, default=list)
 
     # Notes (JSONB array)
-    notes: Mapped[List[dict]] = mapped_column(JSONB, nullable=True, default=[])
+    notes: Mapped[List[dict]] = mapped_column(
+        MutableJSONBList, nullable=True, default=list
+    )
 
     # Timeline events (JSONB array)
-    timeline: Mapped[List[dict]] = mapped_column(JSONB, nullable=False, default=[])
+    timeline: Mapped[List[dict]] = mapped_column(
+        MutableJSONBList, nullable=False, default=list
+    )
 
     # Activities (JSONB array)
     activities: Mapped[Optional[List[dict]]] = mapped_column(
-        JSONB, nullable=True, default=[]
+        MutableJSONBList, nullable=True, default=list
     )
 
     # Resolution steps (JSONB array)
     resolution_steps: Mapped[Optional[List[dict]]] = mapped_column(
-        JSONB, nullable=True, default=[]
+        MutableJSONBList, nullable=True, default=list
     )
 
     # MITRE ATT&CK techniques
@@ -535,7 +549,7 @@ class UserPreference(Base):
     user_id: Mapped[str] = mapped_column(String(100), primary_key=True)
 
     # Preferences as flexible JSONB
-    preferences: Mapped[dict] = mapped_column(JSONB, nullable=False, default={})
+    preferences: Mapped[dict] = mapped_column(JSONB, nullable=False, default=dict)
 
     # User metadata
     display_name: Mapped[Optional[str]] = mapped_column(String(200), nullable=True)
@@ -586,7 +600,7 @@ class IntegrationConfig(Base):
     enabled: Mapped[bool] = mapped_column(Boolean, nullable=False, default=False)
 
     # Configuration (non-sensitive only)
-    config: Mapped[dict] = mapped_column(JSONB, nullable=False, default={})
+    config: Mapped[dict] = mapped_column(JSONB, nullable=False, default=dict)
 
     # Metadata
     integration_name: Mapped[Optional[str]] = mapped_column(String(200), nullable=True)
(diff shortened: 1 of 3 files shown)


nestor-deeptempo:
closing this as jack's pr last week seemed to cover it, one minor bug tracked here https://github.com/Vigil-SOC/vigil/issues/718

PR #558, opened 2026-08-03
fix(api): declare the watcher notification vocabulary

### LINKED ISSUE #553: case_watchers.notification_preferences: promote the flag set to real columns
**Coding:** Low · **Vigil context:** Low · **Lane:** backend / database · **Good for:** a first migration

`case_watchers.notification_preferences` (`Mapped[Optional[dict]]`) holds a handful of notification
booleans in JSONB. Typed columns document which switches exist; JSONB hides them, so the available
options can only be learned by reading writer code or sampling rows.

The smallest and lowest-risk promotion in #468 — a good candidate for whoever wants to establish the
migration pattern before the larger ones.

## Acceptance criteria

- Enumerate the keys actually written before designing columns; do not infer the set from one writer.
- One boolean column per preference, with explicit defaults matching today's effective behaviour when
  the key is absent.
- `to_dict()` keeps emitting a `notification_preferences` object so consumers are unaffected.
- A test covering a watcher row with the column absent/NULL, since today's `Optional[dict]` allows it.

## Migration mechanism, and the compose caveat

Deliverable today via a numbered file in `database/init/` — this does **not** need #411 — but the
two deploy paths differ and the PR must state which it verified:

- **Helm:** `helm/vigil/templates/db-init-job.yaml:68-99` maintains
  `_vigil_schema_versions(filename PRIMARY KEY)` and skips already-applied files, so a new file
  applies on `helm upgrade`.
- **docker-compose:** `database/init/` is mounted at `/docker-entrypoint-initdb.d`, which Postgres
  runs **only when initialising an empty data directory**. An existing local database will not pick
  the file up — a developer must recreate the volume or apply the SQL by hand. Say so in the PR.

Per `CLAUDE.md`, a new init file also needs copying into `helm/vigil/files/database-init/` and adding
to `helm/vigil/values.yaml` `dbInit.sqlFiles` **in execution order**, verified with
`helm template ... | grep -E '^[[:space:]]*apply "NEWFILE\.sql"'`. CI's `diff -r` catches a missing
copy; a missing `dbInit.sqlFiles` entry is **silent**.

**Rationale and full classification:** #468 / #546 (`docs/JSONB_AUDIT.md`).

---

craig-dt:
## What changed

`case_watchers.notification_preferences` is read at `services/case_notification_service.py:297-298`:

```python
prefs = watcher.notification_preferences or {}
if prefs.get(notification_type, True):  # Default to True
```

`notification_type` is a **runtime argument**, not a literal — so the keys that lookup honours are exactly the set of types passed to `notify_watchers`. Nothing in the repo declared that set: no constant, no enum, no Pydantic type, no TypeScript type, no SQL default, no seed row, no test. The available switches could only be learned by finding every call site by hand.

The API made this worse rather than hiding it. `WatcherAdd.notification_preferences` was a bare `Optional[Dict]`, so any JSON object was accepted and persisted verbatim. A caller sending `{"case_assigned": false}` got a **200** and reasonably concluded assignment notices were suppressed. They are not — `case_assigned`, `comment_mention` and `stale_case` call `create_notification` directly and never consult the column. The key was stored and never read.

This PR:

- Declares the vocabulary as `WATCHER_NOTIFICATION_TYPES` next to the lookup that consumes it (`new_comment`, `sla_warning`).
- Rejects unknown keys at the API boundary, so a caller learns immediately instead of trusting a switch that does nothing.
- Tightens the value type to `bool`, which the bare `Dict` never enforced.
- Warns rather than raises in `notify_watchers` on an unregistered type. The `new_comment` call site runs inside `add_comment`'s `try/except`, which rolls back and returns `None` — raising there would trade a silently ineffective preference for a silently dropped comment. It sends, matching the absent-key default.

**No behaviour change for any existing caller.** `to_dict()` still emits `notification_preferences`; a request omitting the field is still valid; the opt-out default (absent key → send) is unchanged.

## Why this is not the schema migration #553 asked for

#553 proposed promoting the column to typed boolean columns. That was declined after tracing every reader and writer — the issue's own first acceptance criterion. Three findings:

1. **The key set is not fixed.** It is keyed by `notification_type`, so typed columns would cap an open vocabulary and cost a schema migration for every new notification type.
2. **No row anywhere has data.** `add_watcher` is insert-only and returns early on an existing row, discarding incoming preferences; there is no PATCH/PUT route; the frontend posts `{user_id}` only (`frontend/src/services/api.ts:271-272`). The migration would have moved nothing.
3. **The prescribed mechanism does not apply.** `case_watchers` has no DDL anywhere — `grep -rn "case_watchers" database/ helm/` returns only `database/models.py:1066`. The table comes from `Base.metadata.create_all()`, so an `ALTER TABLE` in a numbered `database/init/` file would run in the dbInit Job *before the table exists* and hard-fail.

So: no schema change, no migration, no init SQL, no Helm work. The audit doc is corrected accordingly in #546, which also picks up a related finding — six of the promotion target tables have no init SQL at all, so several already-filed issues prescribe a mechanism that cannot reach them.

## How this was tested

`tests/unit/test_watcher_notification_vocabulary.py` — 13 tests, no database, so they run in the main PR gate.

Verified to fail before the change and pass after: **6 failed pre-change, 13 pass post-change.** The 7 that passed pre-change are deliberate regression guards on behaviour this PR must *not* alter (absent key sends, NULL column sends, `False` suppresses, per-type scoping).

Coverage:
- The vocabulary as a change-detector, so adding a `notify_watchers` call site without registering its type is visible in review.
- Types that bypass preferences (`case_assigned`, `comment_mention`, `stale_case`) are asserted *absent*, since listing them would advertise a switch that does nothing.
- The opt-out default: absent key **and** NULL column both still send.
- Unregistered type warns and still sends.
- Unknown key → 422 through FastAPI; the error names the accepted set.

The route test is itself evidence of the bug: **pre-change it reached real Postgres and failed on a foreign-key violation** — proof that an unknown key passed validation and got as far as the database. Post-change it never reaches the service layer.

**Regressions:** unit 1088 passed with the same **17** pre-existing failures as clean `main`; security 84 passed with the same **1**. Both baselined in a throwaway worktree at `311790e` and diffed by test name — **zero new failures**.

**Lint:** flake8 reports **0 findings on added lines**; `black --diff` wants to touch none of them; the new import sits where isort wants it. `black` was deliberately *not* run across `backend/api/cases.py` or `services/case_notification_service.py` — both are pre-existing-dirty and a bulk pass would rewrite unrelated code.

## Follow-ups found while tracing (not in this PR)

Filed separately rather than riding along:

- `CaseNotification(metadata=...)` is silently swallowed — the column is `notification_metadata`, and `hasattr(cls, "metadata")` is True (the declarative `MetaData`), so the kwarg sets a shadowing instance attribute instead of raising. **Every notification's metadata is dropped.** Confirmed empirically. Same failure family as #543.
- `add_watcher` discards preferences for an existing watcher, and no update route exists — so preferences are unsettable after creation.
- Frontend `Watcher.added_at` vs backend `created_at`; "Watching since" always renders from `undefined`.
- `scripts/migrate_schema.py` is invoked by no deploy path, so ORM-only column drift is silent until a human runs it.

Closes #553

---

*Analysis by Claude (Claude Code), reviewed by @craig-dt before posting. All file:line citations were verified against `main` at `311790e`.*


nestor-deeptempo:
Rebased onto current `main` and fixed up the vocabulary test. The diagnosis still holds — I re-verified it against `main` before touching anything.

## Still true

- The lookup is unchanged: `prefs.get(notification_type, True)` at `core/cases/case_notification_service.py:221`.
- Nothing declares the vocabulary. `WATCHER_NOTIFICATION_TYPES` did not exist anywhere.
- Exactly two `notify_watchers` call sites, passing `new_comment` and `sla_warning`. The proposed set is right.
- `comment_mention` and `stale_case` do bypass the column via direct `create_notification` calls.
- The warn-rather-than-raise reasoning holds: `add_comment` wraps everything in `try/except → return None`, so raising would trade an ineffective preference for a silently dropped comment.

## Ported (merge commit)

Every file this touched moved in #481:

- `backend/api/cases.py` → `services/api/routers/cases.py`
- `services/case_notification_service.py` → `core/cases/case_notification_service.py`

`notify_watchers` was also refactored to `unit_of_work()` since, so the warning now sits above that block rather than the old `should_close_session` dance. A caller-supplied session is yielded unchanged, so the test doubles still hold. The test's `sys.path` block pointed at `backend/`, which no longer exists, and its `DEV_MODE` line duplicates `tests/conftest.py` — both dropped.

## Fixed up (second commit)

**The change-detector didn't detect.** It compared the constant against a hardcoded copy of itself, so it only noticed someone editing the constant. Adding a `notify_watchers` call site with an unregistered type — the case the docstring named — left it green. It fired on the correct action and stayed silent on the incorrect one.

Replaced with an AST scan over `core/`, `services/` and `tools/` that collects the literal `notification_type` of every `notify_watchers` call and compares that to the constant. Verified it bites: appending a call site with `notification_type="escalation"` fails the assertion with `Extra items in the left set: 'escalation'`. Added a second test to guard the guard, since a scan matching nothing would pass vacuously.

**`case_assigned` no longer exists.** `notify_case_assignment` was deleted as dead code in #585, three days after this PR was opened. It was named in the constant's comment and asserted absent in a test — where it could no longer fail either way. Dropped from both; `comment_mention` and `stale_case` remain, and both still bypass.

## Verification

14 tests pass. Full unit suite `1583 passed, 32 skipped, 0 failed` — the 17 pre-existing failures noted in the original description are gone from `main`, so there is nothing to attribute. `lint-imports` reports 3 contracts kept. flake8 finds nothing on the added lines; `black --check` is clean on the test file. `isort` still reports `services/api/routers/cases.py` as unsorted, but it reports the same on `main` — pre-existing, left alone rather than bulk-reformatting unrelated imports.

## One note on scope

Worth stating plainly, since the description says no behaviour change: a caller sending `{"case_assigned": false}` got a 200 before and gets a 422 now. That is the intended improvement — the whole point is that the key was never honoured — but it is a break for that caller, and the vocabulary can shrink underneath callers, as `case_assigned` just demonstrated.

Also flagging that #559 (notification metadata silently dropped, found while tracing this) is still open and still reproduces on `main`:

```
notification_metadata column -> None
instance .metadata is       -> dict
```

Every notification's metadata is discarded today. Arguably ahead of this one in priority.

nestor-deeptempo (approved):

Issue #560, opened 2026-08-04
Watcher notification preferences are unsettable: add_watcher discards them and no update route exists


craig-dt:
**Coding:** Low · **Vigil context:** Low · **Lane:** backend / frontend · **Type:** bug (dead feature)

**Context.** `case_watchers.notification_preferences` can never be changed after a watcher row is created, and in practice can never be set at all — so the per-watcher notification opt-out is inert in every deployment.

Two independent causes.

**1. `add_watcher` discards preferences for an existing watcher.** `services/case_collaboration_service.py:294-308`:

```python
existing = session.query(CaseWatcher).filter(
    and_(CaseWatcher.case_id == case_id, CaseWatcher.user_id == user_id)
).first()

if existing:
    logger.info(f"User {user_id} already watching case {case_id}")
    return existing          # incoming notification_preferences silently ignored

watcher = CaseWatcher(
    case_id=case_id, user_id=user_id,
    notification_preferences=notification_preferences or {}
)
```

The early return drops the caller's preferences on the floor and returns HTTP 200 with the untouched row. A caller re-POSTing to change a preference gets a success response and no change.

**2. There is no update route.** `backend/api/cases.py` exposes exactly three watcher endpoints — `POST /{case_id}/watchers` (`:658`), `DELETE /{case_id}/watchers/{user_id}` (`:671`), `GET /{case_id}/watchers` (`:682`). No PATCH, no PUT. Preferences are settable only at creation.

**And nothing sets them at creation.** `frontend/src/services/api.ts:271-272`:

```ts
addWatcher: (id: string, userId: string) =>
  api.post(`/cases/${id}/watchers`, { user_id: userId }),
```

Only `user_id` is sent, so `add_watcher` stores `{}`. There is no seed data, no daemon path, and no test that writes a non-empty dict. Combined: the column is `{}` on every row in every deployment, and `notify_watchers`' `prefs.get(notification_type, True)` always takes the default.

**The UI presents this as working.** `frontend/src/redesign/screens/cases/CaseSections.tsx:605` renders a hardcoded badge for every watcher:

```tsx
<span className="watcher-active" title="Receiving notifications"><span className="wa-dot" />Active</span>
```

It is a static literal, not derived from the column. There is no per-type toggle anywhere in the UI. So the console asserts a notification state it never reads and the user cannot change.

## Fix

Decide first whether the feature is wanted, because "delete it" is a legitimate answer for a switch nobody can reach.

[...]

PR #564, opened 2026-08-04
fix(ui): read the watcher timestamp from created_at

### LINKED ISSUE #561: Watchers card always shows "Watching since —": frontend reads added_at, API sends created_at
**Coding:** Low · **Vigil context:** Low · **Lane:** frontend · **Type:** bug (cosmetic) · **Good for:** a first frontend fix

**Context.** The watchers card always shows `Watching since —` because the frontend reads a field the API does not send.

`frontend/src/redesign/screens/cases/CaseSections.tsx:560-563`:

```ts
interface Watcher {
  user_id: string
  added_at?: string
}
```

But `CaseWatcher.to_dict()` (`database/models.py:1098-1105`) emits:

```python
{"case_id": ..., "user_id": ..., "notification_preferences": ..., "created_at": ...}
```

The field is `created_at`. `added_at` is never present on a watcher, so `fmtD(w.added_at)` at `:604` is always called with `undefined`.

**It fails quietly rather than loudly.** `fmtD` (`:34-38`) guards its input:

```ts
function fmtD(s?: string): string {
  if (!s) return '—'
  ...
}
```

So the row renders `Watching since —` — not `Invalid Date`, and no console error. The optional `?` on `added_at` means TypeScript is satisfied too. Nothing surfaces the mismatch.

**Why the wrong name is plausible.** `added_at` *is* a real column in this schema — on the `case_findings` association table (`database/models.py:58`). It just isn't on `case_watchers`, which uses `created_at`.

The interface also omits `case_id` and `notification_preferences`, both of which the API does return.

## Fix

Rename the field to `created_at` in the `Watcher` interface and at the call site:

```ts
interface Watcher {
  user_id: string
  created_at?: string
}
```

```tsx
<div className="text-xs text-tx-faint">Watching since {fmtD(w.created_at)}</div>
```

Optionally add the two missing fields to the interface so it describes the actual response.

## Acceptance criteria

- A test rendering `WatchersCard` against a response shaped like `CaseWatcher.to_dict()` and asserting a formatted date appears rather than `—`. Must fail before the fix.
- No backend change — the API is correct; the client is wrong. Renaming the column instead would be the wrong direction.

## Notes

- Found while tracing `case_watchers` consumers for #553.
- Related but separate: the same card hardcodes an `Active` / "Receiving notifications" badge that is not derived from any data — see the watcher-preferences issue filed alongside this one.
- This is genuinely small and self-contained; a reasonable first frontend contribution.

---

*Analysis by Claude (Claude Code), reviewed by @craig-dt before posting. All file:line citations were verified against `main` at `311790e`.*

craig-dt:
## What changed

The watchers card always rendered `Watching since —`. The `Watcher` interface declared `added_at`, but `CaseWatcher.to_dict()` (`database/models.py:1098-1105`) emits:

```python
{"case_id": ..., "user_id": ..., "notification_preferences": ..., "created_at": ...}
```

There is no `added_at` on a watcher, so `fmtD` was always called with `undefined`.

```diff
-  added_at?: string
+  created_at?: string
```
```diff
-<div className="text-xs text-tx-faint">Watching since {fmtD(w.added_at)}</div>
+<div className="text-xs text-tx-faint">Watching since {fmtD(w.created_at)}</div>
```

**It failed quietly rather than loudly**, which is why it survived. `fmtD` guards its input — `if (!s) return '—'` — so the row rendered an em-dash rather than `Invalid Date`, with no console error, and the optional `?` on the interface field kept TypeScript satisfied. Nothing surfaced the mismatch at build time or runtime.

The wrong name is plausible because **`added_at` *is* a real column in this schema** — on the `case_findings` association table (`database/models.py:58`). It just was never on `case_watchers`.

This PR also fills in the two response fields the interface omitted (`case_id`, `notification_preferences`) so it describes what the API actually returns, and points the reader at `to_dict()` as the source of truth.

**Backend unchanged.** The API is correct and the client was wrong; renaming the column instead would have been the wrong direction.

## How this was tested

New `frontend/src/redesign/screens/cases/CaseSections.test.tsx`, following the mocking conventions in the neighbouring `CasesScreen.test.tsx`. **2 failed pre-fix, 3 pass post-fix.**

It renders `WatchersCard` against a payload shaped exactly like `to_dict()` and asserts the formatted date appears. It also asserts the em-dash placeholder is **absent** for a real timestamp — asserting only the date would still pass if the component rendered both.

The third test is a **regression guard, not new behaviour**: `created_at` is nullable, so a watcher without a timestamp must keep degrading to the placeholder rather than rendering `Invalid Date`. It passed before this change and must keep passing.

**Regressions:** full vitest suite **66 passed, 0 failed** across 13 files. `eslint` and `tsc --noEmit` both clean on the touched files.

No clean-`main` baseline worktree for this one, unlike the backend PRs: there were no failures to attribute, and worktree copies of the repo lack `node_modules` — `vitest.config.ts` excludes `**/worktrees/**` and `**/.claude/**` for exactly that reason.

## Related

Found while tracing `case_watchers` consumers for #553. The same card hardcodes an `Active` / "Receiving notifications" badge that is not derived from any data — that, and the fact that watcher notification preferences cannot currently be set at all, are #560. Out of scope here.

Closes #561

---

*Analysis by Claude (Claude Code), reviewed by @craig-dt before posting. All file:line citations were verified against `main` at `311790e`.*


nestor-deeptempo:
Rebased onto current `main` and reduced this to the test only.

Two things changed since this was opened:

- The fix itself landed in #584 (2026-08-07), so the `added_at` → `created_at` rename here is now a no-op against `main`. Dropped it, along with the `case_id` / `notification_preferences` interface additions — keeping this to the regression guard, which is the part still missing.
- The reorg (#481) moved `frontend/` to `clients/web/`, which is what made this conflict. The merge commit resolves that.

Three corrections to the test while I was in there:

- **TZ.** `fmtD` formats in local time, so `Jun 15, 2026` depended on where the suite ran — it failed under `TZ=Pacific/Honolulu`, where `09:14Z` rolls back a day. Pinned to UTC; verified under UTC, Honolulu and Kiritimati.
- **Source of truth.** The comments cited `CaseWatcher.to_dict()` in `database/models.py`. That method no longer exists — the endpoint now dumps `CaseWatcherSchema` (`core/storage/schemas/case_entities.py`), same four fields. Repointed.
- **Nullability.** The third test said the column is nullable, but `created_at` is `nullable=False` on the model. It's optional on the *schema*, which is the actual reason that path needs guarding.

Still verified to bite: reintroducing `added_at` fails 2 of the 3 tests.

The `to_dict()` → schema move is a good illustration of #699 — the interface tracked a Python method that got deleted underneath it, and nothing noticed.

nestor-deeptempo (approved):

Issue #469, opened 2026-07-29
Push raw-SQL API handlers down into services/repositories


nestor-deeptempo:
**Coding:** Med · **Vigil context:** Med · **Lane:** backend · **Good for:** mid-level dev

**Context.** About 11 API handlers issue raw SQL directly, bypassing the service/repository layer — so query logic lives in the web tier where it can't be reused or tested and drifts from the ORM models. Move that SQL down into service/repository methods so handlers stay thin and data access lives in one place.

**Acceptance criteria**
- No API handler issues raw SQL; each query moved into a service/repository method.
- Behavior unchanged; covered by a test where practical.

**Files:** the ~11 handlers in `backend/api/*`, target services in `services/`.

**Relates to:** #463 (same "service-layer discipline" theme).

G-r-ay:
Closing as **resolved by the repository restructure**.

This issue was scoped against the old `backend/api/*` tree ("~11 handlers issue raw SQL directly"). That tree no longer exists — handlers were migrated to `services/api/routers/*.py` and `core/<domain>/*_router.py`, and in that migration essentially all raw SQL was pushed out of the web tier.

A sweep of every current router for genuine raw-SQL strings (`text(...)`) finds only two sites, neither being the handler class this issue described:

- `services/api/routers/auth.py:223` — `text("SELECT pg_advisory_xact_lock(:key)")`. A bootstrap advisory lock (infrastructure), not query logic. Out of scope.
- `core/agents/agent_runs_router.py:95–109` — two real raw-SQL data queries inline in `get_run`. This was introduced by the in-flight agent-refactor work, so it postdates this issue; it'll be handled within that effort rather than here.

Everything else that reads as SQL-ish is either SQLAlchemy Core `select()` or ORM `.query()`, not raw SQL.

Note: inline **ORM** query logic still lives in some handlers (e.g. `core/cases/sla_policies_router.py`, `core/cases/case_metrics_router.py`, `services/api/routers/cases.py`). That's the broader service-layer-discipline theme (the #463 family), not the raw-SQL task this issue was scoped to. If we want to pursue it, it deserves its own issue.

Issue #559, opened 2026-08-04
Notification metadata silently dropped: create_notification passes metadata= but the column is notification_metadata


craig-dt:
**Coding:** Low · **Vigil context:** Low · **Lane:** backend · **Type:** bug (silent data loss)

**Context.** Every notification created through `CaseNotificationService` loses its metadata payload. `services/case_notification_service.py:60-69` constructs:

```python
notification = CaseNotification(
    ...
    metadata=metadata or {},
    ...
)
```

But the model's column is `notification_metadata` (`database/models.py:2300-2301`), carrying the comment:

```python
# Metadata (renamed from 'metadata' to avoid SQLAlchemy reserved word conflict)
notification_metadata: Mapped[Optional[dict]] = mapped_column(JSONB, nullable=True)
```

The rename happened; the constructor call was never updated. **It does not raise.** SQLAlchemy's declarative constructor rejects a kwarg only when `hasattr(type(self), key)` is false — and `CaseNotification.metadata` *does* exist: it is the declarative `MetaData` object inherited from `Base`. So `setattr` succeeds, creating an instance attribute that shadows the class attribute, and the payload never reaches a column.

Confirmed empirically against `main`:

```python
>>> n = CaseNotification(user_id='u', notification_type='t', title='x', message='y',
...                      metadata={'threshold_percent': 90})
>>> n.notification_metadata
None
>>> n.__dict__['metadata']
{'threshold_percent': 90}
```

Constructs cleanly, commits cleanly, `notification_metadata` is `NULL` for every row.

**Blast radius.** All four call sites in `services/case_notification_service.py`, so every notification the service creates:

| Site | Metadata lost |
|---|---|
| `:68` | whatever any caller passes to `create_notification` |
| `:127` | `{'assigned_by': ...}` — case assignment |
| `:181` | `{'comment_author': ..., 'comment_content': ...}` — comment mention |
| `:237` | `{'threshold_percent': ..., 'sla_type': ...}` — SLA warning |

This includes everything routed through `notify_watchers`, since it delegates to `create_notification`.

`daemon/orchestrator.py:1097` gets it right (`notification_metadata={"investigation_id": inv_id}`), which is why the column is not empty in *every* deployment — only for the notifications this service creates.

Not affected: the `metadata=` kwargs in `backend/api/timeline.py` target `TimelineEvent`, a Pydantic `BaseModel` with a real `metadata` field. Those are legitimate.

## Fix

[...]

PR #566, opened 2026-08-04
Add per-request unit-of-work boundary (#464)

### LINKED ISSUE #464: Unit-of-work transaction boundary (one per request)

G-r-ay:
Closes #464. **Stacked on #529** (needs `services/unit_of_work.py` from #463) — base
will retarget to `main` once #529 merges. Review the [2-commit
range](https://github.com/Vigil-SOC/vigil/compare/refactor/463-case-repository-seam...refactor/464-unit-of-work-boundary)
or just the second commit.

## Problem

There was no transaction boundary per request. Handlers and services called
`.commit()`/`.rollback()` wherever convenient, so a multi-step write that failed
halfway left the database partially written.

The issue's "74 commits / 57 rollbacks across `services/`" is stale — earlier PRs
cleaned much of it and what remained had **moved into the API layer**:

| Location | `.commit()` | `.rollback()` |
|---|---|---|
| `backend/api/` + `backend/services/` | 33 | 25 |
| `services/` (non-case) | 4 | 3 |

Session acquisition was split three ways: `Depends(get_db)` (closed the session but
never committed — hence the inline commits), direct `get_db_session()` in handler
bodies, and the `case_*` services already on #463's seam.

## Solution

`backend/dependencies.py` wraps #463's `unit_of_work()` as a request dependency:

```python
def request_unit_of_work() -> Generator[Session, None, None]:
    with unit_of_work() as session:
        yield session

UnitOfWorkSession = Annotated[Session, Depends(request_unit_of_work, scope="function")]
```

Handlers take `db: UnitOfWorkSession` and never commit, roll back, or close. It
lives under `backend/` so `services/unit_of_work.py` stays framework-free for the
daemon. 17 files migrated; `get_db` **deleted** from `database/connection.py` so the
non-committing version can't be reused. Non-request paths (`auth_service`,
`case_automation_service`, `sandbox_correlation_service`, `openai_agent_service`)
use `with unit_of_work(session)` directly.

## ⚠️ Why `scope="function"` — please don't remove it

With the **default** `scope="request"`, dependency teardown runs *after the response
is sent*, so a failed commit returns **HTTP 200 with the success body while the write
is rolled back** — silent data loss, strictly worse than the scattered commits this
replaces. Measured both ways on a throwaway app:

```
/fn-scope  (scope="function") -> HTTP 500                  ✅
/req-scope (default)          -> HTTP 200 {"wrote":true}   ❌ DB rolled back
```

`Depends(scope=...)` first exists in **fastapi 0.129.0** (0.128.8 lacks it), hence
the pin bump — that is the pin's only purpose.
`test_request_unit_of_work.py::test_failed_commit_is_reported_as_an_error` is the
regression guard.

Teardown also rolls back on `HTTPException`, which is what lets handlers drop their
`except: rollback; raise HTTPException` boilerplate. I audited all 13 commit-then-raise
sites first: every one is the benign `try/commit/except/rollback/raise 500` shape, so
no endpoint relied on committing and *then* raising.

## Migration rules applied

- `db: Session = Depends(get_db)` → `db: UnitOfWorkSession`
- Delete `.commit()`/`.rollback()`/`finally: close()`, **keep** every `try/except →
  HTTPException` so error contracts and status codes are unchanged
- `commit()` followed by `refresh()`/`to_dict()` → `flush()`, so server defaults
  still land while the boundary owns the commit
- Where a method **swallows** its exception (`except: rollback; return None`), the
  `try/except` goes **outside** the `with` — otherwise the boundary sees no failure
  and commits partial state. Affects `auth_service` and `sandbox_correlation_service`.

## Verification

- `tests/unit`: **1067 passed, 4 failed** — the same 4 pre-existing API-key-leak
  flakes that fail on `main` and pass in isolation (55 passed alone)
- New tests: `test_request_unit_of_work.py` (commits once per request; multi-step
  rollback; `HTTPException` rollback; failed-commit-is-a-500) and
  `TestAttachReportAtomicity` (1 evidence row + N IOC rows — all-or-nothing), which
  is the "representative multi-step operation" the issue asks for
- App boots; **293 OpenAPI paths**; no `session`/`db` parameter leaked into the
  schema — the giveaway for a botched `Annotated` conversion
- Removing a `try:`/`with` forces a 4-space dedent whose diff `git diff -w` hides, so
  I **AST-compared** every mechanically-edited file against HEAD: `case_metrics`
  12/12, `sla_policies`, `auth_service` 24/24, `case_automation` 15/15 — logic
  identical apart from the intended changes
- Lint: every touched file was *already* black/isort-dirty on `main` under current
  tooling and is dirty the same way now — no new debt, and I deliberately did not
  reformat them

## Deliberately left alone

- **`merge_cases`** (`backend/api/cases.py:1060`) keeps its own `session_scope()`: it
  commits and then re-reads via `data_service.get_case()` on a *different* session,
  so deferring the commit would return stale pre-merge data. Fixing it means moving
  the read-back onto the same session — #469-flavored, wants its own issue.
- **`database/config_service.py`** (7 commits) — self-contained, session-per-method,
  not request-scoped.

## Reviewer question

`case_automation_service` sweeps now commit **once per sweep** instead of per item.
That's the atomicity the issue asks for, but it changes failure granularity: one bad
row loses that sweep's batch. They re-run on a timer so it self-heals — say the word
if you'd rather keep per-item boundaries.


craig-dt:
Heads-up on two things I hit while checking this against some open PRs of mine — one is a live problem independent of my work, the other is a merge-order trap worth knowing about before this lands.

## 1. The base is stale, and the "no conflicts" status is misleading

GitHub currently shows this PR as mergeable/clean. It isn't — merging it into `main` today conflicts on `backend/api/findings.py`:

```
git fetch origin main refs/pull/566/head:pr566
git checkout -B t origin/main && git merge --no-ff pr566
# CONFLICT (content): Merge conflict in backend/api/findings.py
```

The cause is the base: `git merge-base pr566 origin/main` is `d854e70`, which predates **#518** (`fix(api): offload blocking I/O off the event loop`, now `cf2cfbb` on `main`). #518 rewrote `findings.py`, so the two sets of changes to that file overlap. A merge/rebase onto current `main` resolves it, but it's worth doing deliberately rather than discovering it at merge time — #518's whole point was moving blocking calls off the event loop, so that's a resolution you want to be conscious about rather than take-ours/take-theirs.

## 2. This PR carries a known data-loss bug forward verbatim

`services/case_notification_service.py` is rewritten here (−142/+70) to wrap the session in `unit_of_work(...)`. Inside the rewritten `create_notification`, line 65 of this branch's version reads:

```python
metadata=metadata or {},
```

That kwarg is the subject of **#559**, fixed in **#563**. `CaseNotification`'s column is `notification_metadata` (`database/models.py:2279-2280`) — renamed to dodge SQLAlchemy's reserved `metadata` attribute. Passing `metadata=` to the declarative constructor doesn't raise and doesn't write a column; it lands on the declarative `MetaData` and is discarded. The insert commits, returns a row, logs success, and the payload is gone. Every caller in that file that passes a metadata dict is affected.

Not a criticism of this PR — it faithfully preserved the line it found. The risk is purely in the merge: because this branch re-types that line inside a restructured block, resolving the #563 conflict by "take the newer refactor" **silently reintroduces the bug**. If #566 lands first I'll re-apply the one-line fix on top; I mostly want it on the record so the resolution isn't made blind.

## 3. Overlap with open PRs — happy to go second

Trial-merged locally against `main`; genuine overlaps only:

| PR of mine | Overlapping files |
|---|---|
| #563 | `services/case_notification_service.py` (the above) |
| #558 | `services/case_notification_service.py` (`notify_watchers` + a module constant) |
| #535 | `backend/api/{ai_config,auth,sla_policies,users}.py` |

(`findings.py` shows up against all nine of my branches, but that's item 1 above, not overlap with my changes.)

All three of mine are small and independent. **I'd rather this PR go first** and rebase mine onto it — a 40-file architectural seam shouldn't queue behind a one-line kwarg fix. Just let me know if you'd prefer the opposite order for any of them.

---

*Analysis by Claude (Claude Code), reviewed by @craig-dt before posting. Conflict claims were reproduced locally against `main` at `311790e` and `refs/pull/566/head`; the `findings.py` and `case_notification_service.py` citations were verified in this branch's own tree.*

mattmorr1:
@G-r-ay pls fix merge conflicts - otherwise lgtm

PR #565, opened 2026-08-04
reorg(clients): frontend/ → clients/web/, desktop/ → clients/desktop/

nestor-deeptempo:
Creates the `clients/` top-level directory from the #481 target tree by moving the two TypeScript clients out of the repo root. This is the one reorg piece that shares zero files with the Python `core/` work, so it lands independently of the in-flight `core/llm/` (R4) PR.

## Moves (history preserved via `git mv`)

- `frontend/` → `clients/web/`
- `desktop/` → `clients/desktop/`

## Repoints (no shims, fail closed)

- **Runtime / Docker:** backend SPA `StaticFiles` mount + `Dockerfile.backend` build/runtime stages → `clients/web/build`.
- **Release / CI:** `release-please-config.json`, `ci-cd.yml` (working dirs, cache paths, build artifact), `release.yml` image smoke test.
- **Dev / setup:** `start.sh`, `setup_dev.sh`, `scripts/{app_up,app_setup,reset-setup}.sh`.
- **Client internals (relative-path depth shift):** `clients/web/vite.config.ts` root-`.env` loader (`..` → `../..`, how `DEV_MODE` reaches the SPA), `vitest.config.ts` `.claude` exclude glob, `clients/desktop/src/main.ts` web references, and the `bundle-image.sh` / `prepare-standalone.js` repo-root `..` walks (one level deeper now).
- **Ignore / config / docs:** `.gitignore`, `.dockerignore`, `pyrightconfig.json`, `integration_secrets` / `integration_bridge_service` docstrings, and the docs (including two relative markdown links that would otherwise break).

## Verification

- Web client builds from `clients/web` (`tsc` + `vite`), output at `clients/web/build/`.
- Backend resolves and finds the SPA at `clients/web/build/index.html` (+ assets).
- Desktop TypeScript compiles with the repointed `main.ts`.
- Full stale-reference sweep is clean — the only remaining matches are the `/frontend/status` HTTP route and the "desktop/dev" prose in `secrets-audit.md`, neither of which is a filesystem path.

The Docker image build is left to CI — `release.yml` runs the exact `/app/clients/web/build/index.html` smoke test, and the Dockerfile stages are internally consistent (`COPY clients/web/` → `/build/web/build` → runtime `./clients/web/build/` → served by the backend).

Draft while the reorg base settles — ready for review.

Closes #507.

PR #575, opened 2026-08-05
refactor: replace deprecated datetime.utcnow() with naive-UTC helper

### LINKED ISSUE #541: Replace deprecated datetime.utcnow() / asyncio.get_event_loop() (repo-wide)
## Problem

Two standard-library calls that Python has deprecated are used throughout the codebase:

| Call | Status | Spread on `main` |
|---|---|---|
| `datetime.utcnow()` | Deprecated in 3.12 | **266 occurrences across 57 files** |
| `asyncio.get_event_loop()` | Deprecated in 3.12; raises without a running loop | **10 occurrences across 5 files** |

Both still work today but are on the removal path, and `get_event_loop()`'s behavior change already bites us — see #442 (event-loop teardown failure on Py3.13).

## Fix

A mechanical sweep, best done as a series of small PRs rather than one mega-diff:

- **`datetime.utcnow()` → `datetime.now(timezone.utc)`.** Careful: `utcnow()` returns a *naive* datetime and `now(timezone.utc)` returns an *aware* one. Several call sites append `'Z'` to `.isoformat()` by hand (e.g. the finding-enrichment metadata stamp) — those need to be normalized so serialization doesn't end up with a double offset.
- **`asyncio.get_event_loop()` → `asyncio.get_running_loop()`** inside `async` contexts (drop-in for the `run_in_executor` sites).

## Scope

Repo-wide `tech-debt`. Good candidate to split: the `utcnow` portion is largely find-and-replace (`good first issue`-friendly), the `get_event_loop` portion is smaller but needs more care and should land with / reference #442.

Surfaced during review of #537 (both patterns are carried, unchanged, into the extracted enrichment module).

arimu1:
## Summary

Repo-wide replacement of deprecated `datetime.utcnow()` (Python 3.12+) with a **naive-UTC helper**, so we kill the deprecation without mixing aware `now` against naive `DateTime` columns (option A from review).

```python
# core/time.py
def utcnow() -> datetime:
    return datetime.now(timezone.utc).replace(tzinfo=None)
```

Also folds in the last three `asyncio.get_event_loop()` sites so #541 can close.

### Changes
| Metric | Count |
|--------|------:|
| Files changed | **61** |
| `from core.time import utcnow` | **58** |
| Remaining `datetime.utcnow()` in `core/` `services/` `tools/` | **0** (helper exempt) |
| `asyncio.get_event_loop()` → `get_running_loop()` | **3** |

### Replacement rules
- `datetime.utcnow()` → `utcnow()` from `core.time`
- SQLAlchemy `default=` / `onupdate=` stay callables (`default=utcnow`)
- Existing `datetime.now(timezone.utc)` sites that were already aware are left alone
- `isoformat() + "Z"` / f-string `{isoformat()}Z` on the naive helper is correct (must not emit `+00:00Z`)
- Lockout comparisons (`auth_service` / `routers/auth`) stay naive-naive

### `get_event_loop` sites
- `core/findings/enrichment/service.py` (`run_in_executor`)
- `core/reporting/ai_insights_service.py` (same)
- `services/daemon/main.py` `_setup_signal_handlers` (called from `async def run`)

### Ratchet
`tests/unit/_ratchets/test_no_datetime_utcnow.py` — same AST style as `test_no_ambient_state.py`. Scans `core/`, `services/`, `tools/`; exempts `core/time.py`. Also forbids `asyncio.get_event_loop()`.

## Test plan
- [x] `pytest tests/unit/_ratchets/test_no_datetime_utcnow.py -q --no-cov`
- [x] `pytest tests/unit/auth/test_auth_lockout_persistence.py -q --no-cov`
- [x] daemon / llm-log / ai-config unit slices — **28 passed, 29 skipped** (skips are pre-existing daemon architecture skips)
- [ ] Full `./tests/run-tests.sh` (not run locally)

## Related
- Fixes #541

### DIFF
--- a/core/auth/auth_service.py
+++ b/core/auth/auth_service.py
@@ -10,6 +10,7 @@ import logging
 import secrets
 import uuid
 from datetime import datetime, timedelta
+from core.time import utcnow
 from typing import Optional, Dict, Any, List
 import bcrypt
 import jwt
@@ -151,7 +152,7 @@ class AuthService:
             else timedelta(days=JWT_REFRESH_EXPIRATION_DAYS)
         )
 
-        now = datetime.utcnow()
+        now = utcnow()
         payload = {
             "user_id": user.user_id,
             "username": user.username,
@@ -238,7 +239,7 @@ class AuthService:
                 # Reject while locked. Lockout is authoritative even over a
                 # correct password — otherwise an attacker who eventually
                 # guesses right would bypass the wait.
-                now = datetime.utcnow()
+                now = utcnow()
                 if user.locked_until and user.locked_until > now:
                     logger.warning(
                         "Login rejected, account locked: %s until %s",
--- a/core/cases/case_automation_service.py
+++ b/core/cases/case_automation_service.py
@@ -6,7 +6,8 @@ Handles SLA monitoring, auto-assignment, escalation, and periodic tasks.
 
 import logging
 import asyncio
-from datetime import datetime, timedelta
+from datetime import timedelta
+from core.time import utcnow
 from typing import Dict, List
 
 from core.storage.unit_of_work import unit_of_work
@@ -71,7 +72,7 @@ class CaseAutomationService:
                     CaseSLA.is_paused == False
                 ).all()
 
-                current_time = datetime.utcnow()
+                current_time = utcnow()
 
                 for sla in active_slas:
                     # Get SLA status
@@ -148,7 +149,7 @@ class CaseAutomationService:
         try:
             with unit_of_work() as session:
                 # Define stale threshold (7 days)
-                stale_threshold = datetime.utcnow() - timedelta(days=7)
+                stale_threshold = utcnow() - timedelta(days=7)
 
                 # Find cases not updated recently
                 stale_cases = session.query(Case).filter(
@@ -178,7 +179,7 @@ class CaseAutomationService:
         while self.running:
             try:
                 # Calculate time until next 9 AM
-                now = datetime.utcnow()
+                now = utcnow()
                 next_run = now.replace(hour=9, minute=0, second=0, microsecond=0)
                 if now.hour >= 9:
                     next_run += timedelta(days=1)
@@ -199,7 +200,7 @@ class CaseAutomationService:
         try:
             with unit_of_work() as session:
                 # Get metrics for last 24 hours
-                yesterday = datetime.utcnow() - timedelta(days=1)
+                yesterday = utcnow() - timedelta(days=1)
                 metrics = self.metrics_service.get_dashboard_metrics(
                     start_date=yesterday,
                     session=session
--- a/core/cases/case_evidence_service.py
+++ b/core/cases/case_evidence_service.py
@@ -6,7 +6,7 @@ Handles file storage, chain of custody, and evidence tracking.
 
 import logging
 import hashlib
-from datetime import datetime
+from core.time import utcnow
 from pathlib import Path
 from typing import Dict, List, Optional
 from sqlalchemy.orm import Session
@@ -100,7 +100,7 @@ class CaseEvidenceService:
 
                 # Initialize chain of custody
                 chain_of_custody = [{
-                    'timestamp': datetime.utcnow().isoformat(),
+                    'timestamp': utcnow().isoformat(),
                     'action': 'collected',
                     'user': collected_by,
                     'notes': 'Evidence collected and added to case'
@@ -117,7 +117,7 @@ class CaseEvidenceService:
                     file_hash_sha256=file_hash_sha256,
                     source=source,
                     collected_by=collected_by,
-                    collected_at=datetime.utcnow(),
+                    collected_at=utcnow(),
                     chain_of_custody=chain_of_custody,
                     tags=tags or []
                 )
@@ -164,7 +164,7 @@ class CaseEvidenceService:
 
                 # Add new entry to chain of custody
                 entry = {
-                    'timestamp': datetime.utcnow().isoformat(),
+                    'timestamp': utcnow().isoformat(),
                     'action': action,
                     'user': user,
                     'notes': notes or ''
--- a/core/cases/case_ioc_service.py
+++ b/core/cases/case_ioc_service.py
@@ -7,6 +7,7 @@ Handles IOC tracking, enrichment, deduplication, and export.
 import logging
 import json
 from datetime import datetime
+from core.time import utcnow
 from typing import Dict, List, Optional
 from sqlalchemy.orm import Session
 from sqlalchemy import and_
@@ -81,8 +82,8 @@ class CaseIOCService:
                     threat_level=threat_level,
                     confidence=confidence,
                     source=source,
-                    first_seen=first_seen or datetime.utcnow(),
-                    last_seen=last_seen or datetime.utcnow(),
+                    first_seen=first_seen or utcnow(),
+                    last_seen=last_seen or utcnow(),
                     tags=tags or [],
                     context=context,
                     is_active=True,
@@ -267,7 +268,7 @@ class CaseIOCService:
                     'modified': ioc.updated_at.isoformat() + 'Z',
                     'pattern': f'[{stix_type}:value = \'{ioc.value}\']',
                     'pattern_type': 'stix',
-                    'valid_from': (ioc.first_seen.isoformat() + 'Z') if ioc.first_seen else datetime.utcnow().isoformat() + 'Z'
+                    'valid_from': (ioc.first_seen.isoformat() + 'Z') if ioc.first_seen else utcnow().isoformat() + 'Z'
                 }
 
                 if ioc.threat_level:
--- a/core/cases/case_metrics_service.py
+++ b/core/cases/case_metrics_service.py
@@ -6,6 +6,7 @@ Handles MTTD, MTTR, MTTA, SLA compliance, and analyst performance metrics.
 
 import logging
 from datetime import datetime, timedelta
+from core.time import utcnow
 from typing import Dict, Optional
 from sqlalchemy.orm import Session
 from sqlalchemy import and_
@@ -199,7 +200,7 @@ class CaseMetricsService:
             )
 
             # Cases opened per day (last 30 days)
-            thirty_days_ago = datetime.utcnow() - timedelta(days=30)
+            thirty_days_ago = utcnow() - timedelta(days=30)
             recent_cases = session.query(Case).filter(
                 Case.created_at >= thirty_days_ago
             ).all()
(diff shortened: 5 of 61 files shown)


mattmorr1:
@arimu1 great change! make sure it passes our tests though.

arimu1:
@mattmorr1 Thanks — fixed the DB-backed unit failure.

Root cause: after the `utcnow` → `datetime.now(timezone.utc)` sweep, `WorkflowRunService.finalize_run` subtracted aware `now` from naive `started_at` (SQLAlchemy `DateTime` without `timezone=True` strips tzinfo on read from SQLite/Postgres).

Tip normalizes naive DB timestamps to UTC before duration math (`workflow_run_service` + SLA resume pause duration). Re-run should be green.

arimu1:
Rebased onto current `main` (soft re-apply after conflicts with Pydantic ORM schemas / UoW / dead-code cleanup).

- Tip: `f317473b687569f362a06a92dd5017d24118b14a`
- Mechanical `datetime.utcnow()` → `datetime.now(timezone.utc)`
- Naive DB datetime UTC normalization kept for duration math (`WorkflowRunService.finalize_run`, `CaseSlaService` resume)

nestor-deeptempo:
@arimu1 Thanks for your patience on this, and sorry it's been sitting. We had a major file refactor while this was open, so the changes need to be redone. We'd still like you to be a contributor! So if you're open could you rebase from main and add the proposed changes that I'm suggesting below? If not I can take on this change since you were caught in the refactor. LMK!

## The design question worth settling before a redo

Reviewing the diff surfaced something we should decide as a project rather than leave to the sweep. All 104 `DateTime` columns in `core/storage/models.py` are timezone-naive, and the init SQL is 61 naive `TIMESTAMP` against 7 `TIMESTAMPTZ`. So "make `now()` aware" and "leave the columns naive" are in tension, here's my suggestion:

A naive-UTC helper** — `utcnow()` returning `datetime.now(timezone.utc).replace(tzinfo=None)`. An exact drop-in that kills the deprecation with zero semantic change, at the cost of not gaining awareness.

## Specifics found in review

These are worth capturing regardless of who does the redo, especially the first group.

**Naive/aware comparisons.** You caught two of these when CI went red (`finalize_run` and the SLA resume) and fixed them locally, but the pattern is broader than those two sites. On current `main` there are five places where an aware `now` would meet a naive value:

- `core/auth/auth_service.py:242` — `user.locked_until > now`, where `locked_until` is a naive `DateTime` column. The `TypeError` lands in the enclosing `except Exception: return None`, so account lockout silently degrades to a generic 401 and `AccountLockedError` never fires. Fails closed, but the feature goes dead without any test noticing.
- `services/api/routers/auth.py:294` — `exc.locked_until - now`, same source.
- `services/daemon/orchestrator.py:531` — `now - last_activity`, which is either a DB value or a `fromisoformat` parse.
- `services/daemon/sandbox_poller.py:266` — `now - submitted`, parsed from a stored ISO string, so in-flight submissions break across the upgrade.
- `core/cases/case_sla_service.py:421` and `core/workflows/workflow_run_service.py:124` — the two you already fixed.

**Double-offset serialization.** `services/daemon/plan_generator.py:148` and `:216` still do `f"...{now.isoformat()}Z"`, which emits `+00:00Z` into the plan-file frontmatter. The PR description says these were normalized, but they weren't — worth a grep for the f-string form specifically, since it doesn't match the `+ "Z"` pattern.

**New sites since your base.** Three files with `utcnow` didn't exist when you branched: `tools/mcp/deeptempo_findings.py` (10 sites), `core/integrations/splunk/ingestion.py` (4), and `core/workflows/run_bridge_router.py` (2). Current `main` is at 268 occurrences across 58 files.

**A ratchet.** 16 new sites appeared in the 11 days after this PR opened, so without a guard it regrows faster than it can be swept. Something like `tests/unit/_ratchets/test_no_datetime_utcnow.py`, following the existing `test_no_ambient_state.py` pattern, would make this the last time anyone has to do it.

**The other half of #541.** `asyncio.get_event_loop()` is down to 3 sites — `core/findings/enrichment/service.py:110`, `core/reporting/ai_insights_service.py:242`, `services/daemon/main.py:59`. Cheap enough to fold in so the issue actually closes.

## How you'd like to proceed

Happy either way:

1. **You redo it** — reset onto current `main` and re-apply. This PR keeps its number, thread, and the `Fixes #541` link. We'd settle the A/B/C question here first.
2. **We take it over** — `maintainerCanModify` is on, so we can push to the branch directly. We'd keep you as commit author and preserve your `Signed-off-by`, so the credit stays with you.

No pressure, if I don't hear back in the next day or two I'll pick it up. Thanks again for taking this on, the diagnosis in your earlier comment about `finalize_run` was good, and it's the thing that made the broader problem visible.

arimu1:
@nestor-deeptempo Thanks — I'll redo this on current `main`.

Going with the naive-UTC helper you suggested (`utcnow()` = `datetime.now(timezone.utc).replace(tzinfo=None)`) so we kill the deprecation without mixing aware `now` against naive columns.

I'll also fold in:
- remaining naive/aware comparison sites
- `isoformat()Z` double-offset in `plan_generator`
- new `utcnow` sites since the last rebase
- a ratchet test like `test_no_ambient_state.py`
- the last 3 `asyncio.get_event_loop()` sites so #541 can actually close

Please don't push over the branch — I'll force-with-lease once the redo is in.

craig-dt:
@nestor-deeptempo — two things, one small and one that might unblock this.

**The small one:** the comment above refers to settling "the A/B/C question", but only option A made it into the text (there's a stray `**` right after "A naive-UTC helper" where B and C look like they were meant to go). So the decision that's gating the redo isn't actually written down anywhere. Since @arimu1 is being asked to wait on it, that seems worth fixing before the clock runs out on them.

I've written out how I'd frame the three, below — entirely so there's something concrete to react to. Correct me where I've guessed wrong about what you had in mind.

## The three options as I'd frame them

**A — Naive-UTC helper.** A shared `utcnow()` returning `datetime.now(timezone.utc).replace(tzinfo=None)`. Exact drop-in, kills the deprecation, zero semantic change. None of the five naive/aware comparison sites you listed can fire, because nothing becomes aware. Cost: no timezone correctness gained, and it makes naive-UTC the explicit house style rather than an accident.

**B — Aware end to end.** Python *and* the schema: `DateTime(timezone=True)` on the models, `TIMESTAMPTZ` in the init SQL. The correct end state, and the only one where the column types stop lying about their contents. Cost: a migration across 104 columns and 61 SQL types, against databases that already hold data. Much bigger than the sweep it's attached to, and probably its own issue.

**C — Aware in Python, naive at the storage boundary.** Convert on read/write in the storage layer. Business logic gets awareness without a migration. Cost: a conversion layer that has to be right at every boundary, and the columns still don't describe what's in them.

My read, for what it's worth: **A and B are each coherent; C is the one that tends to rot**, because the conversion layer is invisible at the call sites that depend on it. And the choice mostly turns on one question — is the goal timezone correctness, or just retiring the deprecation before 3.15? If it's the latter, A does the whole job at a fraction of the risk, and B stays available later as a deliberate schema change rather than a side effect of a sweep.

## On whether a redo is safe to start now

The specific hazard that caught @arimu1 looks closed. The reorg wave is fully merged — #525, #530, #557, #565 and #580, the last of those on 5 Aug — and `main` now has `clients/` and `infra/` with no `backend/`, `frontend/`, `daemon/` or `database/` left at the top level. So a redo shouldn't get moved out from under itself the same way twice.

The remaining hazard is plain conflict, not relocation. Three large PRs are open right now: **#665** (161 files), **#700** (132) and **#655** (68). A 64-file sweep will collide with them textually regardless of who writes it.

One correction to the regrowth argument, since it changes the ordering: those three are roughly **net-neutral** on `utcnow` today — #655 is −2, #700 is +1, #665 is 0. The 16-new-sites-in-11-days figure came from merged work, not from what's currently in flight. The ratchet is still the right call, but the case for it is "stop the next one" rather than "the open PRs are actively making it worse."

Which argues for **landing the ratchet first**, before the sweep — `xfail`ed or thresholded at the current count — so it can't regrow underneath a long-running branch, and so the sweep itself becomes the change that flips it green. `tests/unit/_ratchets/` already has ten of these, so the pattern's well established. Worth noting `test_one_loop_gate.py` doesn't help here despite the name: it gates agent tool-calling loops, not `asyncio`.

## Numbers, re-checked against f61607d

Everything in your review holds up:

| | |
|---|---|
| `DateTime` columns in `core/storage/models.py` | 104, **0** with `timezone=True` |
| init SQL | 7 aware, 61 naive — exactly as you had it |
| `asyncio.get_event_loop()` | 3 sites, as listed |

One clarification worth pinning down since both numbers are now in the thread: `datetime.utcnow()` is **266 occurrences on 179 distinct lines across 58 files**. Your 268 counts occurrences, my first pass counted lines. Both right, different units — but it matters for sizing, because the sweep is 179 edits rather than 268.

## One item that should probably leave this thread

`core/auth/auth_service.py:242` deserves its own issue regardless of which option wins, and regardless of whether this PR ever lands. The `user.locked_until > now` comparison sits inside an `except Exception: return None`, so the moment `now` becomes aware, the `TypeError` is swallowed and **account lockout silently stops working** — degrading to a generic 401, with `AccountLockedError` never raised and no test noticing. It fails closed, so it isn't an exposure, but a security control going quietly dead is worth tracking on its own rather than as a footnote to a deprecation sweep. Happy to file it if it isn't already covered.

## Offer

No wish to step on @arimu1, whose `finalize_run` diagnosis is what made the real problem visible here, and it's their PR. But if the "day or two" has passed and you'd rather not spend your own time on a mechanical sweep, I'm happy to take a slice of it — most usefully the ratchet plus the `asyncio.get_event_loop()` half, which is only 3 sites and can land independently of whichever A/B/C answer you settle on.

Just say which option you want and whether you'd like the help.

arimu1:
@nestor-deeptempo Redone on current main as the naive-UTC helper.

`core.time.utcnow()` is `datetime.now(timezone.utc).replace(tzinfo=None)` — drop-in for naive DateTime columns, so lockout/SLA/finalize_run stay naive-naive and `isoformat()Z` does not emit `+00:00Z`. Existing aware `datetime.now(timezone.utc)` sites were left alone.

61 files (58 helper imports + `core/time.py` + ratchet). Last three `asyncio.get_event_loop()` sites are now `get_running_loop()` (enrichment, ai_insights, daemon `_setup_signal_handlers`). Ratchet at `tests/unit/_ratchets/test_no_datetime_utcnow.py` scans core/services/tools.

Tests: ratchet, lockout, daemon, llm log, ai-config — 28 passed, 29 skipped (pre-existing). Full `./tests/run-tests.sh` not run.

Tip: `af41f1ef9a2fc67190d625baea727bfd64b17d25`

nestor-deeptempo (approved):
@arimu1 thank you for your persistence! changes lgtm and @craig-dt i'll follow up with creating the issue you mentioned

Issue #467, opened 2026-07-29
Typed request/response models on all endpoints


nestor-deeptempo:
**Coding:** Med · **Vigil context:** Med · **Lane:** backend · **Good for:** mid-level dev comfortable with the data model; parallelizable per router (not a solo first-PR — see risk note)

**Context.** Only ~49 of ~339 endpoints declare a typed model; the rest return hand-rolled `to_dict()` output, often wrapped in ad-hoc envelopes like `{"comments": [...]}`, so the API is untyped at the edges — no input validation, no guaranteed output shape, inaccurate OpenAPI. Typing them makes the HTTP contract explicit and checkable. Splits cleanly per router.

**Sequencing & risk.**
- **Do after #465.** #465 turns the `to_dict()` outputs into Pydantic ORM-mode schemas; once those exist, this issue is mostly "attach the schema as `response_model=`." Done first, it becomes reverse-engineering ~290 response shapes from scratch — much harder.
- **`response_model` filters the response.** FastAPI silently drops any field not declared on the schema — so a too-narrow model removes fields the frontend may rely on. Every endpoint needs an output-parity check (diff the response before/after), which is why this isn't a safe solo first-PR.

**Acceptance criteria**
- Every endpoint declares request + response Pydantic models (`response_model=` set).
- Input validated via the request model; output shaped via the response model.
- **Output parity verified** per endpoint (serialized response unchanged unless a drift bug is being deliberately fixed).
- Split per router so several contributors can share it.

**Files:** `backend/api/*`, `backend/schemas/`.

**Blocked by / do after:** #465 (its ORM-mode schemas). Complements the models split #419.

samarmstrong:
🏭 **Factory groom — needs work.** The factory is not picking this up as written. Edit the description or reply here and it gets reviewed again on a later tick.

This issue asks for typed request and response models on every remaining endpoint. That is not a change one PR can land, and the piece that would fit is not this issue’s point.

[...]

PR #579, opened 2026-08-05
Replace hand-rolled to_dict() with Pydantic ORM-mode schemas

### LINKED ISSUE #465: Replace hand-rolled to_dict() with Pydantic ORM-mode schemas
**Coding:** Med · **Vigil context:** Med · **Lane:** backend · **Good for:** mid-level dev; parallelizable per domain

**Context.** The API's response shapes are defined by ~42 hand-written `to_dict()` methods on the ORM models, and only ~49 of ~339 endpoints declare a typed `response_model`. So the real API contract is implicit and drifts: rename or add a model field and the serialized output can silently change, or a value can quietly vanish from responses, with nothing to catch it (the same failure mode behind the case↔finding bug in #463). Replacing `to_dict()` with per-model Pydantic ORM-mode (`from_attributes`) schemas makes the contract explicit and type-checked and removes that whole bug class. The work splits cleanly per domain, so several contributors can share it.

**Acceptance criteria**
- A Pydantic ORM-mode schema per model, replacing its `to_dict()`.
- Serialized output parity verified (shape unchanged except where a drift bug is being deliberately fixed).
- Work split per domain so several contributors can share it.

**Files:** `database/models.py`, `backend/schemas/`, serialization call sites.

**Pairs with:** #463 and #464; feeds the later "typed request/response on all endpoints" work.

G-r-ay:
Closes #465. Stacked on #566 (#464); GitHub retargets it to `main` once the base merges.

## What & why

The API's JSON contract was defined by 42 hand-written `to_dict()` methods on the ORM models, type-checked by nothing. Renaming or adding a column could silently change a response field or drop one — the read-direction twin of the case↔finding bug from #463. This replaces them with one Pydantic `from_attributes` schema per model, so each response shape is declared once, in one place, and checked.

## What's here

- **`database/schemas/`** — 45 schemas over all 42 models (`base.py` + 7 domain modules). They live beside the models, not under `backend/`, so the daemon and service layer don't import the web layer to serialize.
- **`ORMSchema.dump()` / `dump_many()`** — the serialization entry point. `dump_summary()` / `dump_with_secrets()` replace the old `include_*` flags.
- **All ~100 in-scope call sites migrated**, then the 42 `to_dict()` + 2 `to_summary_dict()` methods deleted from `database/models.py`.

## Serialized-output parity (AC #2)

`tests/fixtures/orm_serialization_golden.json` is a frozen capture of the pre-refactor `to_dict()` output over deterministic in-memory instances (no DB — these models can't instantiate on SQLite). `test_orm_schema_parity.py` asserts each schema's dump equals that capture key-for-key, across 100 shapes, populated and bare. **It is not regenerated from the schemas** — that would be circular. Two guards keep it exhaustive: every model has a registered schema, and every captured shape is asserted.

`test_schema_secret_redaction.py` covers the omissions that are security-relevant rather than cosmetic (user credential columns, the provider key reference) — parity alone would let someone widen those and pass by regenerating the fixture.

## Load-bearing details preserved

- `Case` → two schemas (`CaseSchema` with `finding_ids`, `CaseWithFindingsSchema` inlined). `finding_ids` derives from the `findings` relationship via `validation_alias`, deliberately **not** a model property, so `hasattr(case, "finding_ids")` stays false and the #463 trap isn't re-armed.
- `LLMProviderConfig.dump()` **redacts** `api_key_ref` (reports `has_api_key`); the reference needs `dump_with_secrets()`. Redacting by default means a missed call site under-reports rather than leaks.
- `Finding.dump_summary()` drops the 768-float embedding for list responses.
- `CaseNotification.notification_metadata` still serializes as `"metadata"` (SQLAlchemy reserves the name) via `validation_alias`.
- Datetimes serialize via an explicit `.isoformat()` serializer — Pydantic renders tz-aware UTC as `...Z` where the contract emits `...+00:00` (latent today; all `DateTime` columns are naive).

## Out of scope (deliberately untouched)

`WorkflowDefinition`, `CostEstimate`, and `ModelInfo`/registry dataclasses have their own `to_dict()` and are not ORM models.

## Verification

- Full `tests/unit`: **1183 passed**, 4 failures are the pre-existing API-key-leak flakes (pass in isolation), i.e. baseline.
- App boots, **293 OpenAPI paths**.
- No new lint debt: no file flips black clean→dirty; added lines are pyflakes-clean.

## Drift audit

No live silent-drop bugs in the current 42 — every omission is deliberate. The value here is stopping the *next* drift, and feeding the typed-`response_model` work in #467.


mattmorr1:

Issue #602, opened 2026-08-07
Splunk adapter behind the query port, with validation and schema discovery


mattmorr1:
## What to build

A Splunk adapter behind the query port, complete enough that a worker can answer a real question against a real index and have the result land on the ledger as evidence.

The existing Python search path is not safe to put an autonomous loop on top of, and this slice replaces rather than wraps it. Three defects to close, all confirmed by audit — see `services/splunk_service.py` and its `search_by_*` helpers for the current behaviour, and ADR-0004 for the reasoning:

- **No read-only enforcement.** The current path composes a search from caller-supplied text without validating what it contains, so command classes that mutate state, write to storage, notify externally or invoke execution are all reachable, and a query can escape the intended shape entirely. The adapter must validate against a deny-by-default allow-list of the read and aggregation commands hunting actually needs; anything not on that list is refused, including commands the list has never heard of.
- **No argument safety.** The current helpers interpolate caller-supplied values into field expressions without escaping, so a crafted host or user value breaks out of the expression it was meant to sit inside. Values must be escaped or bound, and it must not be possible to alter query structure through a value.
- **No row bounding at the source.** Result size is currently limited by trimming a serialised response after the fact, which yields malformed data the model cannot reason about and cannot tell has been shortened. Bound the result in the query itself.

Schema discovery ships with it — list the available sources, describe one, sample from one. This is required, not optional: embedding an entire schema in a tool description (the approach the DuckDB tool uses) does not survive an index with hundreds of sourcetypes, and field-name confabulation is the failure it prevents.

Deployment should additionally require a read-only role in Splunk where the vendor supports it. The allow-list is the enforceable half of that pair, because it can be verified from code and a customer's role configuration cannot.

## Acceptance criteria

- [ ] Splunk adapter implements the query port and exposes no surface around it
- [ ] Deny-by-default allow-list of permitted read and aggregation commands; anything unrecognised is refused
- [ ] A query cannot escape the shape the adapter intends, verified by test
- [ ] Caller-supplied values cannot alter query structure, verified by test
- [ ] Results bounded in the query itself; a bounded result is always well-formed
- [ ] Schema discovery: list sources, describe one, sample from one
- [ ] A worker answers a question end to end and the finding lands on the ledger with its source system recorded
- [ ] Negative-path suite covering each defect class above, with every case demonstrated refused
- [ ] A command that passes validation but is not on the allow-list is treated as a test failure, not a warning

## Blocked by

- #595

[...]

Issue #593, opened 2026-08-07
Budget through Bifrost, not a local rate table


mattmorr1:
## What to build

Enforce the run budget through Bifrost, which already prices every call, instead of a rate table this layer keeps in step with it.

`core/llm/bifrost/costs.py` calls Bifrost "the authoritative cost source for every LLM call Vigil makes", and every caller in the architecture — daemon, llm-worker, backend — egresses through it. `core/llm/cost/budget.py` already attaches `x-bf-vk` and turns a 402/429 into `BudgetExceeded`. A local `model_rates` table would be a fourth copy of pricing, which is the drift this issue was originally filed about.

The split is: an iteration is not a dollar concept, so the harness counts iterations; dollars are the gateway's to enforce, so the harness asks it. The agent layer's `Budget` reads quota once per iteration and parks before paying for one it has no room for. Within an iteration, Bifrost's own hard stop is the backstop.

`GET /api/governance/virtual-keys/quota` already returns `max_limit`, `current_usage` and `reset_duration` per key, and `get_vk_quota()` already wraps it.

The open part is per-run scoping. Bifrost's budget is per virtual key with a reset period, so a per-run ceiling means minting a VK per run. Only the quota read is wired today; VK create/delete is not.

## Acceptance criteria

- [ ] The agent layer's `Budget` reads spend from Bifrost, not from a local rate table
- [ ] `max_iterations` is enforced by the harness, which is the only thing that knows what an iteration is
- [ ] A run parks as a `checkpoint` before starting an iteration the gateway has no room for
- [ ] An unreadable quota does not refuse: the gateway still caps, so a blip costs the pre-flight check and nothing else
- [ ] A VK is minted per run and torn down with it, so the ceiling is the run's and not the deployment's
- [ ] Spend events journal exact tokens; `cost_usd` is nullable and reconciled from Bifrost
- [ ] No pricing arithmetic in the agent layer, and no rate table it has to keep in step

## Not in scope

Pointing Python's own `compute_call_cost` at Bifrost. `_CACHE_MULTIPLIERS` stays until that has a replacement source; it is a separate issue.

## Blocked by

- #590

mattmorr1:
Superseded by #622 (the wall-clock ceiling). Budget-through-Bifrost is already built in #619 — the harness counts iterations, the gateway reports dollars, and reserve/commit is gone.

Issue #592, opened 2026-08-07
Extract the domain-free harness with its four seams


mattmorr1:
## What to build

Extract the domain-free harness described in ADR-0002 and prove its boundary is real by running a throwaway workflow on it.

The harness owns the tool loop, provider surfaces, tool registry and dispatch, injection scanning and result wrapping, and the four injected seams: memory, budget, state and tool dispatch. It knows about turns, tools, spend and approval. It must contain no vocabulary from any problem domain.

The proof is a deliberately useless workflow with a two-verb vocabulary that does something trivial and terminates. This exists so the harness boundary is first exercised by something that is not threat hunting — if the harness is first driven by the hunt workflow, the boundary gets drawn around hunting and nobody finds out until the second workflow lands.

Recommendation carried from planning, open for confirmation during implementation: ship the memory seam as a port with a null implementation only, and no binding to the existing memory component. A contract shaped by the thing being replaced is the wrong contract.

## Acceptance criteria

- [ ] Harness modules contain no domain vocabulary — enforced by a check, not by review alone
- [ ] All four seams exist as injected ports with at least one implementation each
- [ ] The throwaway workflow runs end to end on the harness and terminates
- [ ] The harness enforces its own turn cap, iteration cap and approval gate independent of any workflow
- [ ] The dollar ceiling is the gateway's: the harness reads quota per iteration and parks, and never prices a call itself
- [ ] Tool results are wrapped and scanned for injection inside the harness, so no workflow can skip it
- [ ] Swapping a seam implementation requires no change to harness or workflow code

## Blocked by

- #590

## Reconciled against the architecture document

Built and matching: the four injected seams, the tier-0 provider/wire split, the tier-1 registry / security / approval set, the harness owning the tool loop, terminal status read from the store, and an approval that parks and returns.

Deferred deliberately, each with an owner, so the boundary moves once rather than drifting:

| Gap | Where it lands |
|---|---|
| `provider.stream` yielding TextDelta / Usage / ToolCall / ToolResult / ApprovalRequired / Done, instead of a blocking `turn()` | #618 |
| Wall-clock in the budget check: the document reads "turns, cost **or wall** exceeded" | with the budget work, #593 |
| `context.py` — frozen prefix, sorted tool schemas, cache breakpoint, transient tail | #601 |
| Second provider surface (`/anthropic`, thinking + `cache_control`) | #601 |

[...]

Issue #591, opened 2026-08-07
Move the ledger to Postgres, gated by fold-equivalence on historical runs


mattmorr1:
## What to build

Replace the file-backed ledger with Postgres, keeping the projection derived rather than stored, and prove the swap changed storage without changing semantics.

The table is append-only with a composite primary key over run and sequence — that key is the single-mutator guarantee, not merely an index. The digest snapshot is stored in its own column and read only by replay, never by the fold: measured on real runs, decision events reach 56.7 KB each and an 8-iteration run is 956 KB against a 5-iteration run's 247 KB, so folding a long run would otherwise read tens of megabytes to build one digest.

The verification is the valuable half. Ten historical run files exist and one of them is a known-good baseline that replays every digest exactly. (Thirteen `.jsonl` files sit in `runs/`, but three are `.inbox.jsonl` sidecars and one is `.corrupt`.) They exist only on the `threat-hunt` branch, so they have to be brought across as fixtures. Load each through both the legacy file reader and the new Postgres path and assert the projections are deep-equal and every digest still reproduces.

Also journal token counts alongside cost: input, output, cache read and cache write. Today only dollars are recorded, which means a run cannot be re-priced and the caching work cannot be measured against a baseline.

## Acceptance criteria

- [ ] Append-only table with a composite primary key over run and sequence
- [ ] Digest snapshots stored separately and excluded from the fold's read path
- [ ] DDL ships as a numbered file alongside the existing database init files; the agent layer issues no DDL
- [ ] All ten historical run files fold to projections deep-equal to the legacy reader's
- [ ] Every digest in those runs still replays exactly; the known-good baseline still reports full reproduction
- [ ] Token counts journaled on every decision and dispatch
- [ ] The equivalence check runs in CI, not just locally

## Blocked by

- #590

mattmorr1:
Superseded by #625, which carries the fold-equivalence gate over the ten historical runs and the fixture move off the `threat-hunt` lineage.

PR #603, opened 2026-08-08
Phase-0: the four interface contracts (#589)

mattmorr1:
Writes down the interfaces every workstream builds against, as types the compiler checks rather than prose. Eleven issues (#591–#602) are blocked on these.

**Four contracts, not five.** The rate table was dropped: Bifrost already prices every call exactly and enforces per-key budgets, and every caller egresses through it, so a local `model_rates` table is a fourth copy of pricing rather than a consolidation of two. Budget-through-Bifrost is #593.

| Contract | Where | Consumed by |
|---|---|---|
| `Tool` port | `services/agent/contracts/tool.ts` | registry and dispatch, the query port and its SIEM adapters, hunt tools |
| Event kinds + `agent_events` | `services/agent/contracts/events.ts`, `infra/database/init/19_agent_ledger.sql` | ledger and resume, hunt, the checkpoint mirror, Python's two permitted reads |
| `Budget` / shared `Spend` | `services/agent/contracts/budget.ts` | the harness seam, any workflow, #593 |
| BullMQ job payload | `services/agent/contracts/job.ts` | resume, deployment, the Python backend |

## Two guarantees the compiler holds

A tool cannot enter the registry having opted out of its bounds: `bounds` is a required parameter of `defineTool`, `defineTool` is the only producer of the branded `RegisteredTool`, and the registry accepts nothing else. Bounds are *verified*, not applied — an adapter exceeding its cap throws, because truncating a serialised result hands the model malformed data it cannot tell has been shortened.

A resume cannot read start-time state: the `resume` arm of the job union carries no `request`, so a resume path that tried to read one would not compile.

Both are pinned by `@ts-expect-error` assertions, so loosening either fails `npm run typecheck` rather than quietly widening the boundary.

## Budget: iterations here, dollars at the gateway

An iteration is not a dollar concept and no gateway knows what one is, so the harness counts them. Dollars are the gateway's to enforce, so `Budget` asks it. `SpendPayload` carries exact tokens and a nullable `cost_usd`, because the thing that bills is the thing that prices.

Reserve/commit is gone. It existed to stop concurrent fan-out overshooting one pool, but a central gateway is a better arbiter of that than a local pool, and estimating dollars for a planned call needs exactly the local pricing table this PR removes.

## Notes

`hunt_events` → **`agent_events`**: it serves hunt, investigate and compose, so naming it for one workflow was wrong, and renaming costs nothing before anything writes to it.

Event kinds are two-level — eight domain-free kinds in the harness, each workflow declaring its own closed set, the repository generic over it. A flat union would make the harness import hunt vocabulary and the domain-free rule uncheckable.

The DDL ships as `19_agent_ledger.sql`; `18_` was taken. It went through all three steps in `infra/database/init/README.md`, including the `values.yaml` list CI does *not* check.

---

**Rebased onto current `main`.** This stack predates the `clients/services/core/infra` reorg (#605), so it carried a 738-file merge (#611) that existed only because the base was stale. That PR is closed and the stack now sits on current main, purely additive.

**Design docs are not in this PR.** The ADRs, glossary and migration plan live outside the repo — working context, not a vigil deliverable.

## Trajectory

`phase0-contracts` is the epic base: everything lands into it, and only this PR faces `main`.

```
#589 contracts (this)  ->  #590 walking skeleton
                             |-- #591 Postgres ledger  -->  #597 resume/lease
                             |-- #592 harness  -->  #618 streaming
                             |                  -->  #595 query port  --> #600 Elastic, #602 Splunk
                             |                  -->  #596 hunt workflow  --> #599 checkpoint from UI
                             |-- #593 budget via Bifrost
                             |-- #594 image, Helm, CI
                             \-- #601 context assembly + cache gate
```

Open PRs: #603 (this), #604 (#590), #610 (#592). #611 and #615 are closed — the first was a stale-base artefact, the second built a rate table that Bifrost makes redundant.


mattmorr1:
Closed in favour of a single bulk PR later. The work is not going anywhere: it lives on `phase0-contracts`, which is 21 commits and 47 files ahead of `main`, purely additive apart from `ci-cd.yml`, `infra/helm/vigil/values.yaml` and `requirements.txt`.

CI was green on this head (`18449b2`): Lint Backend, Unit Tests across Backend, Frontend, Agent Layer and DB-backed, Integration Tests including the walking skeleton, and the security scans.

Before opening the bulk PR, rebase the epic onto main and check the modified-file list:

    git fetch origin && git rebase origin/main phase0-contracts
    git diff --name-status origin/main phase0-contracts | grep -v '^A'

That should show only the three files above. Anything else is something that landed on main while the epic sat, and merging would silently revert it — #616's mempalace bump did exactly that and was caught this way.

PR #609, opened 2026-08-10
fix(integrations): move to httpx, unblock the loop

### LINKED ISSUE #462: Swap the 8 sync requests integrations to httpx

thetosy:
Closes #462.

Moves every integration client off `requests` and onto `httpx`, and closes
the event-loop holes that the swap alone would have left open.

Excludes the 13 `tools/` MCP servers — they run as separate subprocesses, so
they never touch the API event loop.

### The swap — two rules, applied uniformly

Chosen so connection lifecycle is identical:

| Original | Becomes | Why |
|---|---|---|
| `requests.Session()` | `httpx.Client()` | A reused session existed — a client is the faithful translation |
| bare `requests.get/post` | `httpx.get/post` | Both build a throwaway client per call |

I did not introduce a shared client cache for `vstrike_service`:
`get_vstrike_service()` builds a new service per request, so a per-instance
client would leak a pool per API call, and a module-level cache would change
TLS-session reuse and failure modes. Possible follow-up.

### The blocking fixes

**`backend/api/vstrike.py` — 23 handlers were on the event loop.**

The 23 outbound-proxy handlers — `health_check`, `get_asset_topology`,
`list_adjacent_assets`, `get_blast_radius` and 19 more — were still
`async def` calling the synchronous, `requests`-backed `VStrikeService`
directly on the loop.

**Six more sites** where an `async def` called a sync client directly, all
fixed with `asyncio.to_thread` — the pattern `daemon/sandbox_poller.py`
already used:

| File | Site |
|---|---|
| `daemon/poller.py` | `_poll_splunk`, `_poll_crowdstrike` |
| `daemon/federation/adapters/{splunk,crowdstrike}.py` | `fetch()` |
| `services/microsoft_defender_ingestion.py` | `fetch_alerts` (async by interface; fixing it covers both callers) |
| `services/autonomous_response_service.py` | Slack + PagerDuty escalation |

`SplunkService.search`: polls its job with `time.sleep(1)` × 60, so a slow Splunk froze the **entire daemon loop**, not
just that poll.

### Migration details that aren't cosmetic

| | |
|---|---|
| **Redirects** | `requests` followed them by default, httpx does not. `follow_redirects=True` is passed at every call site — this is silent if you get it wrong, and it matters for Jira Cloud and Microsoft's login endpoints. |
| **`httpx.InvalidURL` is not an `httpx.HTTPError`** | requests folded both into `RequestException`. A bare `except httpx.HTTPError` would let a malformed base URL escape and 500 instead of returning `None`. Every handler catches both. |
| **`verify` is constructor-only on `httpx.Client`** | `splunk_service.py:35` did `self.session.verify = verify_ssl` — a silent no-op on httpx that would have broken self-signed deployments. Moved into the constructor. |
| **SOCKS5** | `requirements.txt` documents SOCKS proxying via `HTTPS_PROXY`/`ALL_PROXY` backed by `pysocks` — which **httpx cannot use**. Added `socksio`; kept `pysocks` for the `requests` callers left in `scripts/`/`tools/`. |
| **httpx was never pinned** | Only `respx` was. It arrived transitively via `anthropic`/`openai`. Now explicit. |
| **`urllib3.disable_warnings`** | Dropped from `splunk_service` — httpx doesn't use urllib3 and emits no `InsecureRequestWarning`. |

### On "timeouts and retry behavior preserved"

- **Retries:** nothing to preserve — confirmed no file uses `HTTPAdapter` or
  `urllib3.Retry`. VStrike's JWT re-auth-and-retry is application-level and
  is untouched.
- **Timeouts:** preserved where they existed. But **`SplunkService` passed
  no timeout on any of its 6 requests**, and neither did either Defender
  call — with requests' `None` default that's an unbounded hang against an
  unresponsive vendor. Those were **added**, not preserved. Flagging it as a
  deliberate behaviour change rather than pretending it's a no-op.

## Tests

+31 new tests. Both suites run on `main` and this branch show identical
failure counts (42 unit / 15 integration — all environmental: missing
`openai`, `anthropic`, `sentry_sdk`, uninitialised DB).

- `test_splunk_service.py` (new, 13) and `test_crowdstrike_service.py`
  (new, 10) — first HTTP-level coverage either module has had
- `test_blocking_offload.py` (new, 4) — runs each async path against a
  10ms ticker and asserts the loop kept serving
- `test_vstrike_service.py` +4 respx; existing 61 retargeted to `httpx`
- `test_vstrike_ui_routes.py` — 37 `asyncio.run(...)` wrappers unwrapped

Both new guard families fail against the unfixed code: reverting
`follow_redirects` fails the redirect tests, stashing the `to_thread`
changes fails all 4 offload tests.

## Commits

| SHA | |
|---|---|
| `d4cb308` | `chore(deps): pin httpx and add socksio for SOCKS5 parity` |
| `17c16fb` | `refactor(vstrike): move the outbound client to httpx` |
| `05a4ea0` | `fix(api): run the VStrike proxy handlers off the event loop` |
| `cd3dc47` | `refactor(splunk): move to httpx and bound every request` |
| `6e4b44c` | `refactor(edr,response): move the remaining clients to httpx` |
| `0d522e5` | `fix(daemon,services): keep integration calls off the event loop` |
| `1a37ccd` | `refactor(jira): move the export client to httpx` |
| `b4e939c` | `refactor(daemon): move the enrichment and sandbox calls to httpx` |
| `29aa145` | `docs: tighten the httpx migration comments` |

### DIFF
--- a/services/api/routers/jira_export.py
+++ b/services/api/routers/jira_export.py
@@ -18,9 +18,8 @@ from core.routing import Auth, RouterMeta, UnitOfWorkSession
 
 logger = logging.getLogger(__name__)
 
-# httpx defaults to follow_redirects=False; requests followed redirects.
-# Jira Cloud does redirect (site moves, context-path changes), so this is
-# passed on every call to keep the swap behaviour-preserving.
+# requests followed redirects by default, httpx does not — and Jira Cloud
+# redirects on site moves and context-path changes.
 _FOLLOW_REDIRECTS = True
 
 router = APIRouter()
--- a/services/api/routers/jira_export.py
+++ b/services/api/routers/jira_export.py
@@ -13,10 +13,15 @@ from services.api.middleware.auth import get_current_user
 from core.auth.auth_service import AuthService
 from core.storage.models import User, Case, Finding
 from core.config import get_integration_config
-import requests
+import httpx
 from core.routing import Auth, RouterMeta, UnitOfWorkSession
 
 logger = logging.getLogger(__name__)
+
+# httpx defaults to follow_redirects=False; requests followed redirects.
+# Jira Cloud does redirect (site moves, context-path changes), so this is
+# passed on every call to keep the swap behaviour-preserving.
+_FOLLOW_REDIRECTS = True
 
 router = APIRouter()
 
@@ -152,12 +157,13 @@ def export_case_to_jira(
             }
         }
         
-        response = requests.post(
+        response = httpx.post(
             f"{url}/rest/api/2/issue",
             auth=auth,
             headers=headers,
             json=issue_data,
-            timeout=30
+            timeout=30,
+            follow_redirects=_FOLLOW_REDIRECTS,
         )
         response.raise_for_status()
         result_data = response.json()
@@ -178,12 +184,13 @@ def export_case_to_jira(
                 }
                 
                 try:
-                    sub_response = requests.post(
+                    sub_response = httpx.post(
                         f"{url}/rest/api/2/issue",
                         auth=auth,
                         headers=headers,
                         json=subtask_data,
-                        timeout=30
+                        timeout=30,
+                        follow_redirects=_FOLLOW_REDIRECTS,
                     )
                     sub_response.raise_for_status()
                     subtasks_created += 1
@@ -202,7 +209,7 @@ def export_case_to_jira(
     
     except HTTPException:
         raise
-    except requests.exceptions.RequestException as e:
+    except (httpx.HTTPError, httpx.InvalidURL) as e:
         logger.error(f"JIRA API error: {e}")
         return JiraExportResponse(
             success=False,
@@ -275,11 +282,12 @@ def export_remediation_to_jira(
         auth = (email, token)
         headers = {"Content-Type": "application/json"}
         
-        parent_response = requests.get(
+        parent_response = httpx.get(
             f"{url}/rest/api/2/issue/{request.parent_issue_key}",
             auth=auth,
             headers=headers,
-            timeout=30
+            timeout=30,
+            follow_redirects=_FOLLOW_REDIRECTS,
         )
         parent_response.raise_for_status()
         parent_data = parent_response.json()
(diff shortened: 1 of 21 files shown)


services/api/routers/jira_export.py
```diff
@@ -300,23 +307,25 @@ def export_remediation_to_jira(
...
+                user_response = httpx.get(
                     f"{url}/rest/api/2/user/search",
                     auth=auth,
                     headers=headers,
                     params={"username": request.assign_to},
-                    timeout=30
+                    timeout=30,
+                    follow_redirects=_FOLLOW_REDIRECTS,
                 )
                 if user_response.ok and user_response.json():
```
nestor-deeptempo: `httpx.Response` has no `.ok` — `requests` had it, httpx exposes `is_success` / `is_error` / `is_client_error` only.

The `AttributeError` lands outside the inner `try` at 321, so it unwinds to the `except Exception` at 352: every remediation export with `assign_to` set returns `success=False` after already creating subtasks in Jira.

```suggestion
                if user_response.is_success and user_response.json():

nestor-deeptempo (changes requested):

services/api/routers/jira_export.py
thetosy: Updated

nestor-deeptempo (approved):
lg 👍

Issue #601, opened 2026-08-07
Byte-stable context assembly, token journaling and a cache-hit gate


mattmorr1:
## What to build

The context layout from the architecture document, as the only way a request is assembled, plus the token journaling and cache measurement that prove it works.

The document's layout, top to bottom:

| Band | Rule |
|---|---|
| system prompt | frozen, byte-identical |
| tool schemas | sorted keys, stable order |
| memory recall | rendered once, then persisted |
| — cache breakpoint — | |
| conversation history | grows each turn; middle folds, edges hold |
| transient tail | working state + turn suffix, re-rendered, never persisted |

This is the largest cost lever available: the worker system prompt plus the full schema dump is byte-identical across every tool turn and every worker, and is currently re-billed at full rate each time. It is built as the single assembly path because retrofitting byte-stability means auditing every call site.

The second provider surface belongs here too. Only the OpenAI `/v1` wire exists; `cache_control` and thinking need `/anthropic`, and that is what makes the breakpoint above do anything.

## Acceptance criteria

- [ ] One assembly path; no caller builds a request any other way, enforced by a check
- [ ] The system prompt is byte-identical across turns, asserted rather than assumed
- [ ] Tool schemas serialise with sorted keys in a stable order
- [ ] An explicit cache breakpoint sits below memory recall and above conversation history
- [ ] The transient tail is re-rendered every turn and never persisted
- [ ] Conversation history folds from the middle; the first and last turns survive
- [ ] The `/anthropic` surface exists and carries `cache_control` and thinking
- [ ] Token counts journaled per call: input, output, cache read, cache write
- [ ] A cache-hit-rate assertion runs in CI and fails when the frozen prefix stops being frozen

## Blocked by

- #592

mattmorr1:
Superseded by #622.

Issue #597, opened 2026-08-07
Resume: per-run lease, watchdog, and reaping in-flight work


mattmorr1:
## What to build

Make a run survive the process that was running it. Kill a worker mid-iteration and have another pick the run up from the ledger and continue.

Three mechanisms: a per-run lease so exactly one worker advances a run at a time, a watchdog that reclaims a lease whose holder died, and reaping of the work that was in flight when the process went down — an interrupted dispatch is marked failed and recorded as a gap, and the question it was answering returns to the frontier rather than being silently lost.

The lease is for liveness, not correctness: the ledger's composite primary key already prevents two writers from interleaving. The lease exists so a crashed holder does not block a run forever.

The property to demonstrate is that the ledger is literally the resume point — a crash costs at most one in-flight iteration, and nothing is reconstructed from memory or from a file on the dead worker's disk.

Also close the hole ADR-0006 leaves open and assigns here: the harness loop returns its events rather than appending them, so a workflow can discard them. The in-memory pool still holds, so the cost cap survives, but the ledger's spend fold would then disagree with what was actually spent. This issue owns it because re-entrancy is designed here.

## Acceptance criteria

- [ ] Exactly one worker advances a given run at a time
- [ ] A lease whose holder has died is reclaimed after its expiry, not before
- [ ] An interrupted dispatch is marked failed and recorded as a gap
- [ ] The question an interrupted dispatch was answering returns to the frontier
- [ ] A run killed mid-iteration resumes and completes, with the interruption visible on the ledger
- [ ] Resume reads nothing but the ledger and the job payload
- [ ] Test kills a worker mid-iteration and asserts the run completes
- [ ] A workflow cannot discard the harness's events; the ledger's spend fold agrees with the pool

## Blocked by

- #591

mattmorr1:
Superseded by #633, which carries the open hole: the loop returns events rather than appending them, so a workflow can discard them and the ledger's spend fold would disagree with what was spent.

Issue #590, opened 2026-08-07
Walking skeleton: API to queue to TypeScript worker to ledger


mattmorr1:
## What to build

The thinnest possible run that touches every layer, end to end: the API accepts a request to start a run, it lands on a queue, a TypeScript worker consumes it, appends an event to Postgres, marks the run terminal, and the API reports the outcome.

No model call, no tools, no decision vocabulary. The point is to prove the seam between the Python backend and the TypeScript agent layer works in a deployed shape — queue, connection, schema, ownership, error propagation — before any real logic depends on it.

Per ADR-0001, the agent layer owns its own tables and is their single writer. The backend enqueues plain JSON; it does not write the ledger.

## Acceptance criteria

- [ ] An API endpoint starts a run and returns its identifier
- [ ] The job is consumed by a TypeScript worker process, not by Python
- [ ] The worker appends at least one event and marks the run terminal
- [ ] The API can report the run's status by reading state the worker persisted
- [ ] A worker crash mid-job leaves the run resumable rather than lost or half-written
- [ ] Concurrent writes to the same run are rejected by the composite primary key, with a test that demonstrates it
- [ ] Integration test covering the full path runs in CI

## Blocked by

- #589

Issue #589, opened 2026-08-07
Fix the four Phase-0 interface contracts


mattmorr1:
## What to build

Write down the five interface contracts every other slice depends on, and agree them. This produces committed contract definitions (types plus a short rationale per contract), not a working feature — it is the one place in this migration where a wrong call is expensive to unwind, because five workstreams build against it in parallel.

The five:

1. **`Tool` port** — the signature every tool implements regardless of what backs it (native adapter, MCP surface, in-process). Error shape, and where row/timeout bounds are declared and enforced.
2. **Event kinds and the `hunt_events` payload discriminant** — the closed set of event kinds the ledger accepts, and which workflow each belongs to, so one table serves hunt, investigate and compose.
3. **`Budget` / shared `Spend`** — how a lead, its workers and a critic draw on one pool, what a refusal looks like to the caller, and where the refusal is journaled.
4. **BullMQ job payload** — what the API enqueues to start or resume a run, and what the worker needs to resume from nothing but the payload plus the ledger.
5. **Rate table schema** — the shape TypeScript reads at startup and the budget gate prices against.

Context lives in `docs/migration-plan.md` (Phase 0) and ADRs 0001-0004. The glossary in `CONTEXT.md` is the vocabulary these contracts should use.

## Acceptance criteria

- [ ] All four contracts committed as types (plus the ledger DDL) with a short rationale each
- [ ] Each contract names which workstreams consume it, so a later change knows its blast radius
- [ ] The `Tool` port makes it impossible for a caller to opt out of row and timeout bounds
- [ ] The event-kind set is closed and versioned, with the rule for adding one stated in `events.ts` itself
- [ ] The BullMQ payload is sufficient to resume a run with no other in-memory state
- [ ] Reviewed and agreed by everyone who will build against them before any dependent slice starts

## Blocked by

None - can start immediately

## Amended: four contracts, not five

Contract 5 was a Postgres rate table read by both languages. It is removed. Bifrost already prices every call exactly and enforces per-key budgets, and every caller egresses through it, so a local table is a fourth copy of pricing rather than a consolidation of two. Budget-through-Bifrost is #593.

Contract 3 (`Budget` / shared `Spend`) survives with its shape changed: reserve/commit needed local pricing to estimate dollars, so it is gone. The harness counts iterations, which is the one thing no gateway can know, and reads spend from the gateway. `SpendPayload` carries exact tokens and a nullable `cost_usd`, because the gateway is what bills.

[...]

PR #610, opened 2026-08-10
feat(core): a domain-free harness and a throwaway workflow that proves it

mattmorr1:
Extracts ADR-0002's domain-free harness into `services/agent/core/` and proves the boundary
by running a deliberately useless workflow on it. `Refs #592` rather than
`Closes`: the ticket's own point is that this line moves once, and ADR-0006 now
says which parts.

**Stacked.** Base is `phase0-walking-skeleton` (#604), which sits on
`phase0-contracts` (#603). Review those first; this is the third in the stack.

## The acceptance criteria

| Criterion | Where |
|---|---|
| Harness contains no domain vocabulary, enforced by a check | `services/agent/tests/core/domain-free.test.ts` |
| All four seams as injected ports, one implementation each | `services/agent/core/seams.ts`, `{memory,state,dispatch}.ts`, `budget.ts` |
| The throwaway workflow runs end to end and terminates | `services/agent/workflows/tally/`, `services/agent/tests/core/tally.test.ts` |
| Turn cap, budget and approval gate enforced independently | `services/agent/core/loop.ts`, `services/agent/tests/core/loop.test.ts` |
| Tool results wrapped and scanned inside the harness | `services/agent/core/security.ts`, called only from `loop.ts` |
| Swapping a seam needs no harness or workflow change | `services/agent/tests/core/tally.test.ts`, "swapping a seam" |

`State` is deliberately the public surface of #590's `LedgerRepository`, so the
Postgres implementation satisfies the port with no adapter and no edit to that
file. A type-level assertion in `seams.test.ts` makes loosening either side fail
`typecheck`.

Memory ships as a port with `nullMemory` as its only implementation, confirming
the recommendation the issue carried forward. No MemPalace binding.

## Two of #589's contracts had to grow

Both are one-line additions to files on #603, flagged here because that PR is
still open and its reviewers should see them.

- **`Budget.beginIteration(): Refusal | null`.** Nothing in the contract advanced
  `Spend.iterations`, so `max_iterations` was unenforceable. It is separate from
  `reserve` because a tool loop is many model calls inside one iteration —
  counting calls would exhaust the limit immediately on a wide loop.
- **`RUN_KINDS` gains `tally`**, the conformance workflow's kind. Not a product
  surface. Inside the closed set rather than beside it, so the throwaway
  exercises the same typed path a real workflow does.

## Calls worth a reviewer's attention

Argued in full in the harness-boundary ADR, held outside the repo.

- **`Provider.turn` is one model call; the loop is the harness's.** The prototype
  runs the tool loop inside `llm.ts`, which would put the turn cap, the budget
  gate and the injection scan on the far side of a swappable seam. With the loop
  in `core/loop.ts`, it imports no vendor SDK and a scripted provider is a
  drop-in — which is what lets the conformance workflow run the real harness in
  CI with nothing faked but the model.
- **`costOf` did not come across.** ADR-0005 names it as the reason a run cannot
  be re-priced, so the surface reports tokens and the budget prices them.
- **Approval is a set of tool ids in the turn config**, not a field on
  `RegisteredTool`: #589's tool contract is settled, and the same tool needs
  approval in one deployment and not another. The checkpoint id is derived from
  the run, tool and arguments, so a resumed run recognises the resolution that
  answered that exact call and an approval for other arguments does not admit a
  different one.
- **The turn cap stops the tool loop, not the run.** A role that hit it answers
  over what it gathered and reports `capped`; failing would discard work already
  paid for. This differs from what I first planned, and the reason is in ADR-0006.
- **One provider surface.** The gateway routes to either provider family behind a
  model name, so the `/anthropic` wire carries nothing until #601 needs
  `cache_control` and thinking. The token reader already handles both usage
  shapes, because that difference is reported today.

## One open hole, named

The loop returns its own events rather than appending them, so a workflow commits
them alongside its own and one iteration stays one transaction. **Nothing stops a
workflow from discarding them.** The in-memory pool still holds, so the cost cap
survives, but the ledger's spend fold would disagree with what was spent. #597
owns closing this when it adds the per-run lease, since that is where re-entrancy
is designed. A branded opaque journal in the style of `defineTool` would close it
by construction; ADR-0006 records why that was not built now rather than leaving
it to be rediscovered.

One assumption in the pricing arithmetic is also unmeasured: whether the
gateway's normalised `prompt_tokens` includes a cache write. It is one expression
with a comment on it, so #593 correcting it is one edit.

## Verification

No CI change — #604's `test-agent-layer` job already runs `services/agent/`'s typecheck and
tests with Postgres and Redis, and every test added here is pure.

```
cd ai && npm ci && npx tsc --noEmit && npx vitest run
```

- `tsc --noEmit` clean.
- 94 passing, up from 11 on the base.
- The 10 failures are `tests/ledger.test.ts` and `tests/worker.test.ts`, which
  need `DATABASE_URL`. **Identical to the base branch** — same two files, same ten
  tests, confirmed by running the base. They pass in CI, which has the services.

Both checks that could pass vacuously were confirmed to bite:

- Planting a sentence about hypotheses and telemetry in `core/loop.ts`, and
  importing the tally vocabulary, fails both `domain-free.test.ts` assertions.
- Mangling the remote dispatch's arguments fails the seam-swap ledger equality
  rather than leaving it green.

## Not in scope

Main's Python `core/agents/`, `core/chat/` and the rest of the 16,852-line AI
surface are what this eventually replaces, but they still serve the live backend.
Deleting them belongs to WS-D and later. Context assembly and byte-stable
caching are #601's.

---

**Rebased onto current `main` (force-pushed).** This stack was cut before the
`clients/services/core/infra` reorg (#605) landed, so it carried
`chore/absorb-main-reorg` (#611) — a 738-file merge that existed only because
the base was stale. That PR is closed and the whole stack now sits on current
main, purely additive at 75 files. Paths in this description are the
post-reorg ones.

---

**Design docs are not in this PR.** The ADRs, glossary and migration plan live
outside the repo — they are working context, not a vigil deliverable. Where this
description cites one, it is naming the reasoning, not a file you will find in
the diff.

## Reconciled against the architecture document

The harness composition matches: four injected seams, the tier-0 provider/wire split, tier-1 registry / security / approval, the loop owned here, terminal status read from the store, approval that parks and returns.

**The budget seam changed.** Reserve/commit and local pricing are gone. The document's loop does `budget.add` on streamed usage, and every caller egresses through Bifrost, which prices exactly and enforces per-key. So the harness counts iterations — the one thing no gateway knows — and reads spend from the gateway. An unreadable quota is not a refusal, because the gateway still caps on its own side.

**Deferred, each with an owner:** streaming `provider.stream` (#618), wall-clock in the budget check (#593), `context.py`'s frozen prefix and cache breakpoint plus the second `/anthropic` surface (#601).

Also carries `scripts/check_comment_style.py`, which gates the agent layer at two lines per comment block and no docstrings, running beside `lint-imports`.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -62,6 +62,10 @@ jobs:
       - name: Run import-linter
         run: lint-imports
 
+      # The agent layer's comment style. Gates, like import-linter above.
+      - name: Check comment style
+        run: python scripts/check_comment_style.py
+
   lint-frontend:
     name: Lint Frontend
     if: github.repository == 'Vigil-SOC/vigil'
--- a/scripts/check_comment_style.py
+++ b/scripts/check_comment_style.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+# Gates the agent layer's comment style: no comment block over two lines, and no
+# docstrings. Prose that needs more room does not belong in the code at all.
+
+from __future__ import annotations
+
+import ast
+import re
+import sys
+from pathlib import Path
+
+MAX_LINES = 2
+ROOT = Path(__file__).resolve().parents[1]
+
+# Only what the agent-loop work owns. Widen deliberately, never by accident:
+# core/agents and core/llm predate this and carry docstrings throughout.
+SCOPED = (
+    "services/agent",
+    "core/agents/queue.py",
+    "core/agents/agent_runs_router.py",
+    "core/llm/cost/rates.py",
+    "infra/database/init/19_agent_ledger.sql",
+    "infra/database/init/20_model_rates.sql",
+    "infra/database/init/21_model_rates_seed.sql",
+)
+SKIP = ("node_modules", "package-lock.json", "dist", "__pycache__")
+
+LINE_COMMENT = re.compile(r"^\s*(//|#|--)")
+SHEBANG = re.compile(r"^#!")
+
+
+def scoped_files() -> list[Path]:
+    found: list[Path] = []
+    for entry in SCOPED:
+        target = ROOT / entry
+        if target.is_file():
+            found.append(target)
+        elif target.is_dir():
+            found.extend(p for p in target.rglob("*") if p.is_file())
+    return [p for p in found if p.suffix in {".ts", ".py", ".sql"} and not any(s in p.parts or s == p.name for s in SKIP)]
+
+
+def long_blocks(path: Path) -> list[tuple[int, int]]:
+    runs: list[tuple[int, int]] = []
+    start = count = 0
+    for number, line in enumerate(path.read_text(encoding="utf8").splitlines(), 1):
+        if LINE_COMMENT.match(line) and not SHEBANG.match(line):
+            if not count:
+                start = number
+            count += 1
+            continue
+        if count > MAX_LINES:
+            runs.append((start, count))
+        count = 0
+    if count > MAX_LINES:
+        runs.append((start, count))
+    return runs
+
+
+def docstrings(path: Path) -> list[int]:
+    try:
+        tree = ast.parse(path.read_text(encoding="utf8"))
+    except SyntaxError:
+        return []
+    nodes = (ast.Module, ast.ClassDef, ast.FunctionDef, ast.AsyncFunctionDef)
+    return [node.body[0].lineno for node in ast.walk(tree) if isinstance(node, nodes) and ast.get_docstring(node)]
+
+
+def main() -> int:
+    failures: list[str] = []
+    for path in sorted(scoped_files()):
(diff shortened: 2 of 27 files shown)


mattmorr1:
Merged into the epic branch. `phase0-contracts` was fast-forwarded to this branch's tip, so every commit here is on the epic with its history intact — nothing was squashed.

GitHub's stacked-PR guard would not let this be retargeted to `phase0-contracts`, hence closing rather than merging. `git merge-base --is-ancestor` confirms the head is fully contained.

CI passed on the epic branch at #603: Lint Backend, Unit Tests (Backend, Frontend, Agent Layer, DB-backed), Integration Tests, Security Scans.

PR #611, opened 2026-08-11
chore: absorb main's reorg into the agent-layer stack

mattmorr1:
Prerequisite for #593, and for #603/#604/#610 merging at all.

main landed **#605** (`epic #481: clients/services/core/infra, gated by
import-linter`) after this stack branched. The whole Python tree moved:
`services/*.py` into `core/` domain packages, `database/` and `helm/` under
`infra/`, `backend/api/` into `services/api/routers/`. Every Python file #593
must touch — `core/llm/providers/registry.py`, `core/llm/cost/calls.py`,
`tests/unit/llm/test_model_registry_pricing.py` — exists only on that side of the
merge. Absorbing it here keeps #593's diff about rates instead of about a reorg.

Based on `feat/592-domain-free-harness` (#610) and targeting it, so the stack
stays reviewable one PR at a time: #603 → #604 → #610 → this → #593.

## What git got right

It relocated the stack's new files into the reorganised tree and asked only for
confirmation on five: the two DDL files, their two chart-bundle copies, and the
agent-runs router. All five were already in the correct new location.

## Three things it could not see

**The agent-layer CI job pointed at a directory that no longer exists.** Its
`psql` step applied DDL from `database/init/`, now `infra/database/init/`. The
job would have failed on a missing path. The same stale prefix appears in
ADR-0005 and the migration plan, repointed here too.

**`CONTEXT.md` is two unrelated documents sharing a filename.** main's is the
`core/` domain glossary the reorg introduced; this stack's is the AI-loop
glossary that `docs/migration-plan.md` and the #589 contract comments cite.
Neither can be dropped, so the AI-loop one becomes **`ai/CONTEXT.md`**, beside
the package it describes, and all seven references are repointed
(`docs/migration-plan.md`, ADR-0005, ADR-0006, `ai/contracts/tool.ts`,
`ai/core/loop.ts`, `ai/core/security.ts`).

**#590's API router was left importing deleted modules.** `agent_runs.py` landed
in the right directory but still imported `api._meta` and `database.connection`,
and took a `Session` through a `get_db` the reorg removed. It now uses
`core.routing`'s `Auth`, `RouterMeta` and `UnitOfWorkSession` like every sibling
router, and drops a `sys.path.insert` hack that is no longer needed. Discovery is
by `ROUTER_META` (`services/api/discovery.py`), so nothing registers it by hand.

`services/agent_queue.py` needed no change — it imports only `core.config`, which
the reorg kept, and `services` → `services` breaks no contract.

## Verification

Run on the merged tree:

```
lint-imports                                     # 3 contracts kept, 0 broken
pytest tests/unit/llm/test_model_registry_pricing.py   # 14 passed
cd ai && npx tsc --noEmit && npx vitest run tests/core # clean, 83 passed
```

Not run: the full Python suite. It needs the complete `requirements.txt` plus
Postgres and Redis, which is what CI is for. `services/api/routers/agent_runs.py`
was checked by `py_compile` and by confirming every name it now imports exists in
the module it imports from.

## Noted, not fixed

`tests/pytest.ini` still lists `--cov=backend --cov=database --cov=daemon` for
directories the reorg deleted. Pre-existing on main and unrelated to this stack,
so it is not touched here.


mattmorr1:
Superseded. The agent-loop stack has been rebased onto current main, so absorbing the reorg as a merge commit is no longer needed — the whole stack is now 75 files and purely additive against main, and this PR's 738-file diff was entirely an artefact of a stale base.

Replaced by the rebased stack: #603 → #604 → #610 → #612.

PR #615, opened 2026-08-11
feat(rates): one rate table, read by both languages, gating the budget

mattmorr1:
One source of truth for model rates, read by both languages, gating the budget.
Base is `feat/592-domain-free-harness` (#610), so the stack is #603 → #604 →
#610 → this. Replaces #612, which GitHub closed when its old base branch
`chore/absorb-main-reorg` was deleted and would not let us retarget.

## The acceptance criteria

| Criterion | Where |
|---|---|
| Rate table shipped as DDL alongside the init files | `20_model_rates.sql`, already shipped by #589 |
| Seeded from the Python registry's current rates | `infra/database/init/21_model_rates_seed.sql` |
| The agent layer reads rates from the table | `services/agent/rates/repository.ts` |
| The budget prices in-loop, no per-call round-trip | `services/agent/core/budget.ts`, read once and frozen |
| Redundant cost arithmetic deleted | `_CACHE_MULTIPLIERS` and `get_cache_multipliers` gone |
| A test fails if a model is missing or differently priced in the other place | `tests/fixtures/pricing_parity.json`, read by both suites |

## Three corrections to the issue text

**1. "With its pricing tests still passing" is not achievable, and shouldn't be.**
Four of the fourteen tests in `test_model_registry_pricing.py` assert the
cache-multiplier derivation *by name*. The DDL #589 already shipped decided cache
rates are **stored**, not derived. Those tests assert the mechanism being deleted,
so preserving them would have meant keeping the multipliers — the opposite of
criterion 5. They are rewritten; their intent survives.

**2. The formula was drifting too, and the criterion as written would not have
caught it.** Python billed `input_tokens * in_rate + cache_read * cache_rate` —
input *excluding* the cached share. `services/agent/core/budget.ts` deducted `cache_read`
*from* input — including it. Same tokens, same rates, two different dollar
figures. A rate-parity test passes straight through that, so this PR fixes the
formula and the parity fixture compares **dollars**.

**3. It also closes the open question from #610.** That PR flagged an unmeasured
assumption about whether a cache write sits inside the gateway's normalised
`prompt_tokens`. Python's `_CACHE_MULTIPLIERS` answers it: OpenAI's creation
multiplier is `0.0`, "no premium — just regular input tokens". So it is inside,
and the assumption is now stated rather than guessed.

## The formula, once

`TokenCounts.input` is defined as the **total** input, cached share included.
That definition is what lets one formula serve both gateway surfaces, which
disagree natively. Each reader normalises to it: OpenAI's `prompt_tokens` already
counts the cached share so it passes through; the Anthropic route excludes both
cache counters so they are added back. Before this, an Anthropic-shaped usage
report had its cache read subtracted from an input that never contained it.

Stated once per language — `price_tokens` in `core/llm/cost/rates.py`, `priceOf`
in `services/agent/core/budget.ts` — each naming the other. `compute_call_cost` keeps its
Anthropic-native arguments and converts at its own boundary, so its dollars do
not change and the existing cost tests hold.

## Decisions worth a reviewer's attention

- **`model_id` is the gateway's prefixed `provider/model`**, so the string the
  agent layer configures and the one Bifrost reports cost against are one key.
  Consequence, stated because it is a real cost: `infer_provider_type()` exists
  because `LLMInteractionLog` stores a bare model name, so pricing an analytics
  row is now infer-the-provider *then* reconstruct-the-prefix. That
  reconstruction is `rate_key()` and lives in exactly one place.
- **`rateTableOf` keys on `model_id` alone**, since composing provider and model
  around an already-prefixed id produced `openai/openai/gpt-4o`. The provider
  argument stays load-bearing: a caller that disagrees with itself about which
  provider a model belongs to gets a miss, and therefore a refusal, not a price.
  A row whose id does not carry its own provider throws at load.
- **The wildcard row** prices self-hosted Ollama models, which cannot be
  enumerated for the same reason the tier heuristic cannot. Honoured **only** at
  `pricing_source = 'zero'`, so one row can never make a provider's priced models
  look free.
- **The heuristic stays in Python and never gates.** A regex cannot be a row. A
  model the table misses gets a tier estimate for badging, and the agent layer's
  budget refuses it outright — `unpriced_model` fails closed.
- **The two languages degrade in opposite directions, deliberately.** An
  unreachable table leaves Python estimating; it stops the agent layer from
  starting. A wrong rate in a budget does not mis-report, it removes the cap.
- **The seed was generated, not typed.** This issue exists because someone typed
  a rate 3× too high. The generator read `_CATALOG` and `_CACHE_MULTIPLIERS` and
  is not committed, because the next commit deletes the fields it read; the SQL
  header records the provenance. Both test suites parse the committed SQL rather
  than restating it, so a rate mistyped there fails a price assertion.

## Not in scope, deliberately

`ModelInfo.input_cost_per_1k` keeps its unit. It is an API field reaching
`services/api/routers/ai_config.py` and `clients/web`'s TypeScript types, so
renaming it is a frontend change, not a rate change. One source with one
conversion at the presentation boundary is not the drift #593 was filed for.

**A rate change ships as the next numbered file.** The Helm `db-init` Job records
applied filenames in `_vigil_schema_versions` and skips a file it has already
run, so editing `21_` would never reach an existing deployment. The seed header
says so. A rate-sync mechanism is not invented here.

## Verification

Run against real Postgres 16 and Redis 7:

```
psql -f infra/database/init/{19,20,21}_*.sql       # applies clean, idempotent
cd ai && npx tsc --noEmit && npx vitest run        # 120 passed, 13 files
pytest tests/unit/llm                              # 292 passed, 62 pre-existing failures
lint-imports                                       # 3 contracts kept
```

- **The whole agent-layer suite passes**, ledger and worker Postgres/Redis suites
  included. Those were deferred to CI in #610 and #611; they are verified now.
- **Python has zero new failures.** 62 before and 62 after, byte-identical, all
  environmental (missing optional deps, no DB). 292 passing against 281.
- **The parity fixture was confirmed to bite, both ways.** Reverting the
  TypeScript deduction fails two cases; reverting the Python clamp fails five.
- The seed applies twice with 13 rows both times, and `pricing_source` rejects a
  value outside its closed set.

All three steps in `infra/database/init/README.md` are done: the chart bundle copy
CI diffs, and the `values.yaml` list it does not. `helm` is not installed locally,
so step 3's render is left to the helm-chart workflow; the Job template ranges
over `dbInit.sqlFiles` unconditionally, and the entry is there.

---

**Rebased onto current `main` (force-pushed).** This stack was cut before the
`clients/services/core/infra` reorg (#605) landed, so it carried
`chore/absorb-main-reorg` (#611) — a 738-file merge that existed only because
the base was stale. That PR is closed and the whole stack now sits on current
main, purely additive at 75 files. Paths in this description are the
post-reorg ones.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -62,6 +62,10 @@ jobs:
       - name: Run import-linter
         run: lint-imports
 
+      # The agent layer's comment style. Gates, like import-linter above.
+      - name: Check comment style
+        run: python scripts/check_comment_style.py
+
   lint-frontend:
     name: Lint Frontend
     if: github.repository == 'Vigil-SOC/vigil'
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -220,9 +220,11 @@ jobs:
       - name: Apply agent-layer schema
         run: |
           PGPASSWORD=test psql -h localhost -p 5432 -U test -d agent_test \
-            -f database/init/19_agent_ledger.sql
+            -f infra/database/init/19_agent_ledger.sql
           PGPASSWORD=test psql -h localhost -p 5432 -U test -d agent_test \
-            -f database/init/20_model_rates.sql
+            -f infra/database/init/20_model_rates.sql
+          PGPASSWORD=test psql -h localhost -p 5432 -U test -d agent_test \
+            -f infra/database/init/21_model_rates_seed.sql
 
       - name: Typecheck
         working-directory: services/agent
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -170,6 +170,68 @@ jobs:
         working-directory: clients/web
         run: npm run test:run
 
+  # The agent layer is a second Node package with its own manifest, so it needs
+  # its own job -- the frontend job above is scoped to frontend/. Node 20 rather
+  # than NODE_VERSION because services/agent/package.json requires it.
+  test-agent-layer:
+    name: Unit Tests - Agent Layer
+    if: github.repository == 'Vigil-SOC/vigil'
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: pgvector/pgvector:pg16
+        env:
+          POSTGRES_DB: agent_test
+          POSTGRES_USER: test
+          POSTGRES_PASSWORD: test
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+      redis:
+        image: redis:7-alpine
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 6379:6379
+    env:
+      DATABASE_URL: postgres://test:test@localhost:5432/agent_test
+      REDIS_URL: redis://localhost:6379/0
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v5
+        with:
+          node-version: '20'
+          cache: 'npm'
+          cache-dependency-path: services/agent/package-lock.json
+
+      - name: Install dependencies
+        working-directory: services/agent
+        run: npm ci
+
+      - name: Apply agent-layer schema
+        run: |
+          PGPASSWORD=test psql -h localhost -p 5432 -U test -d agent_test \
+            -f database/init/19_agent_ledger.sql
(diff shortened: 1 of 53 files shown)


mattmorr1:
Closed. The rate table is the wrong shape: Bifrost already prices every call exactly and enforces per-key budgets, so `model_rates` was a fourth copy of pricing rather than a consolidation of two.

Contract 5 is removed from #603, and the Budget seam in #610 now counts iterations locally and reads spend from the gateway. The Python registry keeps its multipliers — deleting those needs a replacement source, and pointing Python's cost path at Bifrost is its own piece of work.

The comment-style gate that rode on this branch moved to #610. Issue #593 is rewritten to cover budget-through-Bifrost.

PR #619, opened 2026-08-11
feat(agent): the domain-free harness, four contracts and the ledger

### LINKED ISSUE #589: Fix the four Phase-0 interface contracts

mattmorr1:
The domain-free harness and the Phase-0 contracts, as one branch. 21 commits,
47 files, **purely additive** — the only modified files are `ci-cd.yml`,
`infra/helm/vigil/values.yaml` and `requirements.txt`.

This is T0 of the one-loop plan: everything that follows branches from main
after it lands.

## What is here

- **Four contracts** — `Tool` port with unopt-outable bounds, a closed and
  versioned event-kind union, `Budget`/`Spend` where the harness counts
  iterations and the gateway reports dollars, and a BullMQ job payload
  sufficient to resume from nothing but itself plus the ledger.
- **The harness** — the tool loop with its own turn cap, iteration cap and
  approval gate; the four injected seams (memory, budget, state, tool
  dispatch); one place where a tool result is wrapped and scanned for
  injection, which no workflow can skip.
- **`agent_events`** — the Postgres ledger, its repository, and the BullMQ
  worker that drains it.
- **A throwaway `tally` workflow** whose only job is to prove the harness
  boundary is real before threat hunting ever touches it. A CI check fails on
  domain vocabulary appearing in `core/`.
- **`POST /api/agent-runs`** — mints a run id, enqueues plain JSON, writes no
  events. The worker opens the ledger.

Budget goes through Bifrost. Reserve/commit is gone: it needed local pricing to
estimate dollars, and the gateway already prices every call exactly and enforces
per-key budgets. `SpendPayload` carries exact tokens and a nullable `cost_usd`,
because the gateway is what bills.

## Verification

CI is green on `18449b2`: Lint Backend, Unit Tests (Backend, Frontend, Agent
Layer, DB-backed), Integration Tests, Security Scans, Build Frontend, Helm Chart.

The one that matters is `tests/integration/test_agent_run_walking_skeleton.py` —
it enqueues from Python and drains with the TypeScript worker, so it proves the
queue name, the job payload and the two BullMQ libraries agree across languages.

`requirements.txt` pins `bullmq==2.15.0`: 3.x pins `redis==7.4.1` and `arq`
requires `redis<6`. 2.15.0 is the newest release that allows it, and wire
compatibility with npm `bullmq@5.81.3` was tested end to end rather than assumed.

## Supersedes

Closes #589, #590, #592 — built here. The remaining agent-loop issues are being
re-cut against the architecture document; see the epic.


mattmorr1:
Rolled into the epic branch `agent-refactor` rather than merging to main separately — the whole refactor lands as one PR later.

`agent-refactor` is cut at `18449b2`, this branch's head, so nothing here is lost. Wave 1 (#621, #623, #624) targets it.

Built here and tracked on #620: #589, #590, #592.

Issue #517, opened 2026-07-30
[Epic] Agent overhaul — parent for the eight parked agent-layer issues


mattmorr1:
**Context.** Eight issues carry the banner *"Folded into the upcoming agent overhaul; picking it up now risks duplicating that work"*, and the repo's `on-hold` label is defined as "Parked pending the agent overhaul" — but no issue tracked that overhaul. This is that umbrella.

It deliberately adds **no new scope**. It exists so the parked work has a parent, and so the overhaul's design can be recorded in one place before any child is unparked or closed as superseded.

### Parked children
- [ ] #497 — unify the ClaudeService + OpenAIAgentService loops into one provider-neutral engine
- [ ] #471 — drop `claude-agent-sdk`; run agents on the neutral loop
- [ ] #474 — Agent module with real behavior (shared `run(context)`)
- [ ] #475 — Investigation-State module (enum + atomic transitions)
- [ ] #476 — Agent-ID registry
- [ ] #472 — Approval-Policy module: one `decide(action)` funnel
- [ ] #473 — MCP config as a single manifest
- [ ] #480 — Workflow convergence: unify file + daemon on the phased engine

### What this epic needs before the children move
- The overhaul's target design written down — at minimum, which of the eight are **absorbed** (close as superseded) versus **sequenced after** it (unpark with a dependency edge).
- An owner and a decision date, so "on hold" has an exit condition instead of being indefinite.

**Label caveat.** `on-hold` is currently also applied to issues unrelated to the agent layer (#308, #332, #333, #442, #513, #514, others), so the label is not a reliable index of this epic's children — the list above is.

PR #638, opened 2026-08-11
feat(agent): one loop — the harness, the tool bridge, and the hunt on it

### LINKED ISSUE #620: Epic: one agent loop — fold the threat hunt and the AI Agent Architecture into main
Vigil runs **five agent loops**, three tool dispatchers and four transport layers, with rate limiting duplicated across three modules. All five are on `main` today:

| Loop | File | Lines |
|---|---|---|
| chat + chat_stream + AgentSDK | `core/llm/harness/claude.py` | 3,487 |
| daemon agent runner | `services/daemon/agent_runner.py` | 1,588 |
| investigation supervision | `services/daemon/orchestrator.py` | 1,337 |
| `_run_phase_loop` | `core/workflows/workflows_service.py` | 1,322 |
| openai harness | `core/llm/harness/openai.py` | 841 |

(Measured on `main` at `886af3b`. The document's own figures differ slightly — it was written in July and the files have moved since.)

This epic replaces all five with one loop, in TypeScript, at `services/agent/`. Every caller — tasks, daemon, generate, compose, llm workers, chat — enters it. The hypothesis-driven threat hunt is the reference workflow. The Python loops are **deleted, not deprecated**.

## Why one loop

An LLM with a hardcoded objective will always search for a result, even when there is nothing there, because that is what a tool loop does. The bar is that untrustworthy AI output is worse than none at all. That is a property of the loop's control flow, and it cannot be enforced five times over in five places.

The three failures the design protects against:

- **Confirmation drift** — every observation updates every hypothesis, so evidence cannot only ever accrue to the favoured one.
- **Greedy querying** — the frontier is scored by expected belief change, not by volume of data returned.
- **Searching for something that is not there** — "no hypothesis can dominate" is a terminal state that reports honestly. Better to fail and not know than to answer inaccurately.

## What already exists

- **The harness** (#619) — four contracts, four injected seams, the tool loop with its own turn cap and approval gate, the `agent_events` ledger, a BullMQ worker, and a throwaway `tally` workflow that proves the domain boundary is real.
- **The hunt** (branch `threat-hunt`, 51 files) — `HuntController`, the closed decision vocabulary, digest, disconfirmation critic, evidence strength, termination, verdicts, entities, enrichment, checkpoints, replay. Tests well. Standalone CLI app, unconnected to Vigil.

Neither is connected to the other, and nothing retires the five loops.

## Decisions

| # | Decision |
|---|---|
| **D1** | **Chat is included**, but does not go through BullMQ — the queue is for durable runs. The agent service exposes SSE directly; `POST /api/claude/chat/stream` keeps its wire contract and proxies, so the web client changes in no way. |
| **D2** | **Tool bridge is HTTP remote dispatch.** One authenticated Python endpoint resolving a `ToolResult`. `RemoteDispatch` implements the existing `ToolDispatch` seam, so neither the loop nor any workflow changes. Native TS adapters stay native. No query-port abstraction: a SIEM is a tool like any other. |
| **D3** | **Architecture per `run_kind`.** `arch/<kind>.yaml` declares roles, prompts, output schemas, dispatch mode and digest policy; a registry maps `run_kind` to it. Routing a swarm for one agent type and a single lead for another is then a file, not a code change. |
| **D4** | **Strangler.** Each Python loop is cut over and deleted in the same PR that proves its replacement. No flags, no two-implementations window. |

## Target layout

```
services/agent/
├── contracts/         tool · events · budget · job              [exists]
├── core/              loop · registry · security · seams · state
│   │                  budget · provider · wire · dispatch       [exists]
│   ├── stream.ts      provider.stream                                NEW
│   ├── context.ts     byte-stable assembly                           NEW
│   └── remote.ts      ToolDispatch over HTTP to Python               NEW
├── arch/              one per run_kind                               NEW
├── workflows/         tally [exists] · hunt · investigate
│                      compose · chat                                 NEW
├── tools/             duckdb · firecrawl · threatfox + remote
├── ledger/            repository.ts                              [exists]
├── worker.ts          BullMQ consumer                            [exists]
└── serve.ts           SSE surface for chat                           NEW
```

## Branching

`#619` merges to `main` first. Everything after it lands on the epic branch **`agent-refactor`**, cut from `main` once #619 is in, and each ticket is one PR onto it. `agent-refactor` rebases onto `main` whenever main moves — a long-lived epic branch silently reverts anything merged behind it, so the check before any merge is:

```
git diff --name-status origin/main agent-refactor | grep -v '^A'
```

Anything unexpected in that list landed on main while the epic sat.

## Tickets

| # | Wave | Ticket |
|---|---|---|
| #619 | 0 | the harness, four contracts and the ledger — **open, merging to main** |
| #621 | 1 | Stream the harness loop |
| #622 | 1 | Byte-stable context assembly and a wall-clock ceiling |
| #623 | 1 | Remote tool dispatch: one endpoint, every Python tool |
| #624 | 1 | Architecture files and run-kind routing |
| #625 | 2 | Hunt ledger on Postgres, gated by fold equivalence |
| #626 | 2 | Re-home the hunt workflow onto the harness |
| #627 | 2 | The hypothesis loop: base rate, discrimination, dominance |
| #628 | 2 | Splunk and Elastic behind remote dispatch |
| #629 | 3 | Investigate workflow: delete `agent_runner` and the orchestrator's loop |
| #630 | 3 | Compose workflow: delete `_run_phase_loop` |
| #631 | 3 | Chat on the harness, over SSE |
| #632 | 3 | Demolition and the one-loop gate |
| #633 | 4 | Resume: per-run lease, watchdog and reaping |
| #634 | 4 | Answer a checkpoint from the console |
| #635 | 4 | Agent image, Helm deployment and the Node CI job |

## Order

```
619  ->  main, then cut agent-refactor
 └─ 621 ─┬─ 622 ─┬─ 626 ─ 627 ─ 628
         │       │     └─ 629 ─┐
         ├─ 623 ─┤             ├─ 632 ─ 633 ─ 634 ─ 635
         └─ 624 ─┴─ 625 ┘      │
                        630 ───┤
                        631 ───┘
```

Wave 1 parallelises three ways. #629, #630 and #631 are independent of each other once it lands.

## The arithmetic

The consolidation is a large net deletion. The epic overall is not, because it also lands a capability that does not exist on main today — and conflating the two would hide whether the consolidation actually paid.

**Replacement — the five loops become one:**

| | lines |
|---|---|
| `claude.py` 3,487 · `agent_runner.py` 1,588 · `openai.py` 841 · `tool_executor.py` 534 | 6,450 |
| partial deletes: the orchestrator's investigation loop, `_run_phase_loop`, the chat router's loop paths | ~2,000 |
| their tests | ~2,000 |
| **deleted** | **~10,450** |
| TypeScript replacing it: harness 1,636 · stream/context/remote/spec/serve ~1,200 · three workflows ~1,200 | ~4,000 |
| **net** | **−6,400** |

**Added on top:** the hunt, ~6,390 lines of source and 4,625 of tests, which is new capability rather than a replacement. Counting it against the deletion would make a flat epic look like a failed consolidation, and hiding it would make a new feature look free.

The document's target for caller code is the sharper version of the same claim: *"callers — each writes ~40 lines"*, against `agent_runner.py`'s current 1,588.

## Done when

- [ ] `main` holds exactly one tool-calling loop, enforced by a CI gate rather than by review
- [ ] A hunt, a daemon investigation, a composed workflow and a chat turn all run through it
- [ ] `claude.py`, `openai.py`, `agent_runner.py`, `_run_phase_loop` and the orchestrator's investigation loop are gone
- [ ] Adding an agent type is an arch file, not a code change
- [ ] The replacement nets **at least 5,000 lines deleted**, measured excluding the hunt, and no caller writes more than ~100 lines to enter the loop

Supersedes #589-#602 and #618, whose content is folded into the tickets above.

---

### The ADR sets

Every ticket belongs to exactly one ADR. If a ticket belongs to none, either it
is out of scope or a decision is missing. Working context in `docs/adr/`
(untracked); this is the index.

| ADR | Decision | Elements | Owned? |
|---|---|---|---|
| 0003 | The ledger is authoritative; the console is delivery | #633, #634 | partial |
| 0006 | Where the harness boundary actually fell | #592, #621 | yes |
| 0007 | One loop; the four it replaces are deleted, not deprecated | #629, #630, #631, #632 | **no** |
| 0008 | A belief that can change: the hypothesis loop | #627, #637 | **no** |
| 0009 | Context assembly: frozen prefix, cache breakpoint, transient tail | #622 | yes |
| 0010 | Remote dispatch: a SIEM is a tool like any other | #623, #628 | yes |
| 0011 | One provider schema, through Bifrost | (enforced by #632) | partial |
| 0012 | Fold equivalence: how a port is proved rather than reviewed | #625, #626 | yes |
| 0013 | The architecture is a file: three layers, topology as a strategy | #624, #636 | yes |
| — | no architectural decision | #635 | n/a |

Also governing throughout: 0001 (TypeScript AI layer), 0002 (domain-free
harness), 0005 (Phase-0 contracts). 0004 (query port) is amended by 0010 — its
guarantees survive, its abstraction did not.

**0007 and 0008 carry the thesis and neither is owned.** Eight ADRs describe
shipped code; the two that make this epic worth doing are still ahead. Nothing
is deleted yet: 9,902 lines of Python loop on `main`, 8,843 still live on
`agent-refactor`.

### Known gaps not owned by any ticket

- **The fold-equivalence gate cannot run in CI** (0012) — the ten historical
  ledgers are personal run data and gitignored. The suite skips cleanly rather
  than passing vacuously. Needs a committable sanitised fixture set.
- **The hunt inbox is still a local file** (0003) — see #634.
- **Net code reduction** is a real criterion but tracked per-PR, not gated.

mattmorr1:
wip. epic branch unifying ai loop. removes 5 loops and replaces it with base arch from threathunt experiments. 

address issue #517


nestor-deeptempo (approved):
🥶 🥶 🥶

Issue #477, opened 2026-07-29
Error taxonomy: replace broad except Exception with typed handling


nestor-deeptempo:
**Coding:** Med · **Vigil context:** High · **Lane:** shared · **Good for:** Vigil-savvy dev; parallelizable per module

**Context.** ~855 broad `except Exception` blocks swallow errors across the codebase, which is exactly how live bugs stay invisible (the approved-MCP no-op and the case↔finding drop both hid behind broad catches). Define a typed exception taxonomy and replace broad catches with specific handling, batched per module. Also decide the fate of the silent JSON-file DB fallback — a dual-source-of-truth hazard.

**Acceptance criteria**
- An exception taxonomy defined; broad catches materially reduced (batch by module).
- The JSON-file fallback keep-or-remove decision recorded (ADR or issue note).

**Files:** across `services/`, `backend/`, `daemon/`; the JSON fallback in the data layer.

nestor-deeptempo:
did a little bit more digging after the convo this morning, maybe this is a better approach

**Coding:** Med · **Vigil context:** High · **Lane:** shared · **Good for:** Vigil-savvy dev; parallelizable per module

**Context.** A typed taxonomy already exists (`core/exceptions.py`: `SOCError` + `DatabaseError`/`AuthError`/etc.) and is effectively dead: raised ~2 times, caught almost nowhere. Adding more classes on top of catch sites that still swallow errors changes nothing.

The real hazards this issue needs to address are behavioral, and two of them are load-bearing enough that they should be settled *before* any renaming:

1. **Swallow-and-return-sentinel.** The dominant pattern is `except Exception: log(...); return []` (or `None`/`0`/`False`). In `core/storage/database_data_service.py` alone, 26 methods do this — a DB error is indistinguishable from a genuinely empty result to every caller. This is exactly how the approved-MCP no-op and the case↔finding drop stayed invisible. (Note the design also doesn't fall through to the fallback on a query error — it just returns empty.)
2. **The JSON-file DB fallback = dual source of truth.** When Postgres is down, the data layer writes to `findings.json`/`cases.json`. This forces every data method to carry two divergent implementations (SQL vs. hand-rolled Python filters that will drift), and on reconnect it simply flips back and orphans whatever was written to the file — no reconciliation. As long as this exists, the broad catches are partly required by it, so we can't cleanly narrow them until its fate is decided.

**Proposed ordering**
1. **Decide the JSON fallback's fate first**
[...]

PR #655, opened 2026-08-13
Refactor/collapse error handling

mattmorr1:
addressing #477 
- error handling was stuck with a ton of try/except loops throughout the code, so I took a look at a couple modules and abstracted it, as well as unified location for imports.
- also pushed service logic out of routers

wip on addressing all remaining issues


nestor-deeptempo:
same idea tracked here https://github.com/Vigil-SOC/vigil/issues/477, i would say abstracting and fixing the catch behavior so it makes sense are the overall goals

mattmorr1:
yeah thats kinda the goal, but only addresses #3 on the proposed approach.

nestor-deeptempo (approved):

PR #658, opened 2026-08-13
Persist and render exact sequence evidence

ShmalexM:
## Summary

- ingest `sequence-evidence.parquet` alongside finding artifacts and attach it idempotently to findings
- validate schema version, exact sequence joins, hashes, and provenance before persistence
- store immutable, content-addressed evidence in a bounded local evidence store
- expose paginated evidence APIs and render deterministic flow and Modbus summaries/tables in finding details
- show coverage, capture provenance, truncation, and evidence-unavailable states explicitly

## Why

Vigil currently receives finding rows but not the original normalized flows that produced each sequence. This change makes raw-flow and Modbus detail independently inspectable without asking an LLM to invent or reconstruct evidence.

The displayed entities are observed endpoints and protocol observations, not inferred asset ownership or network topology.

## Validation

- targeted backend suite — 47 passed
- source-evidence UI tests — 4 passed
- production frontend build passed

### DIFF
--- a/clients/web/src/redesign/data/sourceEvidence.test.ts
+++ b/clients/web/src/redesign/data/sourceEvidence.test.ts
@@ -50,4 +50,68 @@ describe('parseSourceEvidence', () => {
       records: [{ response_code: 'NOERROR' }],
     })).toMatchObject({ status: 'invalid' })
   })
+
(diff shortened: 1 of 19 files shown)


samarmstrong:
Assuming this is stale at this point, please re-open and re-base etc. if not.

Issue #514, opened 2026-07-30
Hardcoded verify=False disables TLS verification in palo_alto and misp tools


craig-dt:
## Summary

Two MCP tool integrations hardcode `verify=False` on their HTTPS requests, disabling TLS certificate verification unconditionally. It is not configurable and not opt-in.

## Why this matters

These are security integrations handling security data, and the calls carry API credentials. Disabling certificate verification exposes them to interception on the path, and it cannot be turned back on without a code change.

It also matters for regulated deployments specifically: customers in these environments are audited on transport security, and a hardcoded verification bypass is the kind of thing that surfaces in a review as a finding against the whole platform, regardless of how narrow it actually is.

## Evidence

- `tools/palo_alto.py:51` and `tools/palo_alto.py:57` — `verify=False, timeout=30`
- `tools/misp.py:52` and `tools/misp.py:61` — `timeout=30, verify=False`

For contrast, other integrations already do this correctly with an operator-controlled flag rather than a constant:

- `services/elastic_service.py:50,72` — `verify=self.verify_ssl`
- `services/vstrike_service.py:237,333,416,485` — `verify=self.verify_ssl`

So the pattern to follow already exists in the codebase.

## Context

Both of these targets are commonly self-hosted with self-signed certificates, which is almost certainly why the flag was hardcoded — it makes a default install work. The fix is to keep that convenience available but make it a deliberate, per-instance choice rather than a permanent default.

## Proposed fix

Replace the hardcoded constant with a configurable `verify_ssl` setting per integration, matching the existing `elastic_service` / `vstrike_service` pattern. Default should be verification **on**; operators with self-signed certificates opt out explicitly, or better, supply a CA bundle.

## Acceptance criteria

- [ ] No hardcoded `verify=False` remains in `tools/`
- [ ] Both integrations expose a `verify_ssl` option consistent with existing integrations
- [ ] Default is verification enabled
- [ ] A grep-based test or lint guard prevents reintroduction

## Related

Surfaced while investigating TLS trust for proxied deployments. Independent of that work — no dependency in either direction.

cc @johnvanlowe @krmayankb

Surfaced during the investigation behind #509, but deliberately not scoped into it.

samarmstrong:
🏭 **Factory groom — needs work.** The factory is not picking this up as written. Edit the description or reply here and it gets reviewed again on a later tick.

The change this issue asked for is already in the tree. #662 landed it; later refactors kept the same behavior in the new layout.

[...]

PR #662, opened 2026-08-13
Issue #514 -- make TLS verification the default and configurable for MCP tools

jmurph1:
Fixes Issue #514 

@mattmorr1


mattmorr1:
@jmurph1 i dont have context for verify ssl check, but why does it fail open (true)

jmurph1:
@mattmorr1 idk what u mean by "fail open"

mattmorr1:
my bad

PR #663, opened 2026-08-14
refactor(tools): move the MCP tool servers to httpx

### LINKED ISSUE #608: Swap the 13 tools/ MCP servers to httpx
**Coding:** Low · **Vigil context:** Low · **Lane:** backend · **Good for:** first-time contributor

**Context.** #462 moved every integration client that runs inside the API server or the daemon off `requests` and onto `httpx`. It deliberately stopped short of the 13 MCP tool servers under `tools/`, because the head-of-line rationale behind #462 / #461 does not apply to them: `mcp-config.json` launches each one as its own subprocess (`{"command": "python3", "args": ["tools/misp.py"]}`), so a blocking `requests` call there stalls only that single-tool subprocess, never the API event loop.

What is left is consistency rather than a defect. After #462, `requests` is gone from `services/`, `backend/`, `daemon/`, `core/` and `database/` — `tools/` is the only place it survives outside `scripts/` and tests.

**Files (13):** `alienvault_otx`, `anyrun`, `azure_ad`, `cape_sandbox`, `carbon_black`, `cloudflare`, `hybrid_analysis`, `ip_geolocation`, `microsoft_defender`, `microsoft_teams`, `misp`, `palo_alto`, `slack`.

Also `tools/url_analysis.py` imports `requests` and never calls it — just delete the import.

**Watch out for** (each of these bit during #462):
- `httpx` defaults to `follow_redirects=False`; `requests` followed redirects. Pass `follow_redirects=True` or endpoints that redirect start reading as failures. Nothing catches this by default.
- `httpx.InvalidURL` is **not** an `httpx.HTTPError`, but `requests` folded both into `RequestException`. `tools/alienvault_otx.py:84` and `tools/cape_sandbox.py:370` catch requests exceptions and need both.
- `tools/ip_geolocation.py:37` and several `tools/palo_alto.py` calls pass no timeout at all. `requests` defaults to none, so add an explicit one rather than relying on httpx's 5s default, which is too short for some of these.
- `tools/palo_alto.py:51,57` use `verify=False` — carry that across.

**Acceptance criteria**
- No `requests` import left under `tools/`.
- Timeouts explicit everywhere; redirect and TLS-verify behaviour preserved.
- `tests/unit/test_backend_tools.py` and `tests/unit/test_cape_sandbox_tool.py` still pass.

**Follows:** #462.

thetosy:
Closes #608.

Moves the 13 MCP tool servers off `requests` and onto `httpx`. One HTTP library
across the integration layer, and the source diff is a straight swap: 39 call
sites, 13 imports, two exception handlers.

## Scope note: three of the 13 have moved

The issue lists all 13 under `tools/`. The #481 reorg landed after it was
filed and moved three into vendor slices, where `mcp-config.json` now spawns
them from:

| Server | Now at |
|---|---|
| cloudflare | `core/integrations/cloudflare/tool.py` |
| microsoft_defender | `core/integrations/microsoft_defender/tool.py` |
| slack | `core/integrations/slack/tool.py` |

10 in `tools/`, 3 in `core/`.

**39 call sites**: cape_sandbox 8, cloudflare 6, azure_ad 4,
microsoft_defender 4, alienvault_otx 3, carbon_black 3, anyrun 2,
hybrid_analysis 2, misp 2, palo_alto 2, and one each for ip_geolocation,
microsoft_teams and slack.

## Migration details that aren't cosmetic

No tool server used a `requests.Session()`, so every site is a bare
`requests.<verb>` → `httpx.<verb>` — same throwaway-client-per-call
lifecycle, same (absent) pooling.

| | |
|---|---|
| **No `follow_redirects`** | requests followed redirects and httpx doesn't, but these are fixed API endpoints: a 3xx means a wrong base URL, not a hop worth chasing. httpx's `raise_for_status()` rejects a 3xx (unlike requests', which passed them), so a misconfigured base URL now surfaces as an error with its status instead of being silently followed. |
| **`HTTPStatusError`, not `HTTPError`** | The one place this branch departs from the issue text, deliberately — see below. |
| **TLS policy** | Untouched. #662 landed mid-review and replaced misp's and palo_alto's hardcoded `verify=False` with `config.get('verify_ssl', True)`; the rebase keeps that expression verbatim at all four sites and only swaps the library around it. `httpx.get`/`post` still take `verify` in 0.28.1. One incidental difference: httpx doesn't use urllib3, so an operator who opts out no longer sees `InsecureRequestWarning` — neither file called `disable_warnings`, so nothing became dead. |
| **Timeouts** | Nothing to do — see below. |

### On the exception mapping

The issue asks for both `httpx.HTTPError` and `httpx.InvalidURL` at
`alienvault_otx` and `cape_sandbox`, reasoning from #609's `_HTTP_ERRORS`.
That mapping is right for what #609 converted and wrong here, because these
two handlers catch a narrower class:

|  | #609's sites | these two |
|---|---|---|
| Caught | `requests.exceptions.RequestException` | `requests.HTTPError` |
| Covers | status + transport + InvalidURL | `raise_for_status()` only |
| Faithful twin | `(httpx.HTTPError, httpx.InvalidURL)` | `httpx.HTTPStatusError` |

`httpx.HTTPError` is the parent of both `HTTPStatusError` and
`TransportError`, so catching it would pull connect errors and timeouts out
of the generic handler they land in today and report them as status errors
with `"status_code": None` — in cape's case also losing the
`logger.exception`. And `InvalidURL` needs no special case here: both files
end in `except Exception`, which is exactly where
`requests.exceptions.InvalidURL` landed, since it subclasses
`RequestException`/`ValueError` and never `HTTPError`.

Two tests fail if this is widened to `HTTPError`.

### On "timeouts explicit everywhere"

Already true, so **no timeout was added or changed**. The issue lists
`ip_geolocation.py:37` and "several `palo_alto.py` calls" as passing none,
but `ip_geolocation` passes `timeout=10` and both `palo_alto` calls pass
`timeout=30` — `git log -S` dates both to the initial commit. Unlike Splunk
and Defender in #609, there was no unbounded-hang defect hiding here.

## Tests

+60 in one file, `tests/unit/integrations/test_tool_servers_httpx.py`, plus
the two existing files retargeted off `module.requests`.

Two halves:

- **Source scan (36)** — AST over `tools/*.py` +
  `core/integrations/*/tool.py`: no `requests` import, and every
  `httpx.<verb>()` states its own `timeout`. Nothing else reaches all 39
  sites, and httpx's 5s fallback is short enough to break a sandbox report
  or a SIEM search. It also covers the `elastic`, `splunk` and `vstrike`
  tool servers, which make no direct `httpx` calls today and so pass
  trivially.
- **Round trips (24)** — respx-mocked, at least one per server: status
  errors reaching the branch that reads `.response` while transport errors
  stay in the generic one; a 3xx surfacing as `{"error": "API error",
  "status": 301}` rather than a silent hop elsewhere; the configured TLS
  policy forwarded (default-on, explicit-on, opt-out) for both appliances;
  and `files=` / `data=` / `params=` / `json=` encoding unchanged — cape's
  two submit branches (open handle and base64 bytes), hybrid's form POST,
  azure_ad's token exchange, Palo Alto's XML-laden query params,
  carbon_black, defender and Teams payloads, and Cloudflare's
  `if resp.content` guard against an empty 204.

Every guard fails against code that lacks it — six mutations, each failing
exactly its own tests and nothing else: dropping a `timeout` fails the scan;
widening `HTTPStatusError` to `HTTPError` fails both transport-error tests;
hardcoding `verify=False` fails 2 TLS params and #662's ratchet; restoring a
`requests` import fails the scan plus that server's round trip; adding
`follow_redirects` back fails the redirect test; and dropping Cloudflare's
empty-body guard fails the 204 test.

Unit suite: 31 failed / 1365 passed on `4a0ed1d`, 31 failed / 1425 passed
here — identical failure list, all `ModuleNotFoundError` in `tests/unit/llm/`.

## Docs

`tools/` is documented as **the 10 tools**: new `tools/README.md` lists them
with their outbound host and the four HTTP conventions (explicit `timeout`, no
`follow_redirects`, config-driven `verify`, `HTTPStatusError`). A server that
lives in a vendor slice belongs to that slice, not to this set. `CLAUDE.md`,
`README.md` and `docs/INTEGRATIONS.md` point at it instead of restating a
count — `CLAUDE.md` said "15+".

## Commits

| SHA | |
|---|---|
| `feaaccf` | `refactor(tools): move the MCP tool servers to httpx` |
| `bcbb818` | `test(tools): cover the tool servers' httpx round trips` |
| `9acc44b` | `docs(tools): document the tool inventory` |
| `a734925` | `chore(deps): narrow the requests note to scripts/` |

Rebased onto `4a0ed1d` (#662); its `verify_ssl` lookup is kept verbatim at all
four misp/palo_alto sites.


mattmorr1:
@thetosy pls fix merge conflicts beforehand, also any chance you could append your test to platform/response somewhere? idk if we really need _ratchet units. also since we now only have 10 tools I think this needs to be documented/adjusted in respective readme

core/integrations/slack/tool.py
mattmorr1, line 1: confused on what you're fixing here re requests follows redirects but slack doesnt, since its a slack api call, pretty sure comment is misleading since there is no redirect needed.

mattmorr1:
also general thing: mcp servers and apis don't need redirect since they're inherently fixed

thetosy:
@mattmorr1  Updated!

nestor-deeptempo (approved):
fixed some merge conflicts and some small changes, thanks!

PR #685, opened 2026-08-15
fix(config): report invalid startup settings cleanly

### LINKED ISSUE #643: config: validate Settings once at process start, exit with a labeled config error
Split out of #576 (part 2 of 2). The dead-fallback cleanup is #642.

## The behavior we want to keep

`get_settings()` builds the whole `Settings` object in one construction and validates every typed field. A single malformed env value — e.g. `DAEMON_BATCH_SIZE=1O` (letter O instead of zero) — raises `pydantic.ValidationError` during that construction. Because `get_settings()` is `lru_cache`d and called from everywhere, the error surfaces at the *first* config read in the process, whichever field that caller actually wanted. A daemon-only typo can fail a request path that never reads a daemon setting.

Fail-fast is the right default for a 24/7 SOC and we are **keeping** it. This issue is about *when and where* the error is raised: once, early, labeled as a configuration failure, with a dedicated exit code — instead of an opaque stack trace from whatever happened to touch config first.

Target behavior: construct `Settings()` once at process start inside `try/except ValidationError`, log `configuration error: <field>: <reason>`, and exit `os.EX_CONFIG` (78). pydantic already names the offending field and value, so no message reformatting is needed.

## Two constraints that make this non-obvious

The naive version — a `try/except ValidationError` inside the backend's `startup_event` — does not work, for two independent reasons. Both need resolving as part of this issue.

**1. `Settings` is already constructed before `startup_event` runs.** There are 14 import-time `get_settings()` call sites across 8 modules:

```
core/auth/auth_service.py:56,57,60,61,64
core/auth/token_blacklist.py:37
core/llm/defaults.py:16
core/storage/connection.py:341
services/api/main.py:99,125
services/api/middleware/auth.py:24
services/api/routers/ingestion.py:44
services/daemon/metrics.py:28
services/worker/jobs.py:16
```

`services/api/main.py:99` is in the app module itself, so importing `services.api.main` at all triggers `Settings` construction. By the time `startup_event` is entered, a bad value has already raised during import and the handler is never reached.

**2. The exit code would not propagate under uvicorn.** uvicorn's lifespan runner catches `BaseException` (`uvicorn/lifespan/on.py`), which includes `SystemExit`, sets `startup_failed`, and the process exits with `STARTUP_FAILURE = 3` (`uvicorn/main.py`). A `sys.exit(78)` raised from inside a startup event is swallowed. The dedicated exit code only survives from an entrypoint outside the ASGI app.

## What is already clean

The daemon path is tractable as specified. `services/daemon/main.py` has almost no module-level imports — just `services.daemon.config` — and none of the 14 sites are in that chain. `main()` calls `DaemonConfig.from_env()`, which is the first `Settings` construction in the process, and it is a plain entrypoint where `sys.exit(os.EX_CONFIG)` propagates normally. Wrapping that call is straightforward.

## The decision to make

The API path needs a placement choice. Rough options, in no particular order:

- A validation call at the very top of `services/api/main.py`, above the imports that reach `get_settings()`. Cheapest, but import-order-sensitive and easy for a later refactor to silently break — would want a ratchet test if we go this way.
- A thin entrypoint module that validates and then imports/serves the app, with the uvicorn command and `scripts/app_up.sh` pointing at it instead of `services.api.main:app`. Keeps the exit code honest; costs an indirection and touches the run scripts, the Dockerfiles and the Helm chart's command.
- Move the 14 import-time reads inside functions so nothing constructs `Settings` at import, then validate in `startup_event` as originally written. Most correct, largest diff, and the exit-code problem from constraint 2 still stands.

Pick one and record the reasoning; whichever wins, the acceptance check is the same.

## Acceptance

With a deliberately malformed value in the environment (e.g. `DAEMON_BATCH_SIZE=1O`):

- `python services/daemon/main.py` logs a single labeled `configuration error: ...` line and exits 78.
- The backend fails at start with the same labeled line rather than a `ValidationError` traceback from an unrelated code path, and does not begin serving requests. Document what its exit code actually is under the chosen approach.
- With a valid environment, neither process logs anything new and startup is unchanged.

Note `os.EX_CONFIG` is POSIX-only — fine for the supported deploy targets, but it is not portable to Windows if that ever matters.

## Out of scope

Adding per-field coerce-or-default validators to restore fail-soft numeric parsing. Considered and declined in #576 — we want fail-fast; noted here only so the decision stays on the record.

mgalore:
## Summary
- validate `Settings` at the daemon and API process boundaries and emit one `configuration error: …` line before configuration-dependent imports proceed
- exit invalid daemon and direct API startup with `os.EX_CONFIG` (78); the API validation lives above its existing import-time settings users so Uvicorn's lifespan handler cannot rewrite the exit code
- make daemon package exports lazy so importing the daemon entry module does not construct settings before its boundary validation
- add subprocess regression coverage for malformed `DAEMON_HEALTH_PORT`

Closes #643.

## Verification
- `uv run --with-requirements <(grep -v '^-e ./mempalace' requirements.txt) python -m pytest tests/unit/api/test_local_startup.py tests/unit/test_startup_config_validation.py -c tests/pytest.ini` — 4 passed
- `uv run --with-requirements <(grep -v '^-e ./mempalace' requirements.txt) python -m pytest tests/unit -c tests/pytest.ini` — 1495 passed, 53 skipped, 2 xfailed; 6 existing approval-workflow tests fail because PostgreSQL is unavailable. The same six tests fail on unmodified `main` with the same connection-refused error.
- `uv run --with-requirements <(grep -v '^-e ./mempalace' requirements.txt) --with import-linter lint-imports` — 3 contracts kept
- `git diff --check` — passed

The repository's standard requirements install is currently blocked in this checkout by `-e ./mempalace`: that directory has neither `pyproject.toml` nor `setup.py`; validation used the identical requirement set with only that unusable editable line excluded.

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -1,10 +1,12 @@
 import json
 import logging
+import os
+import sys
 from functools import lru_cache
 from pathlib import Path
 from typing import Annotated, Any, List, Optional
 
-from pydantic import field_validator
+from pydantic import ValidationError, field_validator
 from pydantic_settings import BaseSettings, NoDecode, SettingsConfigDict
 
 logger = logging.getLogger(__name__)
@@ -258,6 +260,23 @@ class Settings(BaseSettings):
 @lru_cache(maxsize=1)
 def get_settings() -> Settings:
     return Settings()
+
+
+def _format_validation_error(exc: ValidationError) -> str:
+    details = []
+    for error in exc.errors():
+        loc = ".".join(str(part) for part in error.get("loc", ()))
+        msg = error.get("msg", "invalid value")
+        details.append(f"{loc}: {msg}" if loc else msg)
+    return "; ".join(details) or str(exc).replace("\n", " ")
+
+
+def validate_settings_or_exit() -> Settings:
+    try:
+        return get_settings()
+    except ValidationError as exc:
+        print(f"configuration error: {_format_validation_error(exc)}", file=sys.stderr)
+        sys.exit(os.EX_CONFIG)
 
 
 def is_demo_mode() -> bool:
--- a/core/config.py
+++ b/core/config.py
@@ -268,7 +268,7 @@ def _format_validation_error(exc: ValidationError) -> str:
         loc = ".".join(str(part) for part in error.get("loc", ()))
         msg = error.get("msg", "invalid value")
         details.append(f"{loc}: {msg}" if loc else msg)
-    return "; ".join(details) or str(exc).replace("\n", " ")
+    return "; ".join(details) or "invalid settings"
 
 
 def validate_settings_or_exit() -> Settings:
--- a/services/api/main.py
+++ b/services/api/main.py
@@ -19,6 +19,10 @@ project_root = str(_repo_root)
 if project_root not in sys.path:
     sys.path.insert(0, project_root)
 
+from core.config import get_settings, validate_settings_or_exit
+
+validate_settings_or_exit()
+
 from fastapi import FastAPI, Depends
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.staticfiles import StaticFiles
@@ -32,7 +36,6 @@ from services.api.middleware.security_headers import SecurityHeadersMiddleware
 
 from services.api.discovery import mount_routers
 from services.api.middleware.auth import get_current_active_user
-from core.config import get_settings
 from core.platform.monitoring import init_sentry, PROMETHEUS_AVAILABLE, get_metrics_response
 
 # Single source of truth for the "require an authenticated active user"
--- a/services/daemon/__init__.py
+++ b/services/daemon/__init__.py
@@ -1,6 +1,22 @@
 """SOC Daemon - Headless autonomous security operations service."""
 
-from services.daemon.config import DaemonConfig
-from services.daemon.main import SOCDaemon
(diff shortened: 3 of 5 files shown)


nestor-deeptempo (changes requested):
thanks for the contribution! just one small comment worth addressing now before merging

> core/config.py:271
> 
> return "; ".join(details) or str(exc).replace("\n", " ")
> 
> details is built from each error's loc and msg. That's the careful path — pydantic's msg is generic ("Input should be a valid integer"), so the value you actually typed never gets printed. str(exc) is the opposite: it renders the full repr, including input_value='not-an-int', input_type=str.
> 
> So if a Settings field holding a credential ever fails validation and this branch runs, the secret goes to stderr and into logs/backend.log. It can't happen today — a ValidationError always carries at least one error, and pydantic v2 always populates msg, so details is never empty. The exposure is a future custom validator or a pydantic upgrade quietly making it reachable. or "invalid settings" removes the possibility.

but overall looks good!

mgalore:
Addressed in d85b43f: validation formatting now uses the fixed `invalid settings` fallback when no safe field-level details are available, and a regression test covers that path.

nestor-deeptempo (approved):
thank you for the changes! lgtm

PR #686, opened 2026-08-16
fix: daemon mode: LLM worker gated on orchestrator.enabled starve

### LINKED ISSUE #581: Daemon mode: LLM worker gated on orchestrator.enabled starves chat/insights (no queue consumer)
## Summary

In daemon mode (`./start.sh -d`), the ARQ LLM worker is the **only** consumer of the `arq:llm` queue, but its lifecycle is gated on the *autonomous-orchestrator* toggle (`orchestrator.settings.enabled`). When the orchestrator is disabled, the worker is never started — so **every** LLM call that goes through `LLMGateway` (chat, AI insights, agent runner, daemon triage) enqueues a job with no consumer and hangs. This conflates "autonomous SOC mode is off" with "no LLM processing at all," which are unrelated concerns.

## Repro

1. Start the stack in daemon mode: `./start.sh -d`.
2. Ensure the orchestrator is disabled — this is the default: `daemon/llm_worker_manager.py` initializes `self._enabled = False`, and `_sync_enabled_from_db()` only flips it true when an `orchestrator.settings` SystemConfig row exists with `enabled: true`. A fresh install (no row) stays disabled.
3. Open the UI and send a chat message (or hit any endpoint that routes through `services/llm_gateway.py`).
4. `backend/api/claude.py` → `LLMGateway.submit_chat()` enqueues onto `arq:llm`, but no worker is draining it → the request hangs / times out with no result.

## Root cause

- `daemon/main.py` starts the worker **only** via `LLMWorkerManager` (`daemon/main.py:145` → `self._llm_worker_manager.run(...)`); there is no direct worker start in daemon mode.
- `LLMWorkerManager` polls `orchestrator.settings.enabled` and starts/stops the `services.run_llm_worker` subprocess to match it (`daemon/llm_worker_manager.py`).
- But `services/llm_gateway.py` is the enqueue path for **all** LLM traffic, not just autonomous orchestration. Its callers include `backend/api/claude.py` (chat) and `backend/services/ai_insights_service.py`, neither of which has anything to do with the orchestrator toggle.

So the worker's availability — which every LLM feature depends on — is coupled to a flag that is only meant to govern autonomous incident response.

## Impact

- Daemon-mode deployments with autonomous mode off have chat and AI insights silently broken (requests hang, no clear error).
- Turning the orchestrator **off** at runtime tears down the worker and breaks unrelated LLM features until it is turned back on.

## Not affected

- Foreground `./start.sh` — starts the worker directly (`start.sh:137`), independent of the orchestrator flag.
- docker-compose and the Helm chart — run the worker as its own always-on service/Deployment, no manager involved.

## Suggested direction

Decouple the worker's lifecycle from `orchestrator.settings.enabled`. That toggle should continue to govern autonomous incident response only, not LLM queue processing.

The fix is to remove the supervision from the daemon entirely rather than to make it unconditional. Compose and the Helm chart already run the worker as its own container alongside the daemon (`llm-worker` in `infra/docker/docker-compose.yml`, `llm-worker-deployment.yaml` in the chart), so a daemon that always starts one would give those deployments two consumers on the same `arq:llm` queue — double the LLM traffic, inside a StatefulSet sized for the daemon alone, and invisible to the KEDA ScaledObject that owns worker scaling.

Relevant history: `services/daemon/llm_worker_manager.py` exists *only* because of this gating. Until April 2026 the daemon start script started the worker directly, with a pidfile. `c5d898f1` (#90) deleted that and introduced the manager, for one stated purpose — *"toggling Auto Investigate in the Settings UI now takes full effect without any daemon restart — the worker subprocess starts/stops within seconds."* That is the behaviour this issue identifies as wrong, so the module has no remaining job.

Proposed change:

1. Delete `services/daemon/llm_worker_manager.py`.
2. Remove its wiring from `services/daemon/main.py` (construction in `_init_components`, task in `run`).
3. Start the worker in the daemon branch of `start.sh`, alongside the existing `backend.pid` / `daemon.pid` writes — restoring what #90 removed. `shutdown_all.sh:18` already cleans up `logs/llm_worker.pid`, which nothing currently writes. Add `llm_worker` to the `tail -f logs/{...}.log` hint too.
4. Update the `llm_worker_manager.py` entry in the repo-structure tree in `CLAUDE.md`.

The result is one rule in every deployment: whoever starts the stack starts the worker — `start.sh` locally, the platform in containers. Nothing supervises another process's process.

Tradeoff: local daemon mode no longer auto-restarts a crashed worker. That matches how the backend and frontend already behave under `./start.sh -d`; Compose has `restart: unless-stopped` and Kubernetes has the Deployment.

The deletion also removes three latent problems in that file: a `mkdir` outside its `try` that raises `PermissionError` in the container images (where `/app` is root-owned and the process runs as non-root), silently killing the manager task with nothing logged; a file handle leaked on every worker restart; and an unbounded restart loop with no backoff.

Mr-Neutr0n:
Fixes #581

Decoupled the daemon's ARQ LLM worker from the `orchestrator.settings.enabled` toggle: `LLMWorkerManager` in services/daemon/llm_worker_manager.py now starts the worker unconditionally when the daemon runs, restarts it if it dies, and only stops it on shutdown — removing `_sync_enabled_from_db()`, the DB polling loop, and the `_enabled` gating so chat/insights queue consumption no longer starves when the orchestrator is disabled. Updated the log line in services/daemon/main.py:167 to reflect this. The orchestrator toggle itself still governs autonomous incident response only. (The issue text's suggested direction was followed as guidance, not instructions.)

Could not run the suite locally: urn _bootstrap._gcd_import(name[level:], package, level). Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/services/daemon/llm_worker_manager.py
+++ b/services/daemon/llm_worker_manager.py
@@ -1,5 +1,7 @@
-# Supervises the ARQ worker as a child process of the daemon: polls the
-# orchestrator.settings enabled flag and starts, stops or restarts to match it.
+# Supervises the ARQ worker as a child process of the daemon. The worker drains
+# the arq:llm queue for ALL LLM traffic (chat, AI insights, agent runner, daemon
+# triage), not just autonomous orchestration, so it runs unconditionally while
+# the daemon runs — matching foreground, docker-compose and the Helm chart.
 
 import asyncio
 import logging
@@ -15,54 +17,29 @@ PROJECT_ROOT = str(Path(__file__).resolve().parents[2])
 # The -m entrypoint, shared with compose, the Helm Deployment and start.sh.
 WORKER_MODULE = "services.worker"
 
-_POLL_INTERVAL = 5
-
 
 class LLMWorkerManager:
     def __init__(self):
         self._process: subprocess.Popen | None = None
-        self._enabled = False
 
     async def run(self, shutdown_event: asyncio.Event):
         logger.info("LLM Worker Manager started")
 
-        while not shutdown_event.is_set():
-            self._sync_enabled_from_db()
+        if not self._is_running():
+            self._start_worker()
 
-            if self._enabled and not self._is_running():
+        while not shutdown_event.is_set():
+            if not self._is_running():
+                logger.warning("LLM Worker exited unexpectedly — restarting")
                 self._start_worker()
-            elif not self._enabled and self._is_running():
-                self._stop_worker()
 
             try:  # sleep, but wake immediately on shutdown
-                await asyncio.wait_for(shutdown_event.wait(), timeout=_POLL_INTERVAL)
+                await asyncio.wait_for(shutdown_event.wait(), timeout=5)
             except asyncio.TimeoutError:
                 pass
 
         self._stop_worker()
         logger.info("LLM Worker Manager shutdown complete")
-
-    def _sync_enabled_from_db(self):
-        try:
-            from core.storage.connection import get_db_manager
-            from core.storage.models import SystemConfig
-
-            with get_db_manager().session_scope() as session:
-                cfg = (
-                    session.query(SystemConfig)
-                    .filter_by(key="orchestrator.settings")
-                    .first()
-                )
-                if cfg and isinstance(cfg.value, dict):
-                    db_enabled = bool(cfg.value.get("enabled", False))
-                    if db_enabled != self._enabled:
-                        self._enabled = db_enabled
-                        logger.info(
-                            "LLM Worker %s (synced from DB)",
-                            "ENABLED" if db_enabled else "DISABLED",
-                        )
-        except Exception:
-            pass  # DB not ready yet — keep previous state
 
     def _start_worker(self):
         # Exports the parent env into a child process; not a config read.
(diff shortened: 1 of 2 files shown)


nestor-deeptempo:
I've updated the suggested direction on this one.

The original text said the worker "should run whenever the daemon runs (matching how foreground, docker-compose, and Helm already treat it — always-on)." That collapses two different arrangements: foreground means the start script starts it, while Compose and Helm mean it's its own service. Neither is "the daemon starts one." Read as written, it points at making the daemon start a worker unconditionally — which would put two consumers on arq:llm in every containerized deployment. That's my wording, and #686 followed it in good faith.

Digging into why the gating existed in the first place settles the direction. services/daemon/llm_worker_manager.py was added by c5d898f1 (#90, 2026-04-16), which deleted a direct worker start from the daemon start script and replaced it with a subprocess manager, so that flipping Auto Investigate in Settings would stop and start the worker within seconds. That is precisely the coupling this issue is about — so the right fix is to delete the module and restore the worker start to start.sh, not to keep the module and strip its gating.

Details are in the updated section above. Also dropping the good first issue label: this is now a module deletion plus a start-script change, which is past newcomer scope.

It's up to you if you want to take the updated issue, lmk if you still want to tackle this!

nestor-deeptempo:
closing because the issue updated and i didn't hear a response here

PR #700, opened 2026-08-19
feat(hunt): make a threat hunt runnable from the console, and worth reading

mattmorr1:
Makes `threat-hunt` runnable end to end from the console: launch it, watch it,
steer it, stop it, and read what it concluded.

Most of the diff is what that took to actually work. The hypothesis loop ran, but
streamed calls died at 30s, a dead stream was retried three times, tool results
detached from the turn that asked for them, and a run waiting on a person was
killed as stalled. Each of those was found by running the previous fix.

**Shape:** 9,859 lines — 5,445 tests, 4,414 source. 4 new source files
(`hunt/narrative.ts`, `run_cancel.py`, and two dev scripts); everything else edits
files that had to change.

**Worth a look:**
- `services/agent/workflows/hunt/` — the loop: verdicts, termination, cost ceilings
- `clients/web/.../WorkflowsScreen.tsx` — +793, now 2,631 lines / 58 components.
  Wants splitting; better as its own PR than inside this one.
- `core/workflows/playbook_resolver.py` — capability→tool binding is now built from
  `Candidate(server, tools)` rather than guessed strings, with a ratchet checking the
  in-repo names against each server's real `list_tools()`.

**Not in scope, but in the diff:** the firecrawl connector. Unrelated to hunts; say
the word and it comes out.

Green: 14 checks. `lint-imports` 3/3, one-loop gate clean, 1709 backend + 872 agent +
169 web unit, 84 integration.


nestor-deeptempo (approved):
🫡

PR #705, opened 2026-08-21
refactor(cases): keep literal types on lookups

samarmstrong:
The lookup maps and helpers behind the cases screens were annotated in ways
that threw away type information TypeScript already had.

## What changed

**`CasesScreen.tsx`**

- `PRIO_RANK` and the tab `groups` map were annotated `Record<K, V>`. That
  keeps the key-exhaustiveness check but flattens the values. Both now use
  `satisfies`, so a missing tab or priority is still a compile error while
  the concrete value types survive.

**`CaseSections.tsx`**

- `SLA_STATUS_CLASS` was `Record<string, string>` even though `SLA['status']`
  is already the union `active | breached | met | paused`. Keying it off that
  union means the map can no longer drift from the interface.
- `PRIO_ORDER` and `REL_LABEL` are looked up with strings that arrive
  unparsed from the API, so widening the map hid the real question of whether
  the key is known. Each now declares its domain type (`TaskPriority`,
  `RelationshipType`) next to the map and narrows once through a predicate,
  making the fallbacks explicit rather than implied by an `??`. The
  related-cases `<select>` builds its options from `Object.entries`, so the
  option values are the map's own keys instead of bare strings.
- `threatLevel` returned four fixed label/class pairs behind a widened
  `{ label: string; cls: string }` annotation; dropping it lets callers see
  the actual values.

## Behaviour

No change. Each rewritten lookup resolves to the same value as before,
including the unknown-key fallbacks.

## Checks

- `npx tsc --noEmit` — clean
- `npm run test:run` — 14 files, 71 tests passing (unchanged)
- `npm run lint` — 16 problems (2 errors, 14 warnings), identical to `main`

### DIFF
--- a/clients/web/src/redesign/screens/cases/CaseSections.tsx
+++ b/clients/web/src/redesign/screens/cases/CaseSections.tsx
@@ -242,7 +242,12 @@ interface Task {
   status: string
   completed_at?: string
 }
-const PRIO_ORDER: Record<string, number> = { critical: 0, high: 1, medium: 2, low: 3 }
+const PRIO_ORDER = { critical: 0, high: 1, medium: 2, low: 3 } as const
+type TaskPriority = keyof typeof PRIO_ORDER
+const isTaskPriority = (v: string | undefined): v is TaskPriority =>
+  v !== undefined && Object.prototype.hasOwnProperty.call(PRIO_ORDER, v)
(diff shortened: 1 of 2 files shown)


nestor-deeptempo (approved):
lgtm

Issue #707, opened 2026-08-21
pgadmin publishes an unauthenticated DB console on 0.0.0.0:5050


craig-dt:
## Summary

`infra/docker/docker-compose.yml` publishes pgAdmin on `0.0.0.0:5050` with a default password *and* its own authentication turned off. Anyone who can reach the port gets an unauthenticated database console pointed at the Vigil Postgres instance.

This is the same class of problem as #587, but it was deliberately left out of that issue's scope because pgAdmin sits behind an opt-in profile rather than starting by default.

## Current behaviour

From `infra/docker/docker-compose.yml`:

```yaml
  pgadmin:
    image: dpage/pgadmin4:latest
    container_name: deeptempo-pgadmin
    environment:
      PGADMIN_DEFAULT_EMAIL: ${PGADMIN_EMAIL:-admin@deeptempo.ai}
      PGADMIN_DEFAULT_PASSWORD: ${PGADMIN_PASSWORD:-admin}
      PGADMIN_CONFIG_SERVER_MODE: 'False'
    ports:
      - "5050:80"
    profiles:
      - dev
```

Three things compound here:

- **`5050:80` binds all interfaces**, so the console is reachable from the operator's LAN, not just the host.
- **`PGADMIN_PASSWORD` defaults to `admin`**, and the default is committed to this repo.
- **`PGADMIN_CONFIG_SERVER_MODE: 'False'`** puts pgAdmin in desktop mode, which *disables the login prompt entirely*. The default password is therefore not even the last line of defence — there is no prompt to enter it into.

The console has `depends_on: postgres` and is wired to `deeptempo-network`, so it reaches the database holding `findings`, `cases`, `approval_actions` and `llm_provider_configs`.

## Mitigating factor

`pgadmin` is behind `profiles: [dev]`, so a plain `docker compose up` does not start it. Reaching this state takes a deliberate `--profile dev`. That is why #587 scoped it out, and why this is filed separately rather than folded in.

It is still worth fixing: `--profile dev` is exactly what a developer runs, and nothing signals that doing so opens an unauthenticated database console to the network.

## Suggested fix

At minimum, bind to loopback, consistent with the fix in #587:

```yaml
    ports:
      - "127.0.0.1:5050:80"
```

Worth considering alongside it:

- Drop `PGADMIN_CONFIG_SERVER_MODE: 'False'` so the login prompt is restored, or
[...]

PR #708, opened 2026-08-21
fix(compose): bind Postgres and Redis to loopback

### LINKED ISSUE #587: docker-compose publishes Postgres, Redis and Bifrost on 0.0.0.0
## Summary

`docker/docker-compose.yml` publishes PostgreSQL, Redis and the Bifrost gateway on `0.0.0.0`, so all three are reachable from anywhere on the operator's network. For a SOC platform this is worth tightening — the machine running Vigil holds security findings, case data and LLM credentials.

Tested on `v0.4.0` (`8abecc2`).

## Current behaviour

```console
$ docker ps --format '{{.Names}}\t{{.Ports}}'
deeptempo-postgres   0.0.0.0:5432->5432/tcp, [::]:5432->5432/tcp
deeptempo-redis      0.0.0.0:6379->6379/tcp, [::]:6379->6379/tcp
deeptempo-bifrost    0.0.0.0:8080->8080/tcp, [::]:8080->8080/tcp
```

From `docker/docker-compose.yml`:

- line 13 — `- "5432:5432"`
- line 33 — `- "6379:6379"`
- line 207 — `- "8080:8080"`

## Why each one matters

- **Postgres** ships with `POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-deeptempo_secure_password_change_me}`. The default is in the repo, so an install that hasn't overridden it is open to anyone on the LAN. It holds `findings`, `cases`, `approval_actions` and `llm_provider_configs`.
- **Redis** runs with no `requirepass` at all — it's the ARQ job queue and session store.
- **Bifrost** exposes an unauthenticated admin API on the same port as inference (`auth_config.is_enabled: false` by default). Anyone who can reach `:8080` can read and rewrite provider configuration, including `PUT /api/providers/{name}/keys/{key_id}`.

None of the three has a legitimate consumer outside the host: the backend, daemon and llm-worker all reach them over the `deeptempo-network` bridge by service name, and a host-run backend (`start.sh`) reaches them over loopback.

## Suggested fix

```yaml
- "127.0.0.1:5432:5432"
- "127.0.0.1:6379:6379"
- "127.0.0.1:8080:8080"
```

Verified locally: after `docker compose up -d --force-recreate postgres redis bifrost`, `docker ps` shows `127.0.0.1:...` for all three, and the backend, frontend, Bifrost health check and end-to-end LLM traffic all keep working. Bifrost's UI at `http://localhost:8080` is still reachable from the host.

Happy to open a PR.

## Deliberately not included

`backend` (`6987`), the daemon webhook (`8081`) and the daemon's Prometheus/health ports (`9090`, `9091`) are left alone — those have real remote consumers (analyst browsers, SIEM push, metrics scraping), so binding them to loopback would be wrong. The three above have none.

craig-dt:
Closes #587.

## What changed

Two port mappings in `infra/docker/docker-compose.yml` now bind loopback instead of every interface:

| service | before | after |
| --- | --- | --- |
| postgres | `"5432:5432"` | `"127.0.0.1:5432:5432"` |
| redis | `"6379:6379"` | `"127.0.0.1:6379:6379"` |

Each carries an inline comment explaining why, so the next person to read the line doesn't have to find this PR.

> **Bifrost is deliberately not touched here — see the note to reviewers below.**

This implements what @MagicWand18 proposed in #587, at the file's post-reorg path: `refactor/reorg-481` moved it from `docker/` to `infra/docker/`, so the line numbers in the issue no longer match.

## Why

Per the issue: Postgres ships a default password that lives in this repo and holds `findings`, `cases` and `llm_provider_configs`; Redis runs with no `requirepass` at all. Neither has a consumer outside the host — the backend, daemon and workers reach them by service name over the `deeptempo-network` bridge, and a host-run backend reaches them over loopback.

## ⚠️ Note to reviewers: interaction with #703

**#703 already binds Bifrost to `127.0.0.1:8080:8080`**, in this same file, on this same line, with its own rationale comment. Changing it here too would only produce a conflict, so this branch leaves that line **byte-identical to main**.

Bifrost is instead held in the new gate as a **strict xfail**:

```python
PENDING_ELSEWHERE = {
    "bifrost": "loopback bind ships with PR #703 (Bifrost config via Vigil settings)",
}
```

While #703 is open, the Bifrost check fails, that is expected, and CI is green. **The moment #703 merges, the check starts passing, `strict=True` turns that unexpected pass into a failure, and main goes red until those three lines are deleted.** That is intentional — the exception cannot quietly outlive its reason the way an allowlist entry would — but it does mean **whoever merges the second of these two PRs should delete the `PENDING_ELSEWHERE` block in the same PR.** Flagging it so it isn't a surprise. @mattmorr1, since #703 is yours.

If you would rather avoid the interaction entirely, merging #703 first and asking me to drop the block from this branch also works.

## Regression gate

A two-line YAML change is easy to undo by accident, so this adds `tests/security/test_compose_port_binding.py`, in the style of the existing `test_route_auth_coverage.py`. Every service that starts on a plain `docker compose up` must publish on loopback, unless that **exact host port** is declared in `SHARED_HOST_PORTS`.

The allowlist is keyed per `(service, port)` rather than per service on purpose: exempting a whole service would let a debug database port added to `backend` later ride in unnoticed. It currently declares `backend:6987` (analyst browsers) and `soc-daemon:8081/9090/9091`, matching the issue's "deliberately not included" section.

One honest caveat recorded in the code: `8081` is well justified — it fails closed without `DAEMON_WEBHOOK_TOKEN`. The case for `9090`/`9091` is thinner than it looks, since `infra/docker/prometheus.yml` scrapes `soc-daemon:9090` over the compose bridge by service name rather than through the published host mapping, and `9091` serves `/health` and `/status` unauthenticated. This PR keeps the issue's scope and does not change them, but the comment says so rather than asserting a justification the repo contradicts. Narrowing them is worth its own issue.

Also gated: `network_mode: host`, which publishes everything while declaring no `ports:` and would otherwise pass with nothing to check. Two anti-rot tests assert both allowlists name services that still exist, so a rename can't silently stop the gating. The module resolves the repo root from its own path rather than importing `core.config`, so an unrelated dependency break can't disarm it — which is the state `tests/security/test_unauth_endpoints.py` is in right now.

## Out of scope

`pgadmin` publishes `5050` on all interfaces with a default password and `PGADMIN_CONFIG_SERVER_MODE: 'False'`, which disables its login outright. It sits behind the opt-in `dev` profile, so it was outside #587's scope. Filed as **#707**; the test docstring points at it. `clients/desktop/standalone/docker-compose.yml` is clean today but ungated — noted in the docstring as a known gap.

## How it was tested

Written test-first, then verified adversarially by mutating the compose file and confirming the right test fails each time:

| mutation | result |
| --- | --- |
| revert postgres to `"5432:5432"` | ✅ fails on `[postgres]` |
| simulate #703 landing (bifrost → loopback) | ✅ strict xfail self-destructs as designed |
| add a debug `"5432:5432"` to allowlisted `backend` | ✅ fails on `[backend]` — the old per-service allowlist would have passed this |
| `network_mode: host` on a clean service | ✅ fails on `[llm-worker]` |

Also: `docker compose -f infra/docker/docker-compose.yml config` exits 0; `flake8` and `black --check` clean on the new file; CI is unaffected since the workflows use Actions `services:` containers, not this compose file.

Pre-existing failures on this base, confirmed by stashing these changes and re-running: `test_unauth_endpoints.py` collection error (`No module named 'bullmq'`) and two failures in `test_route_auth_coverage.py`.

An independent reviewer went over the first revision; the per-port allowlist, the `network_mode: host` check and the two corrected comments come from that pass. The compose rationale comments and the strict-xfail approach are lifted from a parallel local branch, `fix/587-loopback-data-plane-ports`.

## Note on attribution

@MagicWand18 reported this and offered to send the PR, and @mattmorr1 encouraged them to. Apologies for the overlap — this came up while hardening our own deployment. Happy to close this in favour of theirs if they have one in flight; the regression test could be cherry-picked either way.


mattmorr1:
not sure if 127.0.0.1 is guaranteed ip address for launch. additionally, could you wrap the test minimally into another one of the bifrost/up tests, as well as fix the 8080 issue #587 ? seems related.

nestor-deeptempo (approved):
added bifrost back in as the other pr was closed, lgtm

PR #722, opened 2026-08-26
fix(daemon): stop gating the LLM worker on orchestrator.settings.enabled

### LINKED ISSUE #581: Daemon mode: LLM worker gated on orchestrator.enabled starves chat/insights (no queue consumer)

AmirF194:
## Summary

Daemon mode's `LLMWorkerManager` started the ARQ `services.worker` subprocess
only while the DB row `orchestrator.settings` had `enabled: true`. That
toggle governs autonomous incident response, but `LLMGateway` (chat, AI
insights, agent runner, daemon triage) enqueues onto the same `arq:llm`
queue regardless. A fresh install carries no `orchestrator.settings` row, so
the worker never starts, and every LLM call sent through the daemon hangs
with no consumer to drain it.

Per the root cause in the issue, `LLMWorkerManager` exists only to make the
Settings-UI toggle take effect within seconds. That is the coupling that
causes the hang, so the fix removes the module rather than changing what it
gates: the daemon branch of `start.sh` now starts `services.worker` itself,
unconditionally, the same way the foreground branch already does. Also
writes `logs/llm_worker.pid`, which `shutdown_all.sh` already cleans up but
nothing wrote.

`WORKER_MODULE` moved from the deleted module to `services/worker/__init__.py`,
which is what `tests/unit/worker/test_worker_entrypoint.py` now imports.

## Verification

- New test `test_daemon_mode_starts_worker_unconditionally` fails on `main`
  (daemon branch never names `services.worker`) and passes on this branch,
  run in a clean `python:3.10-slim` container against both trees.
- Reproduced the underlying hang mechanism directly: with no
  `orchestrator.settings` row (the documented fresh-install default),
  `LLMWorkerManager.run()` never calls `_start_worker` across two full poll
  cycles, confirming the worker is never launched.
- `pytest tests/unit/ tests/security/ -m "not external_service"`: 1687
  passed, 34 skipped (pre-existing), 0 failed.
- `lint-imports`: 3 contracts kept, 0 broken. `check_one_loop.py`: clean.
- Did not run `tests/integration/` (needs Postgres/Redis/Kafka) or exercise
  the real chat/insights endpoints end to end; the regression test and the
  `LLMWorkerManager` repro above cover the mechanism, not a live HTTP round
  trip.

Fixes #581


start.sh
```diff
@@ -169,6 +169,10 @@ else
     nohup "${PWD}/venv/bin/python" services/daemon/main.py > logs/daemon.log 2>&1 &
     echo $! > logs/daemon.pid
 
+    # Started unconditionally, independent of orchestrator.settings (#581).
```
nestor-deeptempo, line 172: shutdown_all.sh now has only one kill path for the worker. Before this change the worker was a child of the daemon, so pkill -f "services/daemon/main.py" (shutdown_all.sh:27) → SIGTERM → _shutdown_event → _stop_worker() reliably reaped it. Now it's a top-level process whose only cleanup is logs/llm_worker.pid (shutdown_all.sh:18) — and that loop only rms the pidfile when kill succeeds, so any path that loses or staleifies the pidfile leaves an orphan.

That orphan is reachable, not theoretical: start.sh -d's only "already running" guard is pgrep -f 'uvicorn services.api.main:app' (start.sh:150), and shutdown_all.sh kills the backend three separate ways. So the backend dies, the orphan worker lives, and the next ./start.sh -d starts a second consumer on arq:llm — the exact duplicate-drain the issue is careful to rule out for compose/Helm, except locally, and with the orphan possibly running pre-checkout code.

Every other Vigil process has belt-and-braces coverage here. One line alongside the others:

pkill -f 'services\.worker' 2>/dev/null || true

Escape the dot — unescaped, -f services.worker also matches services/worker/... in an editor or a pytest tests/unit/worker cmdline on a dev box.

nestor-deeptempo (changes requested):
hey @AmirF194, thanks for the contribution! one change i'm requesting before merging, overall looks good!

AmirF194:
Good catch, and the escaping note too. Added the pkill fallback right next to the daemon's own line in shutdown_all.sh: `pkill -f 'services\.worker' 2>/dev/null || true`. Verified the escaped pattern matches the real worker invocation but not a services/worker/... path. Pushed to 8dd31b3.

craig-dt:
Heads-up that the ball may be on the maintainer side here rather than @AmirF194's: the push answering the review (`8dd31b3`, 07:09 today) is sitting at `conclusion: action_required`, so CI hasn't run on it. The two pushes before it — `6a3fa5a3` and `72b91909` — both went green, so this looks like the per-push workflow approval for an outside contributor rather than anything wrong with the branch. Someone with write access clicking "Approve and run workflows" should unstick it.

Worth knowing it isn't a blanket gate on the author: #686 had the same `author_association=NONE` and got its 14 checks without trouble.

nestor-deeptempo (approved):
lgtm! thanks @AmirF194

AmirF194:
Thanks for the review and the merge. Glad the shutdown fix landed cleanly alongside the main change.

PR #724, opened 2026-08-26
fix(daemon): keep synchronous DB calls off the event loop

bluejazz822:
The DB layer is synchronous SQLAlchemy, so a call made from an `async def` freezes the loop for the whole round-trip. `services/daemon/main.py` runs eight subsystems on that one loop, so a blocking write stalls polling, Kafka ingest and the orchestrator alike.

## What changed

Six call sites now go through `asyncio.to_thread`:

| file | call site |
|---|---|
| `processor.py` | `_store_finding`, `_update_finding`, the backfill sweeper batch query |
| `poller.py` | the direct-store fallback in `_enqueue_finding` |
| `orchestrator.py` | the existing-review lookup and the case read in `_maybe_trigger_case_review` |

The orchestrator lookup also moved into a helper so the id is read inside the session scope — returning the ORM instance would hand the caller a detached object.

A second commit makes `max_concurrent_tasks` configurable. It gates both the worker-coroutine count and the enrichment LLM semaphore in `processor.py`, and nothing set it from configuration, so 5 was a throughput ceiling that could only be raised by editing the source. The default is unchanged, so it is inert until somebody sets it.

## Tests

`tests/unit/daemon/test_db_calls_off_event_loop.py` asserts the offload by **thread identity rather than by timing**, so it is deterministic — a call reporting the loop own thread id never went through a worker thread.

Each test was checked against the unfixed code first and fails there with the assertion it is meant to fail with, e.g.:

```
AssertionError: get_case ran on the event loop thread — it is synchronous
SQLAlchemy and must go through asyncio.to_thread
```

## Notes

- No behaviour change beyond where the work runs.
- `flake8` reports no new findings on any touched file. `black` is not run on them: these files are not currently black-clean on `main`, and reformatting them would bury the change.
- Found while running the daemon under load in a downstream deployment.

### DIFF
--- a/services/daemon/poller.py
+++ b/services/daemon/poller.py
@@ -595,7 +595,9 @@ class DataPoller:
 
         try:
             from core.ingestion.ingestion_service import IngestionService
-            IngestionService().ingest_finding(finding)
+            # Synchronous SQLAlchemy — off the event loop, otherwise this
+            # fallback blocks the polling loop for the whole write.
+            await asyncio.to_thread(IngestionService().ingest_finding, finding)
             return True
         except Exception:
             logger.exception(
--- a/core/config.py
+++ b/core/config.py
@@ -196,6 +196,10 @@ class Settings(BaseSettings):
     daemon_auto_triage: bool = True
     daemon_auto_enrich: bool = True
     daemon_batch_size: int = 10
+    # Processor worker coroutines, and the cap on concurrent enrichment LLM
+    # calls — processor.py uses this value for both. Was hard-coded at 5 with
+    # no way to raise it, which put a ceiling on daemon throughput.
+    daemon_max_concurrent_tasks: int = 5
     daemon_enrich_max_inflight: int = 50
     daemon_enrich_backfill: bool = True
     daemon_enrich_backfill_interval: int = 300
--- a/env.example
+++ b/env.example
@@ -536,6 +536,8 @@ DAEMON_AUTO_TRIAGE="true"
 DAEMON_AUTO_ENRICH="true"
 # Findings to process per batch
 DAEMON_BATCH_SIZE="10"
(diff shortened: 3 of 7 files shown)


services/daemon/poller.py
```diff
@@ -595,7 +595,9 @@ async def _enqueue_finding(self, finding: Dict[str, Any], source: str) -> bool:
 
         try:
             from core.ingestion.ingestion_service import IngestionService
-            IngestionService().ingest_finding(finding)
+            # Synchronous SQLAlchemy — off the event loop, otherwise this
+            # fallback blocks the polling loop for the whole write.
+            await asyncio.to_thread(IngestionService().ingest_finding, finding)
```
nestor-deeptempo, line 600: `IngestionService()` is evaluated on the event loop, before `to_thread` is ever called — arguments are built in the caller's frame. And its `__init__` is not cheap:

```python
# core/ingestion/ingestion_service.py:88
def __init__(self):
    ...
    db_manager = get_db_manager()
    if db_manager.health_check():   # session_scope() + SELECT 1
```

`health_check()` opens a session and runs `SELECT 1` synchronously. So this line still makes a blocking DB round-trip on the loop for every finding — the thing the comment two lines up says it doesn't. The write moved off; the connect-and-probe didn't.

Simplest fix is to pull the construction into the thread too:

```python
await asyncio.to_thread(lambda: IngestionService().ingest_finding(finding))
```

Though building one `IngestionService` in `_init_services()` and reusing it would be better still — right now a fresh one is constructed per finding, which means a `SELECT 1` per finding even once this is inside the worker thread.

nestor-deeptempo (changes requested):
hey @bluejazz822, thanks for the contribution! I just have one small suggestion before merging, please take a look and lmk!

PR #725, opened 2026-08-26
fix(response): make should_auto_approve's threshold parameter effective

bluejazz822:
`ApprovalService.should_auto_approve` checks `confidence >= threshold` and then falls through to `confidence >= 0.85`, so the second check answers for every threshold above 0.85:

```python
service.should_auto_approve({"confidence": 0.86}, threshold=0.99)  # True
```

An operator raising the bar to be more conservative gets no error, no warning and no change in behaviour. For a knob whose only purpose is to make the service stricter, that is the dangerous direction to fail in.

## The 0.85 is not the bug

It is the bottom of a flag band that `needs_flag` and `get_action_decision` also draw at 0.85 — actions there are meant to run and be reviewed afterwards, and `test_auto_approve_with_flag` documents exactly that. So this derives the band from `threshold` instead of repeating the literal in three places:

- `FLAG_MARGIN` (0.05) and `MONITOR_ONLY_BELOW` (0.70) name the two lines.
- `_auto_approve_floor()` returns `threshold - FLAG_MARGIN`, clamped so it never drops below the monitor line. Without the clamp a low threshold makes `get_action_decision` return `monitor_only` for an action `should_auto_approve` has already cleared.
- `needs_flag` takes the same `threshold`, defaulted, so existing callers are unaffected.

## Compatibility

**Behaviour at the default threshold is unchanged.** The existing `TestConfidenceThresholds` cases pass unmodified:

| confidence | before | after |
|---|---|---|
| 0.95 | approve | approve |
| 0.87 | approve (flagged) | approve (flagged) |
| 0.85 | approve | approve |
| 0.84 | manual | manual |

`should_auto_approve`, `needs_flag` and `get_action_decision` have no callers outside tests today, so nothing downstream changes shape.

## Tests

`tests/unit/response/test_auto_approve_threshold.py` covers the moved band, the clamp, and a matrix asserting the three methods cannot disagree about the same action — that matrix is what caught the clamp being necessary.

### DIFF
--- a/core/response/approval_service.py
+++ b/core/response/approval_service.py
@@ -116,6 +116,14 @@ def _row_to_pending(row: ApprovalActionRow) -> PendingAction:
 class ApprovalService:
     """Service for managing approval workflow for autonomous actions."""
 
+    #: How far below the auto-approval threshold an action still runs, flagged
+    #: for review. One constant so the threshold, the flag band and
+    #: :meth:`get_action_decision` cannot drift apart.
+    FLAG_MARGIN = 0.05
+
+    #: Below this, an action is watched rather than proposed at all.
+    MONITOR_ONLY_BELOW = 0.70
+
     def __init__(self, data_dir: Optional[Path] = None, dry_run: bool = False):
         """
         Initialize approval service.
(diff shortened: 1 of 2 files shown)


nestor-deeptempo (changes requested):
@bluejazz822 thanks for your contribution! overall i agree with the direction, there's just a couple of gaps i want addressed before we merged

## 1. Blocking: `create_action` is the gate that actually fires

`core/response/approval_service.py:302` makes the auto-approval decision that runs in production:

```python
requires_approval = confidence < 0.90   # hard-coded, no threshold parameter
```

It sets `status = APPROVED`, and `autonomous_response_service.create_isolation_action` branches on that status to execute the isolation. Also reached from `core/response/checkpoints.py` and `core/agents/tool_registry.py`. Nothing passes it a threshold — `daemon_confidence_threshold` only reaches `services/daemon/responder.py:144`, which does its own comparison. So after this merges, an operator raising the threshold to 0.99 still gets host isolations auto-approved at 0.90, silently.

Note this is also where `TestTheThreeAgree` breaks down: at the default threshold `should_auto_approve({"confidence": 0.87})` returns `True` while `create_action(confidence=0.87)` marks the action `PENDING`.

**Proposal — one threshold on the service, every method defaults to it:**

```python
class ApprovalService:
    FLAG_MARGIN = 0.05
    MONITOR_ONLY_BELOW = 0.70

    def __init__(self, data_dir=None, dry_run=False, threshold: Optional[float] = None):
        ...
        self.threshold = (
            threshold if threshold is not None
            else get_settings().daemon_confidence_threshold
        )
```

then in `create_action`, replace the literal with the method that already handles `force_manual_approval`:

```python
requires_approval = not self.should_auto_approve(
    {"confidence": confidence}, threshold=threshold
)
```

`daemon_confidence_threshold` already defaults to 0.90, so nothing changes for anyone who hasn't configured it — but the knob now reaches every path, including agent-initiated actions. (If reusing a daemon-named setting here feels wrong, adding `approval_confidence_threshold` to `core/config.py` works too; it just costs an `env.example` entry for the ratchet test.)

## 2. The band should move above the line, not below it

`threshold=0.99` still auto-approves at 0.94 under this patch. The parameter is effective but still doesn't mean what its name says, and there's no config that exposes the extra 5 points of looseness.

I'd resolve it the other way: `create_action` — the code that actually isolates hosts — is already strict at `>= 0.90`, and so is `responder.py:144`. Those are the shipped semantics. `test_auto_approve_with_flag` was written alongside the buggy fall-through, so I don't think it's evidence of intent so much as a description of the bug. Making the helpers strict aligns them with production; making production loose would mean host isolation at 0.85 for an operator who asked for 0.90.

**Proposal — the threshold is the execute line, `FLAG_MARGIN` only widens *review*:**

```python
    def should_auto_approve(self, action, threshold=None, force_manual=False) -> bool:
        if force_manual or self.get_force_manual_approval():
            return False
        threshold = self.threshold if threshold is None else threshold
        return action.get("confidence", 0.0) >= threshold

    def needs_flag(self, confidence: float, threshold: Optional[float] = None) -> bool:
        """Cleared the line, but only just — worth a look after the fact."""
        threshold = self.threshold if threshold is None else threshold
        return threshold <= confidence < threshold + self.FLAG_MARGIN

    def get_action_decision(self, action, threshold=None) -> str:
        threshold = self.threshold if threshold is None else threshold
        confidence = action.get("confidence", 0.0)
        if confidence < min(self.MONITOR_ONLY_BELOW, threshold):
            return "monitor_only"
        if confidence < threshold:
            return "manual_approval"
        return "auto_approve"
```

This drops `_auto_approve_floor` and with it the clamp. Worth noting what the clamp cost: below `threshold = 0.75` the floor pinned at 0.70, so lowering the knob further did nothing — silently, the same failure shape you're fixing — and the `manual_approval` band went empty, so lowering the threshold *removed* human review. The `min()` above keeps the three methods consistent at any threshold with no dead zone.

This does mean revising the new tests — the 0.85/0.87 auto-approve cases become `manual_approval`, and `needs_flag` moves to `[0.90, 0.95)`. Worth it, I think: the test file then documents the same rule the production gate enforces. The parts I'd keep as-is are `TestTheThreeAgree` (the matrix is the valuable bit and still passes) and `TestForceManualStillWins`. `test_auto_approve_with_flag` in `test_approval_workflow.py` needs its confidence bumped to 0.92 or its threshold dropped to 0.85.

If you'd rather not touch the band in this PR, the fallback is: keep your loose band but leave `create_action` strict at `>= threshold`, and say in the `should_auto_approve` docstring that it's advisory and not the gate. Less churn, but two lines in one class.

## 3. Follow-up: the literals next door

`core/response/autonomous_response_service.py` repeats the same numbers four times — 285-289, 396, and the live `if confidence >= 0.85 and auto_execute:` at 541. Once the above lands, those become:

```python
# _get_recommendation
decision = self.approval_service.get_action_decision({"confidence": confidence})

# process_finding
if auto_execute and self.approval_service.should_auto_approve({"confidence": confidence}):
```

Fine as a separate PR, but it's the other half of the "cannot drift apart" claim, so let's file it rather than leave it implied.

---

PR #744, opened 2026-08-27
Run integrations locally + fix self-hosted MCP servers

G-r-ay:
Makes the SIEM/threat-intel integrations verifiable locally and fixes the bugs that kept self-hosted MCP servers from ever connecting or being usable by the assistant.

## Local integrations (opt-in docker-compose profiles)
- **Elasticsearch** (`--profile elastic`): single-node ES + Kibana, security disabled for local use; mirrors the `splunk` profile. Gives the hunt a local `telemetry_search` provider.
- **MISP** (`--profile misp`): official misp-core/modules + own mariadb/valkey on `https://localhost:8443`; fixes a Redis host collision with Vigil's own redis. Gives `indicator_lookup` a local provider.

## MCP server fixes
- **sys.path shim** on 13 self-hosted `tool.py` servers: spawned with a narrowed env (no `PYTHONPATH`), their `core.*` imports crashed (top-level) or silently no-op'd (lazy). Only splunk/elastic had the fix. All 17 self-hosted tools now start cleanly under the real spawn.
- **uvx integrations spawnable**: `uv` added to `setup_dev.sh` (8 servers launch via `uvx`); pinned `mcp<2` for the 3 GCP servers that crashed on the removed `mcp.server.fastmcp` API.
- **Disabled servers no longer logged as failures**: startup dialled every server and reported each disabled one as "Failed to connect" — dozens of alarming lines on a normal boot. Now skipped.

## Chat / LLM
- **Vertex as a chat provider**: added to backend validation + the provider dialog; fixes the chat 503 on a Vertex-only deployment.
- **Assistant can use connected integrations**: chat declared tools from a static list, so VirusTotal/OTX/MISP/… were never callable. It now merges the live MCP registry (`live_mcp_tools`) into the tool surface — an integration is usable the moment its server connects, always offered regardless of a per-agent tool list. Verified: the assistant answers IP-reputation queries from VirusTotal and OTX.
- **Per-turn registry refresh**: the agent/workflow tool registry syncs to live connection state instead of only at boot.

## Notes
- Integrations stay behind docker-compose profiles — they never auto-start.
- Known open items (not in this PR): `cribl-stream` (uvx resolve fails), `joe-sandbox` (stale submodule path), `mempalace` (uninitialized submodule), and the hunt still binds `indicator_lookup` to the local DB rather than VT/OTX.


core/llm/chat_layers.py
```diff
@@ -29,9 +29,21 @@ def run_id_for(session_id: str) -> str:
...
+    # Two surfaces. Static built-ins are curated: a per-agent recommended-tools
+    # list (``wanted``) narrows them. Connected MCP integrations (server-prefixed,
+    # e.g. virustotal_get_ip_report) are ALWAYS offered when present — a user who
+    # connected an integration expects the assistant to use it regardless of any
+    # agent's tool list, and hunts curate separately (playbook_resolver). So a
+    # ``wanted`` list filters only the built-ins; live integrations are appended.
+    static = {t["name"]: t for t in ALL_TOOLS if t.get("name")}
+    mcp = {t["name"]: t for t in (mcp_tools or []) if t.get("name")}
+    static_names = list(static) if wanted is None else [n for n in wanted if n in static]
+    names = static_names + [n for n in mcp if n not in static_names]
```
nestor-deeptempo: would this open up all 180+ tools? just want to make sure just the tools users enable are available (apart from the default ones)

just what claude recommended for a fix but i'm good to hear your thoughts
`tools = [t for t in registry.get_all_tools() if _server_of(t) not in MCPService.DEFAULT_ENABLED]`

core/llm/chat_layers.py
nestor-deeptempo, line 127: some code cleanup

This guard no longer fires for MCP tools. get_all_tools prefixes every description with f"[{server_name}] ", so a description-less tool arrives as "[virustotal]" — truthy after .strip(). Strip the prefix before the check.

core/llm/chat_layers.py
```diff
@@ -49,12 +61,16 @@ def _declare(wanted: Optional[List[str]]) -> List[Dict[str, Any]]:
...
+) -> str:
     document = {
         "model": model,
         "budgets": DEFAULT_BUDGETS,
         "runtime": DEFAULT_RUNTIME,
-        "tools": _declare(tools),
+        "tools": _declare(tools, mcp_tools),
         # A chat tool asks the person directly rather than parking on a
         # checkpoint: they are already in the conversation.
         "approvals": [],
```
nestor-deeptempo, line 152: i think with all tools being passed through now this empty approval gate means none of them need approval

i would either add the explosive ones or filter the incoming tools or something

services/api/routers/llm_providers.py
```diff
@@ -34,7 +34,7 @@
     auth=Auth.REQUIRED,
 )
 
-VALID_PROVIDER_TYPES = {"anthropic", "openai", "ollama"}
+VALID_PROVIDER_TYPES = {"anthropic", "openai", "ollama", "vertex"}
```
nestor-deeptempo, line 47: i would test via the ui one time to make sure this isn't a problem

This is necessary but not sufficient. _probe ends in raise RuntimeError(f"unsupported provider_type: {provider_type}"), and the wizard gates Continue on a successful test — so Vertex can never be saved through the UI, only via curl. Also vertex isn't in _TIER_HEURISTIC/_PRICED_PROVIDERS, so its usage prices at $0 with pricing_source: unknown, which quietly defeats ORCHESTRATOR_MAX_COST and the budget service. And _sync_models_for returns None for it, so no models reach Bifrost's allowlist.

mcp-config.json
```diff
@@ -306,7 +310,13 @@
     "misp": {
       "command": "python3",
       "args": ["core/integrations/misp/tool.py"],
-      "cwd": "${workspaceFolder}"
+      "cwd": "${workspaceFolder}",
+      "env": {
+        "_setup_note": "MISP via REST API. Set MISP_URL and MISP_API_KEY (generate an Auth Key in the MISP UI) in .env; MISP_VERIFY_SSL=false for a local self-signed cert. The server stays dormant until MISP_API_KEY is set.",
+        "MISP_URL": "${MISP_URL}",
+        "MISP_API_KEY": "${MISP_API_KEY}",
+        "MISP_VERIFY_SSL": "${MISP_VERIFY_SSL}"
```
nestor-deeptempo: `extract_required_env_vars` treats every `${VAR}` here as a required credential, and `connect_to_server` won't spawn the server if any is unset — so this makes `MISP_VERIFY_SSL` mandatory even though the `_setup_note` above calls it situational and the tool treats unset as "verify" (`misp/tool.py:56`). Net effect: someone sets `MISP_URL` + `MISP_API_KEY` as documented and MISP still comes up dormant.

Same thing your Elastic commit fixed by removing passthroughs. And none of the three is needed — the tool reads via `resolve()` → `get_secret()`, which falls through encrypted store → env → `.env` on its own. Keeping just `MISP_API_KEY` would give you one intentional gate and make the note accurate.

infra/docker/docker-compose.yml
```diff
@@ -502,6 +502,131 @@ services:
...
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.15.0
+    container_name: deeptempo-elasticsearch
+    hostname: elasticsearch
+    environment:
+      - discovery.type=single-node
+      - xpack.security.enabled=false
+      - ES_JAVA_OPTS=-Xms1g -Xmx1g
+    ports:
+      - "9200:9200"   # REST API
```
nestor-deeptempo: we just consolidated this in a pr a day or two ago [#708](https://github.com/Vigil-SOC/vigil/pull/708), could you make it match the rest?

Can we make this 127.0.0.1:9200:9200? Same for 5601 (line 540) and 8443 (line 615) — MISP boots with admin@admin.test / admin.

shutdown_all.sh
```diff
@@ -30,9 +31,11 @@ pkill -f 'services\.worker' 2>/dev/null || true
 pkill -f "vite.*opensoc" 2>/dev/null || true
 pkill -f "mcp_servers.*_server" 2>/dev/null || true
 
-# Kill by port
+# Kill by port (6989 agent serve, 6990 agent worker health)
 lsof -ti:6987 | xargs kill -9 2>/dev/null || true
 lsof -ti:6988 | xargs kill -9 2>/dev/null || true
+lsof -ti:6989 | xargs kill -9 2>/dev/null || true
+lsof -ti:6990 | xargs kill -9 2>/dev/null || true
```
nestor-deeptempo: there's an existing problem that both the agent worker + splunk ui have the same port, so i would remove this kill command in case splunk is running for a user and it has that port running

just a heads up, not a huge deal atm since it's existing

nestor-deeptempo (changes requested):

core/llm/chat_layers.py
G-r-ay: Disabled servers never connect (client.py:221), so their tools never enter tools_cache. get_all_tools() only returns enabled+connected servers, so this surfaces defaults and what the user enabled not the whole 180+

core/llm/chat_layers.py
G-r-ay: fixed it in get_all_tools instead: a description-less tool now gets "" rather than a fabricated "[server] ", so the empty-description guard in _declare fires again (and real descriptions keep their [server] prefix).

core/llm/chat_layers.py
G-r-ay: good point though adding them to approvals would actually just hang the call, since chat has no way to resume from a park. so instead i filtered the destructive ones (isolate/block/kill/etc) out of the chat tool surface entirely containment still goes through the approval queue, chat just can't fire them directly anymore.

shutdown_all.sh
G-r-ay: removed those

services/api/routers/llm_providers.py
G-r-ay: All three are already addressed on the current branch: Vertex pricing was fixed deliberately in 421dd55 (it's in _TIER_HEURISTIC/_PRICED_PROVIDERS now), the model-sync gap was closed by the #640 key-sync rework (vertex is _UNMANAGED → static allow-list), and the save-via-UI block no longer bites because the console moved to the Bifrost-native path and the setup step advances on vertex's unknown verdict — the _probe RuntimeError only survives in the now-orphaned LlmProviderDialog, which nothing renders.

shutdown_all.sh
G-r-ay: removed those

infra/docker/docker-compose.yml
G-r-ay: 9200, 5601, and 8443 all bound to 127.0.0.1 now

nestor-deeptempo (approved):
🫡

PR #750, opened 2026-08-28
fix(decisions): land the AI Decisions badge on the tab it counted

### LINKED ISSUE #746: AI Decisions badge counts pending approvals but the screen opens the feedback tab
The nav-rail badge on **AI Decisions** counts one dataset and the screen opens a different one, so a non-zero badge routinely lands the operator on an empty page.

On a development database with 93 pending approvals:

- `SocConsole.tsx:174` sets the badge from `usePendingApprovals().actions.length` — i.e. `GET /api/approvals/pending`. The rail renders `93`.
- `go('decisions')` navigates to `/decisions` with no search string (`SocConsole.tsx:315`, `go` at `:221`).
- `DecisionsScreen.tsx:616` opens on `useState<DecTab>('pending')`, whose rows come from the **feedback** queue, not approvals.

With no decisions awaiting feedback, the screen reads "No decisions awaiting feedback." while the thing the badge counted sits one tab over, under **Pending Approvals**, correctly labelled `(93)`.

This is the same operator-trust failure #675 described from the other end: there, the badge read a constant number because rows never expired; here, it reads a number that leads nowhere. Expiry (`04a7327`) fixed the stale-count half. The routing half is untouched.

## Scope

The badge should land on the tab it counted. `go` already carries a `search` option and both `SettingsScreen` (`?section=`) and `CasesScreen` (`?case=`) derive state from the URL, so the shape is established:

- `DecisionsScreen` derives `tab` from a `?tab=` param with a fallback for missing/unknown values (the `SettingsScreen.tsx:54-58` pattern), keeping the URL as the source of truth so the tab is deep-linkable and survives a reload.
- The rail passes `?tab=approvals` for `decisions` when `parked > 0`, and navigates as it does today when the badge is absent.

Not in scope: changing what the badge counts, or merging the feedback and approval queues. Whether the *default* tab should become `approvals` whenever approvals are pending and feedback rows are not is a product call — the deep link fixes the reported symptom without needing it.

## Verification

`SocConsole.test.tsx` already asserts the badge's accessible name (`AI Decisions (2 waiting)`, `:372`) and clicks through to the screen (`:282`, `:413`), so the regression test belongs there: clicking the badged nav item shows the approvals tab, and clicking it unbadged still shows Pending.

craig-dt:
Closes #746.

## What changed

The AI Decisions rail badge counts pending approvals, but `go('decisions')` carried no tab and `DecisionsScreen` opened on the feedback queue. With 93 approvals parked and no feedback rows, the rail read `93` and the click landed on "No decisions awaiting feedback" — the counted questions were one tab over.

The open tab now lives in `?tab=`, derived from the URL with a fallback to feedback for a missing or unknown value, following the `SettingsScreen` `?section=` pattern. Tabs become deep-linkable as a side effect. The rail passes `?tab=approvals` only when it is showing a count, so an unbadged click still opens the feedback queue — which is what the screen is for when nothing is parked.

## Why the second commit exists

`/code-review` found five issues in the first one; all are fixed in `fa0cf84`.

The one that mattered: `if (selected)` returns before the tab list renders, so a badged click **while a decision detail was open** showed neither the approvals queue nor the decision. It silently re-sourced the detail's rows from `all.rows` and replaced the open decision with "This decision is no longer in the current list." From a detail view the badge still looked dead, and it now also destroyed what was being read. A tab change closes the open decision.

`go`'s dedupe guard (`next === current && !search`) compared only the screen:

- a repeat badged click pushed a duplicate history entry;
- an unbadged click from `?tab=approvals` — the state an operator is left in after approving the last action — was swallowed entirely, parking them on an empty tab.

The guard now compares the query string too. One change, both paths, a test each.

Tab clicks `replace` instead of pushing: a tab is view state, not a navigation step, and pushing made Back walk tabs instead of leaving the screen. The existing params are copied rather than replaced wholesale.

`DecTab` is derived from `DEC_TABS` instead of hand-duplicating it — a fifth tab added to the union and forgotten in the array would have rendered as `pending` with no type error, because an unknown `?tab=` falls back silently.

## How it was tested

- Four new tests in `SocConsole.test.tsx`: badged click lands on approvals; unbadged click lands on feedback; a badged click leaves an open detail; the unbadged click moves off `?tab=approvals` when the badge is gone.
- `npm run test:run` — 75 passed (14 files). `tsc --noEmit` clean. `npm run lint` unchanged (the 2 errors pre-date this branch).
- Against the running console on `main`: reproduced the badge-93-lands-on-an-empty-page bug, then confirmed the fixed click goes to `?tab=approvals` with all 93 rows rendered.
- Back-button behaviour verified in the browser, since tests don't capture it: dashboard → AI Decisions → Analytics tab → one Back returns to the dashboard.


craig-dt:
Synced with `main` (fa38c67), which retired `src/redesign/` in #711 — so this branch's files moved:

- `src/redesign/SocConsole.tsx` → `src/shell/SocConsole.tsx`
- `src/redesign/SocConsole.test.tsx` → `src/shell/SocConsole.test.tsx`
- `src/redesign/screens/decisions/DecisionsScreen.tsx` → `src/screens/decisions/DecisionsScreen.tsx`

Git followed the renames and carried the edits over; two content conflicts, both trivial: the `styles.css` import path changed to `../styles.css` alongside my `useLocation` import, and `main` dropped `DecisionsScreen`'s header comment (it cited the deleted `DECISIONS_WIRING.md`), so I kept main's shape and added only the two imports this branch needs.

Re-verified post-merge on the new structure: `SocConsole.test.tsx` 20 passed, full suite 73 passed (13 files), `tsc --noEmit` clean, and ESLint is now at 0 errors — main fixed the two that pre-dated this branch.

PR #751, opened 2026-08-28
test(db): give DB-backed unit tests a throwaway database

### LINKED ISSUE #747: Approval tests write pending containment actions to the developer's real database with no cleanup
`tests/unit/response/test_approval_workflow.py::TestApprovalQueue` constructs a real `ApprovalService()` and calls `add_to_queue`, which persists to whatever `DATABASE_URL` points at. There is no teardown, no transaction rollback and no isolated database, so every local run of the suite permanently adds rows to the developer's dev database — and those rows are pending containment proposals.

Measured on my dev database this morning, `GET /api/approvals/pending` returns 93 actions, all `created_by: system`, all with `workflow_run_id: null`:

```
 action_type  |     target      | count
--------------+-----------------+-------
 isolate_host | workstation-042 |    31
 isolate_host | host1           |    31
 block_ip     | 1.2.3.4         |    31
```

Exactly the three targets `TestApprovalQueue` leaves behind per run — `test_add_to_queue` (`workstation-042`, `:224`) and `test_list_pending_approvals` (`host1` + `1.2.3.4`, `:242-243`) — multiplied by 31 runs. `test_approve_action` and `test_reject_action` add rows too but move them to a terminal status, so they don't accumulate as pending. Timestamps span 2026-07-24 to 2026-08-27 in batches of three.

This is the mechanism behind the row counts quoted in #675 (45 rows then, 93 now) — that issue treated them as given and fixed expiry; the rows should not be created in the first place.

## Why it matters beyond untidiness

- The queue it pollutes is the containment-approval queue. Test fixtures propose isolating a host and blocking an IP, and they sit in the operator's queue indistinguishable from a real proposal.
- The expiry sweep from `04a7327` only runs in the daemon's cleanup task, so a developer running `./start.sh` without `--daemon` never has them swept — they simply accumulate.
- `pytest.ini` markers don't help: `external_service` gates the tests out of CI's unit job, but the local full-suite run they're written for is exactly where the damage lands.

## Scope

Make these tests not write to a database a human uses. Options, in the order I'd argue them:

1. A conftest fixture that gives DB-touching tests their own database or a rolled-back transaction, so writes are discarded regardless of test hygiene. Strongest, and the only one that covers tests nobody has audited yet.
2. Per-class cleanup that deletes the `action_id`s a test created. Cheap, but relies on every future test author remembering.

There are more writers in this file than `TestApprovalQueue` — `TestAuditTrail::test_get_audit_trail` (`:322`) and `TestApprovalWorkflowIntegration` (`:447`) also persist, and `tests/unit/workflows/test_workflow_phase_approval.py:126` builds an `ApprovalService` too — so the audit of what writes should be part of the fix rather than assumed to be one class.

Also worth deciding: whether `tests/conftest.py` should refuse to run DB-writing tests against a `DATABASE_URL` that isn't a test database, which turns this class of accident into a loud failure instead of silent residue.

## Cleanup

Separately from the fix, the 93 existing rows need clearing on affected dev databases — a daemon cleanup pass with `DAEMON_APPROVAL_EXPIRY_DAYS` set will expire them via the #675 path.

craig-dt:
Closes #747.

## What changed

A unit test that wrote to the database wrote to whatever the environment pointed at — for a developer, the database their own console reads. `TestApprovalQueue` left three pending containment proposals behind per run (`isolate_host workstation-042`, `isolate_host host1`, `block_ip 1.2.3.4`) and 93 had accumulated between 24 Jul and 27 Aug, sitting in the operator's approval queue indistinguishable from a real proposal. A local suite run reproduced it: 93 → 96.

Per-test cleanup would fix the tests that exist. This provisions a throwaway `vigil_test_<pid>_<rand>` from the ORM models instead, retargets the `DatabaseManager` at it, and drops it afterwards — so a test that forgets to tidy up **cannot** reach anyone's data.

The fixture is autouse but lazy: only a test marked `external_service` requests it, so the 1630 unit tests that need no server still run without one.

## Judgment calls worth reviewing

**`external_service` is the sole trigger, deliberately.** The first commit also keyed on `database`, which broke CI — `tests/unit/storage/test_dsn_parsing.py` carries `pytest.mark.database`, parses strings, and touches no server, so all 25 of its tests ERROR'd in the no-service unit job (`-m "not external_service"`, no postgres service). `external_service` is what CI's DB-backed job selects on, so a test needing PostgreSQL must already carry it to run in CI at all — which makes it the one reliable signal.

**No skip when the server is unreachable.** The DB-backed CI job exists to run these; a fixture that skipped its way to green would retire that gate silently. A missing Postgres is an error here.

**`tests/integration/` is untouched.** It runs against a schema provisioned from `infra/database/init/`, including tables no ORM model covers (the agent-layer ledger), so `create_all` cannot stand in for it.

**The marker boundary was audited, not assumed.** 12 other files under `tests/unit` reference the DB manager; all 12 pass in CI's unit job, which has no postgres service, so they mock it. Every real writer carries `external_service`.

## Second commit — review fixes

`/code-review` found the CI break above plus four more, all fixed in `24872ef`:

- Maintenance DDL runs on the manager's own engine rather than a hand-built `create_engine`, so the platform DB proxy's host/port rewrite is honoured instead of bypassed.
- `refresh_if_stale()` is pinned for the session. It re-reads `DatabaseConfig()` when the secrets-file mtime changes, and would have retargeted silently back onto the developer's database mid-run — a running backend saving Settings is enough. Invisible in CI (no encrypted backend), so it would only ever have bitten the machine this protects.
- The teardown drop runs even if the retarget back to origin raises, and names the leaked database if the drop itself fails.
- The name carries a random suffix as well as the pid, which alone collides across PID namespaces — two containers sharing a server could pick the same name, and the second run's `DROP ... WITH (FORCE)` would take the first's live database mid-test.
- Extension creation tolerates failure per statement (matching `create_tables`), so an image without pgvector fails on the `CREATE TABLE` that names the real problem.

## Requirements this adds

The role running unit tests needs `CREATEDB` and enough privilege for `CREATE EXTENSION vector`. Both CI DB jobs already use a superuser `test` role against `pgvector/pgvector:pg16`.

## How it was tested

- New `tests/unit/storage/test_db_isolation.py`: DB-marked tests run against a `vigil_test_*` database, and it carries the ORM schema.
- `pytest tests/unit/ -m external_service` — 22 → 24 passed, same 5 skips and 2 xfails.
- `pytest tests/unit/ -m "not external_service"` — 1630 passed, unchanged from `main`.
- The CI-break case, against an unreachable server: `test_dsn_parsing.py` went 25 errors → 25 passed.
- The dev database stayed at 96 rows across a full DB-backed run (the pre-fix baseline took it 93 → 96), and no `vigil_test_*` database survives.

## Noted, not fixed here

`CLAUDE.md`'s env table presents `DATABASE_URL` as the PostgreSQL connection, but `DatabaseConfig` never reads it — it reads the encrypted `POSTGRESQL_CONNECTION_STRING`, then `POSTGRES_*`. Only the TypeScript agent layer and `scripts/migrate_schema.py` read `DATABASE_URL`. `nightly.yml` sets only `DATABASE_URL` at job level and runs `pytest tests/` against `postgres:16-alpine` (no pgvector), so its DB-backed tests are likely not hitting the CI database at all. Both belong in their own issue; this PR only corrects the paragraph it added.


craig-dt:
Synced with `main` (0d4cb00).

One resolution worth flagging: `main` deleted `CLAUDE.md` in #748 (product docs moved to vigilsoc.org), so the documentation paragraph this PR originally added has no home. I took the deletion and dropped the paragraph rather than relocating it — `CONTEXT.md` is a domain-vocabulary doc for `core/`, not a testing guide, and no repo-level testing doc survived. The fixture's own module docstring carries the explanation, which is where anyone opening `tests/unit/conftest.py` will read it.

Re-verified post-merge: `pytest tests/unit/ -m external_service` 24 passed, `pytest tests/unit/ -m "not external_service"` 1629 passed (main removed one test), no `vigil_test_*` database survives.

PR #770, opened 2026-08-31
fix(web): UserMenu no longer white-screens when full_name is null

### LINKED ISSUE #768: UserMenu white-screens the whole app when `full_name` is null (unguarded `.split()`)
## Summary

`UserMenu` computes avatar initials with `user.full_name.split(' ')`. It guards `!user` but not a null `full_name`, so when `full_name` is null the component throws during render, React unmounts the tree, and the whole app renders as a blank white page. The only clue is a `TypeError` in the browser console; the backend logs stay clean and 200-only.

`full_name` is nullable, so this is reachable by any user without one — not only the case below.

## Reproduction

Fresh database, `DEV_MODE=true`, no `admin` row:

```console
$ curl -s localhost:6987/api/auth/me | jq '{username, full_name}'
{
  "username": "dev-user",
  "full_name": null
}
```

Load any authenticated route (`/dashboard`, `/cases`). Blank white page; console:

```
TypeError: Cannot read properties of null (reading 'split')
    at lk (http://localhost:6987/assets/SocConsole-LunyK41E.js:36:17354)
    at Is (http://localhost:6987/assets/index-BwM3Eeqo.js:39:16961)   <-- React render
```

Confirmed against the published 0.5.0 image and reproduced in a clean browser session. Setting a non-null `full_name` and restarting the backend fixes it; the page renders and the avatar shows the initials.

## Cause

`clients/web/src/shell/UserMenu.tsx:48-55` (current `main`):

```ts
if (!user) return null          // guards user…

const initials = user.full_name
  .split(' ')                   // …but not user.full_name
  .map((n) => n[0])
  .join('')
  .toUpperCase()
  .slice(0, 2)
```

The trigger in the case above is `services/api/middleware/auth.py::_get_dev_user()`, which synthesises an in-memory `User` when no `admin` row exists and never sets `full_name`:

```python
_dev_user = User(
    user_id=str(uuid.uuid4()),
    username="dev-user",
    email="dev@localhost",
    password_hash="",
    role_id=admin_role.role_id if admin_role else str(uuid.uuid4()),
    is_active=True,
    mfa_enabled=False,
)   # no full_name
```

Note this path is reached far more often than it looks: `_get_dev_user` only finds a real user by the literal `username == "admin"`, and on a container-only deployment the `roles`/`users` tables can be empty (see the companion issue on compose seeding), so the synthesised user is the normal case rather than the exception.

`full_name` is also rendered directly at lines 75, 103 and 104. Those tolerate null (React renders nothing); only the `initials` computation throws.

## Suggested fix

The UI guard is the actual fix, since `full_name` is nullable in the schema:

```ts
const displayName = user.full_name?.trim() || user.username
const initials = displayName.split(/\s+/).filter(Boolean).map(n => n[0]).join('').toUpperCase().slice(0, 2)
```

Falling back to `username` also gives a sensible avatar and menu label instead of an empty one.

Worth doing alongside, but not instead:
- set `full_name` on the synthesised dev user (removes this trigger only);
- decide whether `full_name` should be nullable at all, or whether the API should coalesce it — that's a schema/contract call rather than a UI one.

An `ErrorBoundary` around the shell chrome would also stop one header widget from blanking the entire application. #447 was a different white-screen with the same blast radius, which suggests the boundary placement is worth revisiting on its own.

## Impact

Blank, unusable UI with no in-app error. Anyone whose account has no `full_name`, and every fresh `DEV_MODE` install without an `admin` row, hits this.

---

## Factory grooming notes

Fix `clients/web/src/shell/UserMenu.tsx` only: `full_name` can be null at runtime (`UserSchema.dump` + the in-memory `_get_dev_user` User) even though the DB column is NOT NULL. Fall back to `username` for initials and the menu labels. Do not change the schema, do not add a chrome ErrorBoundary (`SocConsole.tsx` already wraps only the screen), and do not treat setting `full_name` on the mock user as the fix.

samarmstrong:
Closes #768

`UserMenu` computed avatar initials with `user.full_name.split(' ')` after only guarding `user`. `full_name` is null at runtime (`UserSchema.dump` and the in-memory `_get_dev_user` User), so React threw during render and unmounted the whole console.

## Change

- Fall back to `username` for initials, the nav label, tooltip, and popover name
- Trim whitespace so a blank `full_name` uses the same fallback
- Guard the same crash if `username` is also missing (empty initials, no throw)

No schema change, no chrome ErrorBoundary, and the mock/dev user is left alone — those were explicitly out of scope.

## Validation

- `UserMenu` tests: named user still shows initials; `null` / `undefined` / `''` / `'   '` fall back to `username` without throwing, including the open menu
- `npm run test:run` — 83 passed
- `npx tsc --noEmit` — clean
- `npm run lint` — no new issues (existing warnings only)

<div><a href="https://cursor.com/agents/bc-172ba7ec-9115-4f01-be01-15014b16b312?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-172ba7ec-9115-4f01-be01-15014b16b312&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/clients/web/src/shell/UserMenu.test.tsx
+++ b/clients/web/src/shell/UserMenu.test.tsx
@@ -1,0 +1,48 @@
+import { describe, expect, it, vi } from 'vitest'
+import { fireEvent, render, screen } from '@testing-library/react'
+import { MemoryRouter } from 'react-router-dom'
+import UserMenu from './UserMenu'
+
+const auth = vi.hoisted(() => ({
+  user: {
+    username: 'dev-user',
+    full_name: 'Test User' as string | null,
+    email: 'dev@localhost',
+    role_id: 'role-admin',
+    mfa_enabled: false,
+  },
+  logout: vi.fn(),
+}))
+
+vi.mock('../contexts/AuthContext', () => ({
+  useAuth: () => auth,
+}))
+
+function renderMenu() {
+  return render(
+    <MemoryRouter>
+      <UserMenu />
+    </MemoryRouter>,
+  )
+}
+
+describe('UserMenu', () => {
+  it('renders initials and labels from full_name', () => {
+    auth.user.full_name = 'Test User'
+    renderMenu()
+    expect(screen.getByText('TU')).toBeInTheDocument()
+    expect(screen.getAllByText('Test User').length).toBeGreaterThan(0)
+  })
+
+  it('falls back to username when full_name is null', async () => {
+    auth.user.full_name = null
+    renderMenu()
+    expect(screen.getByRole('button', { name: 'Account menu' })).toBeInTheDocument()
+    expect(screen.getByText('D')).toBeInTheDocument()
+    expect(screen.getAllByText('dev-user').length).toBeGreaterThan(0)
+
+    fireEvent.click(screen.getByRole('button', { name: 'Account menu' }))
+    const menu = await screen.findByRole('menu', { name: 'Account' })
+    expect(menu).toHaveTextContent('dev-user')
+  })
+})
--- a/clients/web/src/shell/UserMenu.test.tsx
+++ b/clients/web/src/shell/UserMenu.test.tsx
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { fireEvent, render, screen } from '@testing-library/react'
 import { MemoryRouter } from 'react-router-dom'
 import UserMenu from './UserMenu'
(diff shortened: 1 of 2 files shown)


samarmstrong:

PR #771, opened 2026-09-01
feat(agents): switch 13 SOC agents from MemPalace to read-only recall_entity

### LINKED ISSUE #735: The Python agents use the memory tool instead of MemPalace
## Parent

#727

## What to build

The 13 Python agents read episodic memory through the new tool instead of MemPalace.

Register the memory tool for them, and rewrite the prompt block and the per-agent principles that currently instruct agents to store and search their own IOCs, attributions, conclusions and false-positive reasoning in MemPalace.

**The direction changes.** Today agents write their own notes into MemPalace and search them back as prose. Under this they only read — a triage agent seeing that an address was ruled a false positive three times is close to the highest-value thing in this project, but its own output reaches memory through the **Distil** at a run's terminal, not by writing mid-run. A run that dies after writing would otherwise leave a Verdict asserted that the run never reached.

Chat keeps read access and never writes. An analyst thinking aloud is not a Verdict, and letting speculation reach memory is the promotion of prose to fact that ADR 0015 forbids.

**Compare before switching.** Same entity keys through both stores, count what each returns. Without it we would be swapping a store we cannot measure for one we have not measured — and it is the cheapest possible eval.

This does not remove MemPalace. Retiring it is separate work, ordered after this: pull the store before the replacement is proven and the agents lose memory with nothing in its place.

## Acceptance criteria

- [ ] The memory tool is available to the Python agents
- [ ] The prompt block instructs agents to read memory, not to write it
- [ ] Per-agent principles referencing MemPalace storage are rewritten
- [ ] A triage agent asking about an entity with false-positive history receives it
- [ ] Chat can read memory and has no path to write it
- [ ] Recall of known prior sightings is compared across both stores and the result recorded

## Blocked by

- #732

joshuacox:
fixes #735 

- Replace _MEMORY_PALACE_BLOCK and _memory_palace_section with read-only _MEMORY_BLOCK and _memory_section in core/agents/prompts.py, gating prompt inclusion on recall_entity presence in ALL_TOOLS.
- Add 'recall_entity' to recommended_tools for all 13 built-in SOC agents in core/agents/builtins.py.
- Rewrite all 13 built-in agent 'Memory:' principles in extra_principles to uniform read-only directives ('do not write to memory').
- Add 'recall_entity' to all workflow phases across incident-response, full-investigation, forensic-analysis, and cloud-incident WORKFLOW.md definitions.
- Add test_memory_grant.py covering tool grants, prompt honesty gating, read-only memory directives, and workflow phase definitions.

[plan](https://gist.github.com/joshuacox/6a13cef1954bdf6c671245968f25e4aa) created by antigravity + opus 

executed by gemini 3.7 flash


joshuacox:
ok one human fix (ws), and black did a bit of reformatting

nestor-deeptempo (changes requested):
no problems as of yet, just want to park this as i'm working on the epic and want to make sure this isn't merged earlier than it should. i'll review today once i get to this point

joshuacox:
let me know if I can modify

nestor-deeptempo (approved):
overall looks good! i'm working on the epic and i'm going to merge it in to the epic branch for further testing before main, thanks @joshuacox

PR #772, opened 2026-09-01
fix #769 ensure full db schema and reference data seeding on compose

### LINKED ISSUE #769: Compose deployments get a half-seeded database: initdb aborts at 05, files 06-21 never run
## Summary

`infra/docker/docker-compose.yml` mounts the schema SQL into `/docker-entrypoint-initdb.d`, which Postgres runs **before** the backend's `create_all`. Several of those files touch tables only `create_all` builds, so the pass fails early — and because the Postgres entrypoint uses `ON_ERROR_STOP`, **every file after the first failure never runs**.

Observed on a fresh volume: only `01`, `04` and `05` executed. Files **06–21 did not run at all**.

The host path is unaffected because `./start.sh` runs `scripts/seed_reference_data.py` as a second pass. That script's own docstring describes exactly this fix — "Re-applying the same files here, after `init_schema.py`, gets the roles…" — but `scripts/` is **not copied into the backend image**, so a container-only deployment has no way to run it, and the compose file has no equivalent step.

## Reproduction

Fresh volume, then `docker compose up -d`:

```console
$ docker compose logs postgres | grep -oE "running /docker-entrypoint-initdb.d/[0-9_a-z]+\.sql"
01_init_schema.sql
04_config_tables.sql
05_case_management_extended.sql
                              # <-- stops here; 06-21 never run

$ docker compose logs postgres | grep ERROR | head -3
psql:/docker-entrypoint-initdb.d/05_case_management_extended.sql:25: ERROR:  relation "sla_policies" does not exist
ERROR:  relation "ai_model_configs" does not exist
ERROR:  relation "agent_run_leases" does not exist
```

## Consequences

None of these name the real cause, which makes it expensive to diagnose:

- **`roles` is empty** — `06_auth_tables.sql` seeds it. So `POST /api/auth/bootstrap` cannot assign `role-admin`, and with `DEV_MODE=true` the backend's `_get_dev_user()` finds no `admin` row and synthesises one with a null `full_name`, which **white-screens the SPA** (companion issue filed separately).
- **`agent_run_leases` missing** — `agent-worker` logs `sweep failed: relation "agent_run_leases" does not exist` every 30 seconds, indefinitely. Its healthcheck stays green throughout, because that probes a side port.
- `ai_model_configs` and `sla_policies` likewise absent or unseeded.

Verified fix: replaying all 20 files after `create_all` populates `roles` (5 rows incl. `role-admin`), creates the missing tables, and silences the `agent-worker` sweep errors. Every `INSERT` across those files is `ON CONFLICT`-guarded, so replay is idempotent.

## Suggested fix

Add a second seed pass to `infra/docker/docker-compose.yml`. `clients/desktop/standalone/docker-compose.yml` already carries one (`db-seed`) that solves precisely this, including the subtlety that a healthy backend is not a sufficient signal — `/api/health` answers while `create_all` is still running, so it waits on a `create_all`-only table (`sla_policies`) as a sentinel instead.

The Helm chart's `db-init-job.yaml` is the better model: it records applied files in a `_vigil_schema_versions` table and handles per-file failure, so a partial run self-heals on the next upgrade rather than leaving a half-built database. Borrowing that versioning into the compose path would fix the ordering bug and make it diagnosable.

Options worth weighing (an engineering call, not obviously one-way):
- a `db-seed` service, as the desktop compose does — smallest change, but requires an extra command after `up`;
- copying `scripts/` into the image so a container can run `seed_reference_data.py` itself;
- making the data-only SQL tolerant of a missing table so the first pass no longer aborts.

At minimum, the failure should be loud: a half-seeded database currently produces a stack of healthy containers.

Related: #567 (`seed_default_roles` failing on empty-roles databases) and #568 (`seed_default_admin`) cover the same seeding territory from the `migrate_schema.py` side.

## Impact

Any compose-based deployment that isn't launched via `./start.sh` gets a silently half-initialised database, presenting as a blank UI and a permanently erroring agent worker.

---

joshuacox:
- Add infra/docker/initdb/00_apply.sh to execute all SQL files in /db-init with ON_ERROR_STOP=0 during postgres initdb and db-seed passes.
- Update postgres service volume mounts in infra/docker/docker-compose.yml to mount SQL files to /db-init:ro and 00_apply.sh into initdb.d.
- Add db-seed service to infra/docker/docker-compose.yml that waits for the sla_policies sentinel table after backend create_all and replays all SQL files to seed roles, case templates, and non-ORM tables.
fixes #769

### DIFF
--- a/infra/docker/docker-compose.yml
+++ b/infra/docker/docker-compose.yml
@@ -49,9 +49,9 @@ services:
       - "127.0.0.1:5432:5432"
     volumes:
       - postgres_data:/var/lib/postgresql/data
-      # Path is relative to this compose file (infra/docker/), so ../ points
-      # at infra/, where the schema SQL now lives (infra/database/init/).
-      - ../database/init:/docker-entrypoint-initdb.d
+      # SQL files at /db-init (not in initdb.d directly — see 00_apply.sh).
+      - ../database/init:/db-init:ro
+      - ./initdb/00_apply.sh:/docker-entrypoint-initdb.d/00_apply.sh:ro
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U deeptempo -d deeptempo_soc"]
       interval: 10s
@@ -118,6 +118,37 @@ services:
       bifrost:
         condition: service_healthy
     restart: unless-stopped
+    networks:
+      - deeptempo-network
+
+  # Second pass: replay init SQL after the backend's create_all has built
+  # ORM-managed tables. Mirrors clients/desktop/standalone/docker-compose.yml
+  # db-seed. /api/health answers while create_all may still be running, so
+  # the sentinel is an ORM-only table (sla_policies), not the health endpoint.
+  db-seed:
+    image: pgvector/pgvector:pg16
+    container_name: deeptempo-db-seed
+    depends_on:
+      backend:
+        condition: service_healthy
+    environment:
+      PGHOST: postgres
+      PGUSER: deeptempo
+      PGDATABASE: deeptempo_soc
+      PGPASSWORD: ${POSTGRES_PASSWORD:-deeptempo_secure_password_change_me}
+    volumes:
+      - ../database/init:/db-init:ro
+      - ./initdb/00_apply.sh:/apply.sh:ro
+    entrypoint:
+      - /bin/sh
+      - -c
+      - >
+        for i in $$(seq 1 60); do
+          [ "$$(psql -tAc "SELECT to_regclass('public.sla_policies') IS NOT NULL")" = t ] && break;
+          sleep 1;
+        done;
+        exec sh /apply.sh
+    restart: "no"
     networks:
       - deeptempo-network
 
--- a/infra/docker/initdb/00_apply.sh
+++ b/infra/docker/initdb/00_apply.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+# Apply every schema/seed file, tolerating failures.
+#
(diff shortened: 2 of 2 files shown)


samarmstrong:

The second seed pass this PR is for is already on `main`. #778 (`660079bfbdbb622dadda6405b27e2d91dd179e73`) closed #769 on 2026-09-01: `infra/docker/initdb/00_apply.sh` applies every `infra/database/init/*.sql` with `ON_ERROR_STOP=0`, and `infra/docker/docker-compose.yml` runs a profile-free `db-seed` on `up` that waits on `to_regclass('public.sla_policies')`. `tests/unit/_ratchets/test_compose_db_seed.py` locks that contract.

Merging `main` into `fixes/769` (`e62a2c3`) already brought that in. The only lines still unique to this PR are a second `db-seed` key in `infra/docker/docker-compose.yml` (the block that uses `pgvector/pgvector:pg16` and `depends_on: backend: condition: service_healthy`). Compose keeps the later key, so that block replaces the service #778 landed:

- The image no longer matches `postgres`, which is `postgres:16-alpine` because the schema only needs `uuid-ossp` and `pg_trgm`.
- Readiness is `service_healthy`. `/api/health` returns 200 while `create_all` is still running, which is why the existing service depends on `backend` with `service_started` and waits on `sla_policies`. The ratchet test asserts that condition.
- The postgres `service_healthy` dependency and the "applying anyway" path after the 60s wait are dropped.

`00_apply.sh` and the postgres `/db-init` mounts are not in this diff. They arrived with #778. Nothing left here is a change worth landing.

PR #777, opened 2026-09-01
fix(ci): start release images in smoke instead of importing packages

### LINKED ISSUE #767: release.yml smoke test cannot catch an unstartable image: it imports `services.api`, not `services.api.main`
## Summary

The release smoke test passes on a backend image that cannot start. It runs `import core, services.api`, but `services/api/__init__.py` is a docstring with no imports, so `main.py` — the module uvicorn actually loads — is never imported. Any failure at that module's import time is invisible to CI.

This is not hypothetical: **v0.5.0 shipped this way.** The smoke-test job existed at that tag and was green, while all three Python services in the published images crash-looped on startup.

## Evidence

`.github/workflows/release.yml`, "Backend — version check":

```yaml
docker run --rm -e DEV_MODE=true ... \
  ghcr.io/vigil-soc/vigil-backend:${{ needs.version.outputs.version }} \
  python -c "import core, services.api; print('backend ok')"
```

`services/api/__init__.py` is 84 bytes — a module docstring, no imports.

Run both against the actual published 0.5.0 image:

```console
$ docker run --rm -e DEV_MODE=true -e DATABASE_URL=postgresql://x:x@localhost/x \
    -e REDIS_URL=redis://localhost:6379 \
    ghcr.io/vigil-soc/vigil-backend:0.5.0 \
    python -c "import core, services.api; print('backend ok')"
backend ok                                    # <-- exactly what CI ran. Green.

$ docker run --rm -e DEV_MODE=true -e DATABASE_URL=postgresql://x:x@localhost/x \
    -e REDIS_URL=redis://localhost:6379 \
    ghcr.io/vigil-soc/vigil-backend:0.5.0 \
    python -c "import services.api.main"
  File "/app/core/integrations/mcp/service.py", line 215, in MCPService
    _STATE_WRITE_FILE = vigil_path("mcp_server_enabled.json", write=True)
  File "/app/core/config.py", line 30, in vigil_path
    (target.parent if parts else target).mkdir(parents=True, exist_ok=True)
PermissionError: [Errno 13] Permission denied: '/app/.vigil'
```

Adding four characters — `.main` — turns the green check red on a release that could not run.

The underlying image defect is **already fixed on main** by #701 (`useradd -d /home/vigil -m`, `ENV HOME=/home/vigil`), so this issue is only about the test's depth — the gate that should have stopped it.

## Why an import check isn't sufficient either

The daemon step has the same shape (`import services.daemon`), but deepening it to `services.daemon.main` would **not** have caught the daemon's failure: `services/daemon/main.py` imports the MCP client inside `_init_components()`, at runtime rather than import time. Only actually running the process reaches it.

## Suggested fix

Two layers, in order of value:

1. **Actually start each container and probe it.** Bring up the backend against a throwaway Postgres/Redis and require `GET /api/health` to answer 200; start the daemon and require its `/health` on 9091. This is the only check that catches runtime-path failures like the daemon's.
2. **Deepen the import checks** as a cheap backstop: `services.api.main` instead of `services.api`. Worth doing even with (1), since it fails faster and localises the error.

A regression test for the specific class of bug would also help: assert the image's state directory (`$HOME/.vigil`) is writable by the running user, since that is what broke and nothing asserts it.

## Impact

Every release is published on the strength of a check that cannot detect "the application does not start." v0.5.0 is still tagged Latest and remains unstartable for anyone pulling it.

---

samarmstrong:
## Why

Release smoke imported `services.api` and `services.daemon`. Those `__init__.py` files do not load the modules the images actually run (`services.api.main` for uvicorn, `services/daemon/main.py` for the daemon). v0.5.0’s smoke job was green while the published images crash-looped.

Deepening the import to `.main` would still miss the daemon: MCP client construction happens in `SOCDaemon._init_components()` at runtime, so only a live process reaches it.

## What

`smoke-test` in `.github/workflows/release.yml` now runs each image’s default command and fails if the process exits or the Dockerfile HEALTHCHECK URL does not answer:

- backend: `GET /api/health` on 6987
- daemon: `GET /health` on 9091

The import steps are gone, not kept as a second layer. HTTP 200 means the server accepted a connection, not that storage is healthy — `/api/health` always returns 200. No Postgres, Redis, or compose: `DEV_MODE=true` is already how this job runs, and if a container cannot listen without a dep that is the failure.

Agent `node --eval` of `worker.js` / `serve.js` is unchanged. GHCR push and `latest` tagging are unchanged; this job still only gates `update-release`.

## Verification

- YAML parse of `release.yml`
- `bash -n` on the new `run:` scripts
- actionlint 1.7.12 (same pin as CI)
- Independent review of the full diff

Could not run the job itself here: it pulls the just-built GHCR images, and this environment has no Docker.

Closes #767

<div><a href="https://cursor.com/agents/bc-a6b8fb06-9a9c-4887-90da-bb77cba8e918?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a6b8fb06-9a9c-4887-90da-bb77cba8e918&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -250,6 +250,12 @@ jobs:
   # ---------------------------------------------------------------------------
   # Smoke-test: pull the images and verify they start
   # ---------------------------------------------------------------------------
+  # Backend and daemon run their default command and must answer the image
+  # HEALTHCHECK URL. An import of the package __init__ never loaded the
+  # entry point, which is why v0.5.0's smoke was green while the published
+  # images crash-looped. Agent has no HEALTHCHECK (the port depends on the
+  # command), so it still loads worker.js / serve.js via node --eval.
+  # ---------------------------------------------------------------------------
   smoke-test:
     name: Smoke test images
     if: github.repository == 'Vigil-SOC/vigil'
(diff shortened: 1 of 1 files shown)


samarmstrong:

PR #778, opened 2026-09-01
fix(compose): re-apply init SQL after create_all on docker compose up

### LINKED ISSUE #769: Compose deployments get a half-seeded database: initdb aborts at 05, files 06-21 never run

samarmstrong:
## Why

`docker compose up` mounted `infra/database/init/*.sql` at `/docker-entrypoint-initdb.d`. Postgres runs those files with `ON_ERROR_STOP=1` **before** the backend’s `create_all`. `05_case_management_extended.sql` inserts into ORM-only tables (`sla_policies`, `case_templates`), so it errors and every later file is skipped.

A fresh volume therefore never runs `06_auth_tables.sql` (empty `roles`, no `role-admin`) or `21_agent_run_leases.sql` (worker sweep fails forever). Host `./start.sh` recovers via `scripts/seed_reference_data.py`; that script is not in the backend image, and this compose file has no wrapper that `compose run`s a seed profile.

## What

Copy the desktop standalone pattern onto the default `up` path:

- `infra/docker/initdb/00_apply.sh` applies every `infra/database/init/*.sql` with `ON_ERROR_STOP=0` (initdb first pass, so 06 and 21 still run when 05 fails).
- `db-seed` (no profile) waits on `to_regclass('public.sla_policies')`, not `/api/health`, then reapplies the same files. Existing half-seeded volumes skip initdb; this second pass still runs on every `up`.

Out of scope: Helm, `_vigil_schema_versions`, copying `scripts/` into `Dockerfile.backend`, a roles/healthcheck gate, #768/#767/#567/#568.

## Test plan

- [x] `pytest tests/unit/_ratchets/test_compose_db_seed.py tests/security/test_compose_port_binding.py` — compose contract (no seed profile, glob `*.sql`, wait on `sla_policies`, `restart: no`)
- [ ] Full-stack `docker compose up` was not run here (no Docker daemon in this environment). On a fresh volume, `roles` should contain `role-admin` and `agent_run_leases` should exist; on a volume that previously aborted at 05, `up` should still seed those.

Closes #769

<div><a href="https://cursor.com/agents/bc-abcdf0ab-9bc0-453a-83b4-d0149b8a56d5?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-abcdf0ab-9bc0-453a-83b4-d0149b8a56d5&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/infra/docker/docker-compose.yml
+++ b/infra/docker/docker-compose.yml
@@ -70,10 +70,10 @@ services:
   # Second pass over the same files, once the backend's create_all has built the
   # tables the data-only scripts need. On the default `up` (no profile): a seed
   # profile would never run — this compose file has no wrapper that `compose
-  # run`s it. A healthy backend only means /api/health answers and create_all
-  # may still be running, so wait for a table only it creates — its DDL is one
-  # transaction, so that sentinel implies the rest. Re-runs on every `up` so a
-  # volume that aborted at 05 still gets the rest of the files.
+  # run`s it. /api/health always returns 200, so wait for a table only create_all
+  # creates — its DDL is one transaction, so that sentinel implies the rest.
+  # Re-runs on every `up` so a volume that aborted at 05 still gets the rest of
+  # the files.
   db-seed:
     image: postgres:16-alpine
     container_name: deeptempo-db-seed
@@ -93,14 +93,18 @@ services:
           [ "$$(psql -tAc "SELECT to_regclass('public.sla_policies') IS NOT NULL")" = t ] && break;
           sleep 1;
         done;
+        if [ "$$(psql -tAc "SELECT to_regclass('public.sla_policies') IS NOT NULL")" != t ]; then
+          echo "sla_policies still missing after wait; applying anyway" >&2;
+        fi;
         exec sh /apply.sh
     depends_on:
       postgres:
         condition: service_healthy
-      # service_started, not service_healthy: /api/health answers "healthy"
-      # even when create_all is still running (or failed). The wait loop above
-      # is the real gate. Started implies postgres/redis/bifrost are already up,
-      # which a postgres-only depends_on would not.
+      # service_started, not service_healthy: /api/health always returns 200
+      # "healthy" (including on error), so compose treating the backend as
+      # ready is not a schema signal. The wait loop above is the real gate.
+      # Started implies postgres/redis/bifrost are already up, which a
+      # postgres-only depends_on would not.
       backend:
         condition: service_started
     restart: "no"
--- a/infra/docker/docker-compose.yml
+++ b/infra/docker/docker-compose.yml
@@ -51,13 +51,59 @@ services:
       - postgres_data:/var/lib/postgresql/data
       # Path is relative to this compose file (infra/docker/), so ../ points
       # at infra/, where the schema SQL now lives (infra/database/init/).
-      - ../database/init:/docker-entrypoint-initdb.d
+      # Must run here, not just in db-seed: the extensions have to exist before
+      # the backend's create_all (its indexes use pg_trgm's gin_trgm_ops), and
+      # the auth tables have to be created by this SQL — with the DEFAULT
+      # clauses create_all does not emit. 00_apply.sh explains why it isn't
+      # plain *.sql (Postgres initdb uses ON_ERROR_STOP=1 and would abort at 05).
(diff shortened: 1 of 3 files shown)


samarmstrong:

PR #779, opened 2026-09-01
fix(splunk): search all indexes and map all-time earliest_time

G-r-ay:
## Problem
Two bugs in `SplunkService`'s entity-search helpers (`search_by_ip/hash/username/hostname`), both hit in normal use — the Splunk MCP tool (`core/integrations/splunk/tool.py`) calls these with the all-time sentinel by default.

1. **Invalid all-time window.** Callers pass a large `hours` to mean "all time" (`_ALL_TIME_HOURS` = 876000). The helpers built `earliest_time=-876000h`, which Splunk rejects as `Invalid earliest_time` → HTTP 400. So any all-time entity search failed.
2. **Default-index-only search.** Queries had no `index=` clause, so they searched only the user's default indexes and silently missed data in any custom index.

## Fix
- Add `_earliest_from_hours()` — maps an absurd/non-positive window (≥ ~10y) to Splunk's all-time epoch `"0"`.
- Prefix the four entity searches with `search index=*` so they reach all indexes.

Independent of the Watcher / botsv3 work — this is a standalone bug fix.

### DIFF
--- a/core/integrations/splunk/client.py
+++ b/core/integrations/splunk/client.py
@@ -242,6 +242,19 @@ class SplunkService:
             logger.error(f"Error executing search: {e}")
             return None
 
+    @staticmethod
+    def _earliest_from_hours(hours: int) -> str:
+        """Turn a look-back window in hours into a valid Splunk earliest_time.
+
+        Callers pass a very large ``hours`` to mean "all time" (the tool's
+        all-time sentinel is 876000h). Splunk rejects such an offset as
+        ``Invalid earliest_time``, so map an absurd or non-positive window to
+        Splunk's all-time epoch ``"0"`` instead.
+        """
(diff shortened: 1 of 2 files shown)


nestor-deeptempo (changes requested):
overall good just one small change please

with this now we're searching through every index in all times right? i think we should still include an error so that any searches that don't find anything don't auto count as success, here's what i'm thinking but feel free to do something else

```
Fix, in tool.py for both splunk_search_ip and splunk_search_host:

results = splunk.search_by_ip(ip, args.get("hours") or _ALL_TIME_HOURS)
if results is None:
    return result({"error": "Splunk search failed or timed out", "ip": ip})
return result({"success": True, "ip": ip, "count": len(results), "results": results})
```

nestor-deeptempo (approved):

nestor-deeptempo (approved):

PR #780, opened 2026-09-01
feat(watcher): Slice 1 — alert to hypothesis, spec, launch

G-r-ay:
## What
Slice 1 of the Watcher (autonomous daemon decision core): turn one alert into the hypothesis + context a human types in the console today, then ride the existing hunt enqueue → worker → projection path. Net-new is one reasoning step; the rest is wiring.

- **`hypothesizer.py`** — enriched finding → `WatcherHypothesisSet` (hypotheses + narrative + techniques + data domains + scope). Gathers reputation (VT/Shodan off the finding) and builds scope from entities. **No cross-alert memory** — history recall is deliberately out of this slice.
- **`spec.py`** — maps the set to the exact `execute_workflow("threat-hunt", params)` call the console makes.
- **`launch.py`** — launches via `WorkflowsService.execute_workflow` and reads the run's projection back, so a Watcher hunt is indistinguishable from a console one on the agent layer.
- **`schemas.py`** — the `WatcherHypothesisSet` contract.

## Scope / guardrails
Stops at "alert → hypothesis + evidence gathered". No response actions, no cross-alert memory, existing daemon path untouched (this is additive). See `docs/WATCHER_BUILD_PLAN.md`.

## Tests
`tests/unit/daemon/watcher/` — 26 unit tests (fake gateway + fake workflows/reader), green on top of main.

### DIFF
--- a/docs/WATCHER_BUILD_PLAN.md
+++ b/docs/WATCHER_BUILD_PLAN.md
@@ -1,0 +1,221 @@
+# The Watcher — Build Plan (Slice 1: alert → evidence)
+
+> Companion to [WATCHER_SCOPING.md](WATCHER_SCOPING.md) (the *what* and *why*).
+> This is the *how*: an input→output build, one stage at a time, each with an
+> acceptance test and a running status log so it stays clear what was done and why.
+>
+> Branch: `feat/watcher-loop`.
+
+---
+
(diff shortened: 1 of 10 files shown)


G-r-ay:
Pulling this back for now.

PR #783, opened 2026-09-01
fix(llm): push custom OpenAI-compatible base_url to Bifrost

### LINKED ISSUE #586: Custom OpenAI-compatible provider base_url never reaches Bifrost (and the whole Bifrost sync is a no-op)
## Summary

A custom OpenAI-compatible provider (OpenRouter, LiteLLM, vLLM, Together) can be registered in the UI, passes the connection test, and still never routes: every completion fails with `no keys found that support model: <model>`. Bifrost stays pointed at `api.openai.com`.

Root cause is two stacked defects in `services/bifrost_admin.py`. The second one is broader than the title suggests — it silently disables **all** runtime sync to Bifrost, keys included.

Tested on `v0.4.0` (`8abecc2`) with `maximhq/bifrost:latest`.

## Defect A — `base_url` is never propagated

`bifrost_admin.py` syncs the key value (`push_provider_key`) and the model allow-list (`sync_provider_models`). Neither touches `network_config.base_url`.

`LLMProviderConfig.base_url` *is* read — `_do_sync_all_provider_models` puts `"base_url": row.base_url` into the row dict at `services/bifrost_admin.py:309` — but it is only used to feed model discovery and the connection test, both of which call the upstream **directly**. That's why the test passes while the gateway is misconfigured: the two paths never meet.

## Defect B — the whole sync is a no-op against current Bifrost

Bifrost moved keys out of the provider document into their own sub-resource:

```console
$ curl -s http://127.0.0.1:8080/api/providers/openai | python3 -c "import json,sys;print(list(json.load(sys.stdin)))"
['name', 'network_config', 'concurrency_and_buffer_size', 'proxy_config', 'send_back_raw_request', ...]
# note: no 'keys'

$ curl -s http://127.0.0.1:8080/api/providers/openai/keys
{"keys":[{"id":"912bfc2e-...","name":"openrouter-key","models":[...]}],"total":1}
```

Both `push_provider_key` and `sync_provider_models` do `keys = prov.get("keys") or []` against the provider document. That is now always empty, so both log a warning and `return False`. Visible on every backend start:

```
Bifrost: provider ollama has no keys slot to update
Bifrost: provider openai has no keys slot to update
```

So neither the API key nor the model allow-list has been reaching Bifrost at all. The only thing keeping installs working is the bootstrap `docker/bifrost/config.json`.

**Why tests didn't catch it:** `tests/unit/test_bifrost_admin.py`'s `_RecordingClient` returns the same payload for every GET regardless of URL, so a request to the provider document and one to the keys sub-resource are indistinguishable and the mock always supplies `keys`. The suite is green against a contract the gateway no longer has.

## Reproduction

1. Start the stack (`./start.sh`).
2. Register an OpenAI-compatible provider pointing at a custom endpoint:

```bash
curl -s -c cj.txt -o /dev/null http://127.0.0.1:6987/api/config/general
CSRF=$(grep csrf_token cj.txt | awk '{print $7}')
curl -X POST http://127.0.0.1:6987/api/llm/providers/ -b cj.txt -H "X-CSRF-Token: $CSRF" \
  -H "Content-Type: application/json" -d '{
    "provider_type":"openai","name":"OpenRouter",
    "base_url":"https://openrouter.ai/api/v1","api_key":"sk-or-v1-...",
    "default_model":"anthropic/claude-sonnet-5","is_active":true,"is_default":true}'
```

3. The connection test reports success.
4. `curl -s http://127.0.0.1:8080/api/providers/openai` → `network_config` has no `base_url` (or the stock one).
5. Any chat completion → `no keys found that support model: anthropic/claude-sonnet-5`.

## Current Bifrost admin API

| Method | Path | Notes |
|---|---|---|
| GET | `/api/providers/{name}` | provider doc, includes `network_config`, **no `keys`** |
| PUT | `/api/providers/{name}` | accepts `network_config` |
| GET | `/api/providers/{name}/keys` | `{"keys":[...],"total":N}`, values masked |
| PUT | `/api/providers/{name}/keys/{key_id}` | 404 on unknown id |
| POST | `/api/providers/{name}/keys` | creates |

## The `/v1` subtlety

Vigil stores what the SDK needs (`https://openrouter.ai/api/v1`); Bifrost stores the API root and appends the versioned path itself. Pushing the stored value verbatim yields 404 on every upstream call, so the `/v1` suffix has to be trimmed for `openai`-type providers.

## Suggested fix

I have a working patch and am happy to open a PR if you'd like:

1. `_get_provider_keys()` reading the sub-resource, falling back to the embedded array for older gateways.
2. `_put_provider_key()` writing to `/api/providers/{name}/keys/{key_id}`.
3. `push_provider_key` / `sync_provider_models` migrated onto those.
4. New `sync_provider_base_url()` writing a normalized `network_config.base_url`, preserving the rest of `network_config`, skipping the PUT when already correct.
5. Called from `_do_sync_all_provider_models` **before** the model sync — the allow-list is meaningless if the gateway resolves the wrong host.

Deliberately excluded from base_url sync:
- `anthropic` — Bifrost's built-in Anthropic provider ignores its own `network_config.base_url`; that redirect happens on the SDK client via `ANTHROPIC_BASE_URL` (already documented at `services/llm_clients.py:41`).
- `ollama` — Vigil stores the host-side URL (`http://localhost:11434`) while Bifrost resolves it from inside a container, where that address means the container itself (cf. `services.ollama_process.container_base_url`).

## Verification

Restoring `docker/bifrost/config.json` to stock and force-recreating the container gives the broken baseline:

```
base_url: None
models:   ['gpt-4', 'gpt-4-turbo', 'gpt-4o', 'gpt-4o-mini']
```

A single `POST /api/llm/providers/refresh-models` with the patch applied:

```json
{"bifrost": {"ollama": true, "openai": true}, "bifrost_base_url": {"openai": true}}
```

```
base_url: https://openrouter.ai/api      <- from the DB row, /v1 trimmed
models:   400                            <- live OpenRouter catalog
anthropic/claude-sonnet-5 present: True
```

A real chat completion then succeeds in ~7s. With the fix the manual `config.json` edit is no longer needed — Vigil configures the gateway itself.

26 unit tests pass, including new cover for the keys sub-resource, the legacy fallback, `/v1` normalization, the idempotent no-op, and the excluded provider types. The existing `_RecordingClient` tests needed updating since they encoded the stale contract.

## Unrelated note

`docker/docker-compose.yml:204` pins `maximhq/bifrost:latest`. The gateway's API contract changed under the code — which is exactly Defect B. A pinned tag would have surfaced this as a deliberate upgrade instead of a silent breakage.

## Factory grooming notes

Defect B is done (#640). Do not reimplement the keys subresource, a legacy embedded-`keys` fallback, or any “whole sync is a no-op” work. Do not pin the Bifrost image tag. Do not add an AI Router (or any vendor) fixture or preset. Do not reopen or wait on #703 — that rewrite is closed and the release branch is gone.

Ship only Defect A against current `main`: the OpenAI-compatible `base_url` stored on `LLMProviderConfig` must reach Bifrost so inference uses that host instead of `api.openai.com`.

How:
- Keep one egress. `core/llm/router/router.py` already points the OpenAI SDK at `{bifrost_url}/v1`. Do not point the SDK at the custom URL (that is the Anthropic `ANTHROPIC_BASE_URL` bypass in `core/llm/providers/clients.py`; Bifrost ignores `network_config.base_url` for `anthropic`).
- Add the push in `core/llm/bifrost/admin.py` next to the existing key/model sync. GET/PUT `/api/providers/{name}` (the provider document, not `/keys`). Write `network_config.base_url`; preserve every other `network_config` field; do not send `keys` on that PUT.
- Call it from the existing `_do_sync_all_provider_models` pass (before the allow-list write is fine). No new module, config surface, or sync pipeline. Bifrost has one document per provider type: when several `openai` rows exist, use the default row, same as the key already does.
- Scope: `openai` rows with a custom `base_url` only. Skip `anthropic` and `ollama` (see `_UNMANAGED_PROVIDER_TYPES` and the `clients.py` comment). Do not invent first-class OpenRouter/vLLM/Together types.
- Normalize only a trailing `/v1` (Bifrost appends the versioned path). Keep any other path prefix. Skip the PUT when the value is already correct. Empty/`https://api.openai.com/v1` may stay as Bifrost’s stock host.
- Tests in `tests/unit/llm/test_bifrost_admin.py`: `_RecordingClient` must distinguish GET `/api/providers/{name}` from GET `.../keys` (one shared payload is what hid Defect B). Cover `/v1` trim vs no-suffix, idempotent no-op, anthropic/ollama skipped, unrelated `network_config` fields kept, a namespaced model still on the key allow-list after a base_url refresh, and the regression “connection test can succeed while Bifrost `network_config` still has no custom host.” Do not change the key `value` payload shape in this PR.

samarmstrong:
Closes #586.

## The remaining break

Defect B (keys moved to `/api/providers/{name}/keys`, so the whole sync was a no-op) already shipped in #640. Defect A did not: `LLMProviderConfig.base_url` is persisted, validated, and read into the catalog-sync row dict, but nothing writes Bifrost `network_config.base_url`.

Discovery and the connection test both dial the custom host directly, so OpenRouter / LiteLLM / vLLM / Together can look healthy in the UI while every completion still goes to the stock OpenAI host.

## What changed

`sync_provider_base_url` GET/PUTs `/api/providers/{name}` (the provider document, not `/keys`) from the existing `_do_sync_all_provider_models` pass, before the allow-list write:

- Scope is `openai` rows with a custom host. Anthropic is skipped (Bifrost ignores `network_config.base_url` for that provider). Ollama is skipped (the stored URL is host-side `localhost` and would mean the Bifrost container itself).
- Only a trailing `/v1` is trimmed; other path prefixes stay. Empty / stock OpenAI URLs leave Bifrost's default host, and a previously pushed custom host is cleared so revert/switch-default does not keep routing to the old gateway.
- Unrelated `network_config` fields are preserved. Only writable provider fields are sent; `keys` and read-back fields (`status`, `config_hash`, `name`) are never echoed.
- When several `openai` rows exist, the default row is used, matching the key already chosen for the allow-list.

SDK traffic still goes through `{bifrost_url}/v1`. This does not point the OpenAI client at the custom URL.

## Tests

`_RecordingClient` now distinguishes GET `/api/providers/{name}` from GET `.../keys` (one shared payload is what hid Defect B). Coverage includes `/v1` trim vs no-suffix, idempotent no-op, anthropic/ollama skipped, unrelated `network_config` kept, a namespaced model still on the key allow-list after a base_url refresh, clearing a stale custom host on revert to stock, omitting read-back fields on PUT, and the regression that discovery/connection-test success does not configure Bifrost's host.

The key `value` payload shape is unchanged (bare string).

## Verification

- `tests/unit/llm/test_bifrost_admin.py`: **34 passed** (plus 3 one-egress ratchet tests)
- `black --check`, `isort --check-only`, `flake8` clean on both files

Not verified against a running Bifrost image in this environment (no gateway). The original report's failure mode is covered by the recording-client suite.

<div><a href="https://cursor.com/agents/bc-d36bdc3d-3d67-44c8-895e-e867c4ed0952?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d36bdc3d-3d67-44c8-895e-e867c4ed0952&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/core/llm/bifrost/admin.py
+++ b/core/llm/bifrost/admin.py
@@ -68,9 +68,25 @@ def _bifrost_base_url() -> str:
 _UNMANAGED_PROVIDER_TYPES = frozenset({"ollama", "vertex"})
 
 # Public OpenAI cloud — Bifrost already points there. Custom OpenAI-compatible
-# hosts (OpenRouter, LiteLLM, vLLM, ...) are the ones we must push.
+# hosts (OpenRouter, LiteLLM, vLLM, ...) are the ones we must push. Derived
+# from the SSRF allow-list so this file never names the stock host (one-egress
+# ratchet); ``.openai.com`` keeps Azure-style hosts from matching.
 _STOCK_OPENAI_HOSTS = frozenset(
-    host for host in DEFAULT_ALLOWED_PROVIDER_HOSTS if "openai" in host
+    host for host in DEFAULT_ALLOWED_PROVIDER_HOSTS if host.endswith(".openai.com")
+)
+
+# Fields Bifrost accepts on PUT /api/providers/{name}. Everything else on the
+# GET document is read-back (name, status, config_hash, keys, ...).
+_PROVIDER_WRITE_FIELDS = frozenset(
+    {
+        "network_config",
+        "concurrency_and_buffer_size",
+        "proxy_config",
+        "send_back_raw_request",
+        "send_back_raw_response",
+        "store_raw_request_response",
+        "custom_provider_config",
+    }
 )
 
 # Read-only/derived fields Bifrost returns but rejects or ignores on write.
@@ -413,7 +429,9 @@ def sync_provider_base_url(provider_type: str, base_url: Optional[str]) -> bool:
     stored URL is host-side and would resolve to the Bifrost container itself.
 
     GET/PUT ``/api/providers/{name}`` (the provider document, not ``/keys``).
-    Other ``network_config`` fields are preserved; ``keys`` are never sent.
+    Other ``network_config`` fields are preserved; only writable provider
+    fields are sent, never ``keys``. Reverting to empty/stock clears a
+    previously pushed custom host so inference does not keep using it.
 
     Returns True on success or when there is nothing to write. Failures are
     logged and return False so the catalog sync still proceeds.
@@ -421,23 +439,30 @@ def sync_provider_base_url(provider_type: str, base_url: Optional[str]) -> bool:
     if not provider_type or provider_type != "openai":
         return True
     target = _custom_openai_base_url(base_url)
-    if target is None:
-        return True
 
     with httpx.Client() as client:
         doc = _get_provider_document(provider_type, client)
         if doc is None:
-            return False
-        doc.pop("keys", None)
+            # Stock/empty has nothing to write; a custom host cannot land if
+            # the provider document is missing.
+            return target is None
         network = dict(doc.get("network_config") or {})
         current = (network.get("base_url") or "").rstrip("/")
-        if current == target:
+        if target is None:
+            # Leave Bifrost's default host. If we previously pushed a custom
+            # URL, drop it so inference does not keep hitting the old gateway.
+            if not current or _custom_openai_base_url(current) is None:
+                return True
+            network.pop("base_url", None)
+        elif current == target:
             return True
-        network["base_url"] = target
-        doc["network_config"] = network
+        else:
+            network["base_url"] = target
+        body = {k: v for k, v in doc.items() if k in _PROVIDER_WRITE_FIELDS}
+        body["network_config"] = network
         url = _provider_url(provider_type)
         try:
-            r = client.put(url, json=doc, timeout=_DEFAULT_TIMEOUT)
+            r = client.put(url, json=body, timeout=_DEFAULT_TIMEOUT)
             if r.status_code >= 400:
                 logger.warning(
                     "Bifrost: PUT %s returned %s: %s",
@@ -449,7 +474,7 @@ def sync_provider_base_url(provider_type: str, base_url: Optional[str]) -> bool:
             logger.info(
                 "Bifrost: set network_config.base_url for provider %s to %s",
                 provider_type,
-                target,
+                target or "default",
             )
             return True
         except Exception as e:
--- a/core/llm/bifrost/admin.py
+++ b/core/llm/bifrost/admin.py
@@ -12,7 +12,7 @@ manager under ``llm_provider_<id>_api_key``. Pushing via the API keeps a
 single source of truth in the secrets manager, and is why the seeded
 ``config.json`` key is an empty placeholder rather than a real credential.
 
-Three properties of that API shape the code below, all learned the hard way:
+Four properties of that API shape the code below, all learned the hard way:
 
 * **Keys are a subresource.** They live at ``/api/providers/{name}/keys``,
   *not* on the ``/api/providers/{name}`` document — which returns no ``keys``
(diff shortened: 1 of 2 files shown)


samarmstrong:
🏭 **Factory environment agent.** Reading this PR (alongside 2 other recent factory PRs) it found a gap in the cloud-agent environment and opened https://github.com/Vigil-SOC/vigil/pull/784 to close it.

Issue #516, opened 2026-07-30
18 of 40 MCP servers fetch their own code at launch; 13 unpinned, 2 from git HEAD


craig-dt:
## Summary

18 of the 40 MCP servers in `mcp-config.json` do not ship with Vigil — they download their own code from npm, PyPI, or GitHub every time the server process launches. 13 of those resolve to whatever version the registry happens to serve at that moment.

## Why this matters

Three separate problems, all stemming from the same design:

**Failures surface at the worst possible time.** MCP servers are spawned lazily, so a fetch failure is not a startup error — it is an agent stalling mid-investigation. The platform passes every health check and looks healthy until an analyst actually uses it. Support cost is high because the symptom is far from the cause.

**The dependency set is not reproducible.** Two identical deployments installed a week apart can be running different integration code. That defeats change control, and in a regulated or audited environment an unpinned dependency pulled at runtime is difficult to defend.

**It is a supply-chain exposure.** Whatever the registry serves at launch executes inside Vigil with the credentials for that integration. Two servers pull from a git default branch, so any push to those repositories lands in the next process start.

## Evidence

Runtime-fetched servers, by how they resolve:

**Pull from a git HEAD at launch (2):**
- `sentinelone` — `uvx --from git+https://github.com/Sentinel-One/purple-mcp.git`
- `cribl-stream` — `uvx --from git+https://github.com/pebbletek/cribl-mcp.git`

**Version-pinned (3):** `azure-sentinel`, `splunk`, `jira` — all via `mcp-remote@^0.1.16`

**Unpinned — resolves to whatever the registry serves (13):**
`github`, `security-detections`, `loglm`, `crowdstrike`, `pagerduty`, `okta`,
`joe-sandbox`, `virustotal`, `shodan`, `gcp-secops`, `gcp-threat-intel`, `gcp-scc`,
and `aws-security` (explicitly `@latest`)

Launch mechanisms in use: `npx -y` (6), `uvx` (8), `uv --directory` (1), `docker run` (1),
plus `mcp-remote` for the remote-endpoint servers.

## Proposed fix

Two directions, not mutually exclusive:

1. **Make it deterministic.** Pin every runtime-fetched server to an exact version, and
   replace the two git-HEAD references with a tagged release or a vendored copy. Cheap, and
   removes the reproducibility and supply-chain arguments immediately.
2. **Make failures loud and early.** Add a startup preflight that verifies each enabled
[...]

PR #785, opened 2026-09-01
feat(mcp): connect enabled servers concurrently at API boot

samarmstrong:
Closes #702

## Why

Production API boot still dialed enabled MCP servers one at a time. Those stdio children do not share a process, so overlapping the awaits is the actual startup cost. Settings/UX “deep integration” work is out of this PR.

## What

- `_connect_external_services` now connects enabled servers concurrently via `asyncio.gather(..., return_exceptions=True)`.
- Per-server failure isolation is unchanged: a failed or raising connect is logged and the rest continue.
- Disabled servers are still skipped before dial; credential-dormant servers still return without spawning.
- After the connect wave, `MCPClient.list_tools` still reads `tools_cache` for servers that connected. No second parallel connect path, no connect deadline, no `.mcp-capability-cache.json`, no worker-pool type, and `_connection_locks` is left unused. `close_all` stays sequential. Existing `data/mcp_tools_cache.json` / `tools_cache` is reused.

## Test plan

- [x] `pytest tests/unit/api/test_local_startup.py` — concurrent overlap, failure isolation (failing server is dialed alongside successes), dormant not counted as connected
- [x] Related MCP unit tests (`test_mcp_required_env.py`, `test_mcp_registry_population.py`)
- [x] `black` / `flake8` on the changed files

This is not a user-visible UI change; no browser walkthrough.

<div><a href="https://cursor.com/agents/bc-a5594a86-cd9e-4e9e-bb2e-38caa9f33dbc?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a5594a86-cd9e-4e9e-bb2e-38caa9f33dbc&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #787, opened 2026-09-01
Migrate in-repo MCP servers to the mcp 2.x API and unpin

### LINKED ISSUE #698: Planning: migrate the 21 MCP tool servers to the mcp 2.0 API (and unpin)
## Why this exists

#697 pinned `mcp>=1.23.0,<2.0.0` to stop CI and fresh installs from breaking. That pin is a holding action, not a decision — it should come out, and this issue is where we work out what "coming out" costs and what shape we want on the other side. Scoping is settled enough below to argue about; the approach is not.

## What mcp 2.0 changed

The decorator registration on the lowlevel `Server` is gone. Every affected module does this today:

```python
server = Server("ip-geolocation")

@server.list_tools()
async def handle_list_tools() -> list[types.Tool]:
    return [types.Tool(name=..., description=..., inputSchema={...}), ...]

@server.call_tool()
async def handle_call_tool(name, arguments): ...   # one dispatcher, switches on name

async def main():
    async with mcp.server.stdio.stdio_server() as (read, write):
        await server.run(read, write, InitializationOptions(...NotificationOptions()...))
```

In 2.0 the lowlevel `Server` keeps only `add_request_handler()` / `add_notification_handler()`, and the batteries-included path is a new `MCPServer` exposing `tool` / `add_tool` / `run_stdio_async` and friends. So the delta isn't a rename — it's a different registration model. `MCPServer` derives tool schemas per decorated function, where we currently hand-write `inputSchema` dicts and dispatch by name ourselves.

## Scope

21 modules, ~4,163 lines, **74** `types.Tool(...)` definitions, and **21** near-identical copies of the `stdio_server()` + `InitializationOptions` boilerplate.

They fall into three groups, and the third is the interesting one:

**18 vendor servers** — `core/integrations/{alienvault_otx,anyrun,azure_ad,cape_sandbox,carbon_black,cloudflare,elastic,hybrid_analysis,ip_geolocation,microsoft_defender,microsoft_teams,misp,palo_alto,slack,splunk,url_analysis,vstrike}/tool.py` plus `custom_integration_service.py`.

**3 Vigil-facing servers** — `tools/mcp/{approval,attack_layer,tempo_flow}.py`.

**`core/integrations/custom_integration_service.py`, which is not just another server.** It carries a code *template* it hands to users for authoring custom integrations (~line 311), it *validates* user-supplied code by string-matching the 1.x API — `"@server.list_tools()"`, `"@server.call_tool()"`, `"Server("`, `"async def main():"` at lines 622-625 — and it **writes generated servers to disk** as `{integration_id}_server.py` (line 524). So any custom integration a user has already generated is a file on their disk, written against the 1.x decorator API, that we did not author and cannot see. Dropping the pin breaks those, and a sweep of our own tree does not touch them.

## Options, to be argued with

**A. Minimal port on the lowlevel `Server`.** Rebuild the two decorators as `add_request_handler()` calls, ideally behind one shared helper so it's written once rather than 21 times. Smallest diff, keeps the hand-written schemas and the name-dispatch shape, lowest risk of behaviour drift. Doesn't reduce the duplication meaningfully and leaves us on the low-level API.

**B. Port to `MCPServer` idiomatically.** One decorated function per tool, schemas derived from signatures, `run_stdio_async()` replacing the boilerplate. Deletes a lot of the 4,163 lines and the 74 hand-written schemas. Much larger diff, and the derived schemas will not be byte-identical to the current ones — which matters, because those schemas are what the agents see.

**C. A compatibility shim.** A thin module presenting the 1.x decorator surface on top of 2.x, imported by all 21. Unpins quickly and covers user-authored files if they can be made to import it. Adds a layer we own forever and would want an exit plan for.

**D. Stay pinned deliberately.** Write down that we track mcp 1.x, with a trigger for revisiting (1.x losing support, or something in 2.x we actually want). Cheapest, honest, and defensible — but `claude-agent-sdk` already permits `mcp<3.0.0`, so we would be holding a line the SDK has let go of.

A and B aren't exclusive: A unpins us, B follows per-server once the seam exists.

## Open questions

1. **What do we owe already-generated custom integrations?** This is the one I'd want settled first, because C looks much more attractive if the answer is "they must keep working untouched" and much less so otherwise. Is there any inventory of what users have generated?
2. **Do the derived schemas in B change agent behaviour?** 74 tool schemas are part of the prompt surface. A schema refactor is a silent behaviour change unless something pins it down.
3. **Does the shared helper come first?** 21 copies of the same boilerplate is the reason this migration is 21 files instead of 1. Consolidating before migrating may be the cheaper order, and is worth doing regardless of which option wins.
4. **Who validates the vendor servers actually still work?** `tests/unit/integrations/` covers imports and handler behaviour, but these run as subprocesses against real vendors in production — passing tests is not the same as a working integration.

## Non-goals

- Not proposing we unpin on a deadline. The pin is stable; there's no forcing function beyond wanting to be current.
- Not choosing between A–D here.
- Not touching the MCP servers that aren't ours — #516 covers the 18 that fetch their own code at launch, which is a different problem with a similar smell.

## Related

- #697 — the pin this issue exists to remove, with the timeline of how mcp 2.0 arrived.
- #516 — unpinned external MCP servers.
- #696 — dependency reproducibility planning. This break is the worked example: our protection was `claude-agent-sdk`'s own `mcp<2.0.0` ceiling, and it evaporated when they relaxed it in 0.2.140.

samarmstrong:
Closes #698

The `mcp>=1.23.0,<2.0.0` pin was a holding action around a deleted 1.x API. This ports the in-repo servers to the official 2.x surface, then drops the cap.

## What changed

Lowlevel servers (`core/integrations/*/tool.py` and `tools/mcp/{approval,attack_layer,tempo_flow}.py`) now register with `Server(..., on_list_tools=..., on_call_tool=...)`. Hand-written `types.Tool(..., inputSchema={...})` lists and name-dispatch bodies are unchanged. Thin `_on_*` wrappers keep `handle_list_tools()` / `handle_call_tool(name, arguments)` callable as today, so existing tests do not need a `(ctx, params)` rewrite.

`_on_call_tool` catches handler exceptions and returns `CallToolResult(is_error=True)`. mcp 2.x no longer does that itself, and an uncaught raise would become a JSON-RPC error (a dropped session to our client). Handler-caught failures still return JSON in `TextContent` with `is_error=False`, as they did on 1.x.

Also ported:

- `tools/mcp/deeptempo_findings.py`: `FastMCP` → `MCPServer`. `mcp.run()` and `@mcp.tool()` stay.
- `tools/mcp/base.py`: unused `create_server()` deleted with the FastMCP import.
- `core/integrations/mcp/client.py`: `tool.input_schema` and `result.is_error` (the `hasattr(..., "isError")` fallback would silently report success on 2.x). Cache/registry dict keys remain `"inputSchema"`.
- Custom-integration generator template and `validate()` substring checks updated to the 2.x surface. No compatibility layer for already-generated `~/.vigil/custom_integrations/*_server.py`.

`core.integrations.mcp.service.MCPServer` (process record) is untouched. `_detect_server_type`'s `"fastmcp"` label stays. Tool servers stay on `httpx`; mcp 2.x's `httpx2` is internal.

## Out of scope

- FastMCP-style per-function rewrite of the 74 hand-written schemas
- A 1.x decorator shim
- External servers in `mcp-config.json` (#516)

## Verification

- black / isort / flake8 on `services/` and `core/`
- 560 unit tests in `tests/unit/integrations/`, `tests/unit/_ratchets/`, and related MCP tests
- Smoke-import of every migrated server under mcp 2.1.1
- Wrapper test: vstrike `int("bad")` still raises from `handle_call_tool`, and `_on_call_tool` turns it into `CallToolResult(is_error=True)`

<div><a href="https://cursor.com/agents/bc-8fb0998b-0192-4973-b912-3a5223dcd352?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8fb0998b-0192-4973-b912-3a5223dcd352&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/core/integrations/alienvault_otx/tool.py
+++ b/core/integrations/alienvault_otx/tool.py
@@ -149,9 +149,14 @@ async def _on_list_tools(_ctx, _params):
 
 
 async def _on_call_tool(_ctx, params):
-    return types.CallToolResult(
-        content=await handle_call_tool(params.name, params.arguments)
-    )
+    try:
+        content = await handle_call_tool(params.name, params.arguments)
+    except Exception as exc:
+        return types.CallToolResult(
+            content=[types.TextContent(type="text", text=str(exc))],
+            is_error=True,
+        )
+    return types.CallToolResult(content=content)
 
 
 server = Server(
--- a/core/integrations/alienvault_otx/tool.py
+++ b/core/integrations/alienvault_otx/tool.py
@@ -22,7 +22,6 @@ from core.integrations._base.config import resolve
 from core.integrations.alienvault_otx.descriptor import ALIENVAULT_OTX
 
 logger = logging.getLogger(__name__)
-server = Server("alienvault-otx")
 
 
 def result(data):
@@ -33,7 +32,6 @@ def get_config():
     return resolve(ALIENVAULT_OTX)
 
 
-@server.list_tools()
 async def handle_list_tools():
     return [
         types.Tool(
@@ -66,7 +64,6 @@ async def handle_list_tools():
     ]
 
 
-@server.call_tool()
 async def handle_call_tool(name: str, arguments: dict | None):
     config = get_config()
     api_key = config.get("api_key")
@@ -145,6 +142,23 @@ async def handle_call_tool(name: str, arguments: dict | None):
         )
     except Exception as e:
         return result({"error": str(e)})
+
+
+async def _on_list_tools(_ctx, _params):
+    return types.ListToolsResult(tools=await handle_list_tools())
+
+
+async def _on_call_tool(_ctx, params):
+    return types.CallToolResult(
+        content=await handle_call_tool(params.name, params.arguments)
+    )
+
+
+server = Server(
+    "alienvault-otx",
+    on_list_tools=_on_list_tools,
+    on_call_tool=_on_call_tool,
+)
 
 
 async def main():
--- a/core/integrations/anyrun/tool.py
+++ b/core/integrations/anyrun/tool.py
@@ -22,14 +22,12 @@ from core.integrations._base.config import resolve
 from core.integrations.anyrun.descriptor import ANYRUN
 
 logger = logging.getLogger(__name__)
-server = Server("anyrun")
 
 
 def result(data):
     return [types.TextContent(type="text", text=json.dumps(data, indent=2))]
 
 
-@server.list_tools()
 async def handle_list_tools():
     return [
         types.Tool(
@@ -53,7 +51,6 @@ async def handle_list_tools():
     ]
 
 
-@server.call_tool()
 async def handle_call_tool(name: str, arguments: dict | None):
     config = resolve(ANYRUN)
     api_key = config.get("api_key")
@@ -99,6 +96,23 @@ async def handle_call_tool(name: str, arguments: dict | None):
         return result({"error": f"Unknown tool: {name}"})
     except Exception as e:
         return result({"error": str(e)})
+
+
+async def _on_list_tools(_ctx, _params):
+    return types.ListToolsResult(tools=await handle_list_tools())
+
+
+async def _on_call_tool(_ctx, params):
+    return types.CallToolResult(
+        content=await handle_call_tool(params.name, params.arguments)
+    )
+
+
+server = Server(
+    "anyrun",
+    on_list_tools=_on_list_tools,
+    on_call_tool=_on_call_tool,
+)
 
 
 async def main():
--- a/core/integrations/anyrun/tool.py
+++ b/core/integrations/anyrun/tool.py
@@ -103,9 +103,14 @@ async def _on_list_tools(_ctx, _params):
 
 
 async def _on_call_tool(_ctx, params):
-    return types.CallToolResult(
-        content=await handle_call_tool(params.name, params.arguments)
-    )
+    try:
+        content = await handle_call_tool(params.name, params.arguments)
+    except Exception as exc:
+        return types.CallToolResult(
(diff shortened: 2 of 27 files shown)


samarmstrong:

I read [#787](https://github.com/Vigil-SOC/vigil/pull/787) and [#785](https://github.com/Vigil-SOC/vigil/pull/785). Neither implementer reported a check they would have run and could not because the machine lacked a tool, package, or daemon.

**#787** (mcp 2.x migration) ran black/isort/flake8, smoke-imported the migrated servers, and passed 560 related unit tests under mcp 2.1.1. **#785** (concurrent MCP connects) ran the targeted pytest files plus black/flake8. Both then had green GitHub CI. They skipped browser walkthroughs because these are not UI changes, not because a browser was missing.

The repo already has a working Docker daemon in `.cursor/` from #781. Python venv/deps are not preinstalled, but both agents installed them and completed verification. That is bootstrap, not a blocked check. No environment PR.

PR #791, opened 2026-09-02
Stop leaking raw exception text on leftover 5xx handlers

samarmstrong:
Closes #540

`HTTPException.detail` still interpolated `str(e)` on a handful of 5xx paths, so the `#655` sanitizer never ran. Clients could see gateway/S3/DB internals on those responses.

## What changed

- Drop wrap-only `except Exception` catch-alls on enrichment, MCP reload, and Darktrace/Cloudflare ingestion so unhandled failures hit the existing handler (`"Internal server error"` + trace id).
- Keep FindingNotFound / NoProviderConfigured / ProviderUnavailable mappings on enrichment.
- Autostart `OSError` stays (sibling `ValueError` 400), but the 500 detail is now a static string.
- Enrichment unit tests no longer assert the leaked text; unexpected failures are expected to propagate.

## Out of scope

No new sanitizer, no EnrichmentError→SOCError conversion, no repo-wide re-audit of 4xx or admin test-connection 502s.

<div><a href="https://cursor.com/agents/bc-f2978b1e-323d-4de5-96c5-6b812fc9aec1?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f2978b1e-323d-4de5-96c5-6b812fc9aec1&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #793, opened 2026-09-02
Generate frontend types from the cases API OpenAPI contract

### LINKED ISSUE #699: Epic: make frontend types derive from the API instead of being hand-written
## Problem

The frontend hand-writes its beliefs about what every endpoint returns, and nothing checks those beliefs. When they're wrong, nothing fails — a field renders as `—`, or a card renders its empty state forever.

Two confirmed casualties so far, both in `clients/web/src/redesign/screens/cases/CaseSections.tsx`:

- **#561** (fixed in #584) — the watchers card read `added_at`, the API sends `created_at`. Every row showed "Watching since —" for months.
- **The SLA card** (not yet filed separately) — wrong response envelope *and* zero overlap between the declared interface's field names and the real payload. `GET /cases/{id}/sla` returns its dict at the top level, but the client reads `r.data?.sla`; the interface declares `id / policy_name / status / due_date / created_at`, the response sends `case_id / sla_policy_id / response_due / resolution_due / is_breached / is_paused / health_status`. The card has never worked.

Current scale of the exposure: **167** hand-written interfaces across `src/redesign`, and **137** `as SomeType` casts in `src/redesign` + `src/services`. Every one of those casts is a place TypeScript has been told to stop checking.

## Why this keeps happening

Axios responses are `any`. A cast at the fetch site turns `any` into a confident-looking type without verifying anything:

```ts
casesApi.getSLA(caseId).then((r) => (r.data?.sla ?? null) as SLA | null)
```

Both bugs above passed `tsc --noEmit`, eslint, and code review. There is no mechanism that could have caught either one.

## End state

The frontend imports generated types produced from the backend's own OpenAPI schema. A field rename on the backend becomes a red squiggle and a failed build, not a blank card discovered later by a user.

## Why this is a big tackle

**The generator is the easy half.** The hard half is that the schema currently doesn't say much: roughly **56 of 346** routes declare a `response_model` (~16%). Most return bare Python dicts, so OpenAPI records them as "returns something."

Worse, the coverage gap lines up exactly with the damage — `services/api/routers/cases.py` has **0 of 40** routes annotated, and that's the file behind both known bugs.

So codegen alone fixes nothing today. The work is annotating routes, and that's ~290 routes across ~50 router modules. This cannot land as one PR and should not try to.

## Approach: one module per stab

Each slice is independently valuable and independently mergeable. A module is done when its routes declare response types, its generated types exist, and the screens consuming it no longer cast. Nothing depends on a later slice.

**Suggested ordering** — by damage done, not by size:

| Priority | Module | Routes | Annotated | Why |
|---|---|---|---|---|
| 1 | `services/api/routers/cases.py` | 40 | 0 | Both known bugs live here |
| 2 | `core/cases/*_router.py` (metrics, sla_policies, templates) | 26 | 0 | Same screens, same risk |
| 3 | `services/api/routers/config.py` | 32 | 0 | Largest unannotated surface |
| 4 | `core/workflows/workflows_router.py` | 14 | 0 | |
| 5 | `services/api/routers/orchestrator.py` | 16 | 0 | |
| 6 | `services/api/routers/findings.py` | 8 | 0 | |

Later slices can be picked by whoever is already in that code. Routers that are already annotated (`skills.py` 8/9, `graph.py` 5/5, `ingestion.py` 7/12) are cheap wins — they mostly just need the frontend side swapped over.

**Per-slice definition of done:**

1. Every route in the module declares a `response_model`. Reuse the Pydantic schemas in `core/storage/schemas/` where they exist; write one where the route returns a computed dict (the SLA status payload is the archetype).
2. Regenerate the TypeScript types.
3. Replace that module's hand-written interfaces in the frontend with the generated ones, and delete the corresponding `as` casts.
4. Fix whatever drift step 3 surfaces — expect each slice to uncover bugs. That is the point, not a side effect.

Report what step 4 turned up in the PR description so we can see whether the effort is paying.

## Sequencing

Phase 0 is the only slice that blocks anything else: pick the generator, wire it into the build, and land it against one already-annotated module (`graph.py` is 5/5 — the smallest honest proof). After that, module slices can run in parallel and in any order.

## Open decisions

- **Generator.** `openapi-typescript` (types only, no runtime, minimal footprint) vs. a client generator like `orval` that would also replace the hand-written axios layer in `clients/web/src/services/api.ts`. Types-only is the smaller bite and does not disturb a working service layer; worth deciding before Phase 0.
- **Generated types: committed or built?** Committing them makes drift visible in PR diffs and keeps CI simple. Building them requires a running backend at build time. Leaning committed, with a CI check that regeneration produces no diff.
- **Where generated output lives** — `clients/web/src/services/generated/` is the obvious spot, but it should be clearly marked do-not-edit.

## Non-goals

- Not rewriting the axios service layer (unless the generator decision above says otherwise).
- Not changing any endpoint's actual response shape. This epic documents what the API already returns; behaviour changes belong in their own issues.
- Not blocking on 100% coverage. Partial coverage is strictly better than none, and the epic stays open as long as it takes.

## Alternative considered

A test that fetches `/openapi.json` and asserts field names match the frontend interfaces. Cheaper, but it needs the same response types to exist before it can check anything meaningful — so it postpones the actual work rather than avoiding it, and gives a weaker guarantee. Not recommended as a substitute, though a slice could add one for a module it cannot fully convert yet.

---

*Route counts are grep-based (`@router.<verb>` decorators vs. `response_model=` occurrences) — good enough for ordering the work, not exact.*

samarmstrong:
Closes #699

The frontend was inventing response shapes and casting `any` axios payloads into them. Two cards already paid for that: watchers (`added_at` vs `created_at`, #561) and SLA (wrong envelope *and* zero field overlap). This PR lands the generator and converts the cases module — the slice the epic said to do first.

## What landed

- Every route in `services/api/routers/cases.py` declares a `response_model`. Entity rows reuse the dump schemas; computed dicts (SLA status, list envelopes, purge/merge/search) get their own models. GET `/cases/{id}` stays `CaseSchema` (`finding_ids` only) — switching to `CaseWithFindingsSchema` would change behaviour.
- `scripts/generate_frontend_types.py` dumps `app.openapi()` (no running server) and runs `openapi-typescript`. Output is committed at `clients/web/src/services/generated/schema.d.ts`. CI regenerates and fails on a diff.
- `casesApi` is typed from those schemas. `ApiCase` and the CaseSections wire interfaces are gone; view models (`CaseRow`, `Finding`) are unchanged. The axios layer is still hand-written.

## Drift this uncovered

Adapting the cards to the real payload (not the other way around):

- **SLA** — GET returns the `get_sla_status` dict at the top level (`health_status`, `response_due`, `resolution_due`, `is_paused`, …). The card was reading `r.data.sla` with `policy_name` / `due_date` / `status`. It has never worked. A 404 is now the empty state.
- **Tasks** — `task_id` and `PUT`, not `id` and `PATCH`.
- **Evidence** — `evidence_id`; `EvidenceAdd` requires `collected_by` and has no `url` (file path covers both).
- **IOCs** — `ioc_id` and `context`, not `id` / `description`.
- **Related cases** — `/relationships` with `{relationships: …}` (`relationship_id`). Title/status/priority are looked up from the already-loaded case list; they are not on the wire.
- **GET /cases** — `{cases, total}`, already what `useCases` read.
- **`getAuditLog` / `updateTags` / `bulkUpdate`** — still hit routes that do not exist. Not added.

## Out of scope (later slices)

config, workflows, orchestrator, findings, `core/cases/*_router.py`. Unannotated routes stay empty in the spec.

## Tests

- `CaseSchema` dump round-trips through `response_model` re-validation (`finding_ids` as strings, not ORM objects).
- Every cases route declares a `response_model`; GET `/{case_id}` is `CaseSchema`, not `CaseWithFindingsSchema`.
- Watchers card still pins `created_at`. SLA card pins the computed status payload and 404-as-empty. Tasks card toggles by numeric `task_id`.
- `tsc --noEmit`; frontend vitest for cases screens; pytest for the new unit tests and ORM schema parity.

<div><a href="https://cursor.com/agents/bc-b57ebde2-2b61-436a-95ea-4ec0bcfa0f53?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b57ebde2-2b61-436a-95ea-4ec0bcfa0f53&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


nestor-deeptempo (approved):
lgtm

samarmstrong:

Neither PR was blocked by a missing tool on the agent machine. I am not opening an environment PR.

**[#793](https://github.com/Vigil-SOC/vigil/pull/793)** ran the checks that actually prove the OpenAPI slice: `app.openapi()` regen, `tsc --noEmit`, cases vitest, pytest on the new response-model tests, then CI including Generated API Types. The Tests section lists those as done. Internal notes skipped a browser look because there was no live backend with real case payloads — that is not a missing binary. Docker is already on the image. Compose would still boot an empty database, so a screenshot of empty cards would not have checked the contract. The honest check is the regen + typecheck + card tests they already ran.

**[#791](https://github.com/Vigil-SOC/vigil/pull/791)** ran enrichment/webhook pytest (44 passed) plus black/isort/flake8 on the touched files, then all 16 CI jobs. No verification heading, and no “could not run X because the machine lacked Y.” Missing pytest on PATH at boot is normal; they created a venv and continued.

Both agents installed project deps themselves. That is not a gap `.cursor/environment.json` needs to close this week. Helm and Docker are already in the image from [#790](https://github.com/Vigil-SOC/vigil/pull/790) and [#781](https://github.com/Vigil-SOC/vigil/pull/781).

Issue #511, opened 2026-07-30
MCP child processes do not inherit egress proxy configuration


craig-dt:
## Summary

MCP servers are spawned with a filtered environment that drops `HTTPS_PROXY`, `NO_PROXY` and CA-bundle variables. In a proxied network every MCP integration is therefore unreachable, and the failure is invisible until an analyst runs an investigation.

## Why this matters

This is the single blocking defect for regulated-network deployments. MCP is how agents reach every security tool, so when it can't traverse the proxy, Vigil starts cleanly and then does nothing useful. Support cost is high because nothing in the startup logs indicates a problem.

## Evidence

The production spawn path filters child environment:

- `backend/main.py:551` — the backend connects MCP servers at startup via `mcp_client.connect_to_server(server_name, persistent=True)`.
- `services/mcp_client.py:254-257` — that path spawns with `StdioServerParameters(command=..., args=..., env=server.env)`.
- The MCP Python SDK merges the supplied `env` against `get_default_environment()`, whose inherit allowlist is exactly:

  ```
  ['HOME', 'LOGNAME', 'PATH', 'SHELL', 'TERM', 'USER']
  ```

  (verified against the installed `mcp` package: `mcp.client.stdio.DEFAULT_INHERITED_ENV_VARS`)

  So `HTTPS_PROXY`, `HTTP_PROXY`, `NO_PROXY`, `REQUESTS_CA_BUNDLE`, `SSL_CERT_FILE` and `NODE_EXTRA_CA_CERTS` are all dropped — even when the operator sets them on the Vigil process itself.

The two spawn paths also disagree:

- `services/mcp_service.py:86` — `env = os.environ.copy()` then `env.update(self.env)`. Full inheritance. Not the path the backend uses; tracked separately as #515.
- `services/mcp_client.py:257` — allowlist only. **This is the production path.**

## Impact

- All MCP integration traffic bypasses the configured proxy and is dropped at the egress firewall.
- 18 of the 40 servers in `mcp-config.json` launch via `npx -y`, `uvx`, `uv`, or `docker run` and fetch their own package at launch. Without proxy env they cannot reach npm/PyPI, so they fail before their own configuration is ever read. The launch-time fetch itself is tracked separately as #516.
- Behind a TLS-inspecting proxy, every MCP server fails certificate validation with no means to trust the internal CA.

## Proposed fix

[...]

Issue #125, opened 2026-04-23
MCP: Settings UI 'Start server' button always fails for stdio servers


mando222:
## Bug

`POST /api/mcp/servers/{name}/start` returns 500 for **every** server configured in `mcp-config.json`, because the implementation short-circuits stdio servers:

[services/mcp_service.py:461-463](services/mcp_service.py)
```python
if server.server_type == "stdio":
    logger.warning(f"Server {server_name} is stdio-based and designed for advanced MCP integration, not standalone monitoring")
    return False
```

Every server in `mcp-config.json` is stdio-based. So the UI's "Start" action on the Settings → MCP page is non-functional across the board.

The 4 currently-connected servers (`deeptempo-findings`, `tempo-flow`, `approval`, `attack-layer`) connect via a completely different path during backend FastAPI startup ([backend/main.py:405 `connect_to_server`](backend/main.py)). That path isn't exposed as a per-server action.

## Expected behavior

Clicking "Start" on a stdio server should:
1. Trigger `MCPClient.connect_to_server(server_name, persistent=True)` — the same code the startup loop uses.
2. Report the result back (status + last_error from `mcp_client.get_last_error()` if it failed).

## Related

- Issue #<this one> depends on [services/mcp_client.py](services/mcp_client.py) already tracking `last_errors` (landed in the MCP audit).
- Complements the credential-gating issue.

mando222:
Shipped.

## Root cause

The frontend enable toggle (\`PUT /api/mcp/servers/{name}/enabled\`) only persisted a flag to \`~/.deeptempo/mcp_server_enabled.json\` — it never actually triggered a connect. The backend startup loop at \`backend/main.py:405\` was the only place that called \`mcp_client.connect_to_server\`. Net effect for users: enabling a server did nothing until the next backend restart.

Separately, \`POST /servers/{name}/start\` + \`start-all\` / \`stop-all\` were dead code. They routed to a Popen-monitor path that explicitly refused stdio servers (\`services/mcp_service.py:461-463\`), which is *every* server in \`mcp-config.json\`. No frontend component called them.

## Fix (Option A — one lever, not two)

- \`PUT /servers/{name}/enabled\` is now transactional: on enable it calls \`mcp_client.connect_to_server\`; on disable it calls \`disconnect_from_server\`. Response carries \`{connected, error}\` so the UI can flip the toggle back and surface the real reason on failure.
[...]

PR #796, opened 2026-09-02
Remove unused MCP Popen spawn; one child env path remains

### LINKED ISSUE #515: Two MCP spawn paths with different env semantics; one is dead code
## Summary

There are two code paths for spawning MCP servers, they apply different environment semantics, and one of them is dead code. Reading the codebase gives a misleading picture of how child environment is actually handled.

## Why this matters

This is what made the proxy defect in #511 hard to see. `mcp_service.py` inherits the full parent environment, which reads as "MCP children get the process env" — but the path that actually runs does not. Anyone reasoning about child environment from the wrong file draws the wrong conclusion, and a fix applied to the wrong path would appear correct in review and change nothing at runtime.

## Evidence

Two spawn implementations with different behaviour:

- `services/mcp_service.py:78-95` — `MCPServer.start()` uses `subprocess.Popen` with
  `env = os.environ.copy()` then `env.update(self.env)`. Full parent inheritance.
- `services/mcp_client.py:254-257` — `StdioServerParameters(command=..., args=..., env=server.env)`.
  The MCP SDK merges that against a six-variable allowlist, so nothing else is inherited.

Only the second one runs. `backend/main.py:551` connects servers via
`mcp_client.connect_to_server(server_name, persistent=True)`.

`MCPServer.start()` appears to have no callers. Searching for `.start()` across
`services/mcp_service.py`, `services/mcp_client.py`, `backend/main.py` and `daemon/` returns
only unrelated aiohttp sites (`daemon/metrics.py:224`, `daemon/poller.py:558`). Worth a
second pair of eyes before deleting, in case something reaches it dynamically.

## Proposed fix

Confirm `MCPServer.start()` is unreachable and remove it, along with any now-unused
supporting state on `MCPServer` (`self.process` and friends). If it turns out to be
reachable, consolidate the two paths onto one environment policy instead.

Either way the outcome is that there is exactly one answer to "what environment does an MCP
child get".

## Acceptance criteria

- [ ] Exactly one MCP spawn path remains, or both provably apply identical environment policy
- [ ] Confirmed no caller reaches the removed path (including dynamic dispatch)
- [ ] Test coverage asserts the surviving path's environment behaviour
- [ ] No behaviour change for currently-working deployments

## Related

Part of epic #509. Split out of #511, which found the divergence while investigating proxy
support. #511 fixes the environment contents; this issue removes the ambiguity about which
code sets it.

cc @johnvanlowe @krmayankb

samarmstrong:
## Summary

Closes #515.

`MCPServer.start` in `core/integrations/mcp/service.py` still spawned via `subprocess.Popen` with its own `os.environ.copy()`. Nothing called it: service-level `start_server` / `start_all` and `POST /servers/{name}/start` were already removed in #125. Runtime spawn is only `MCPClient.connect_to_server` → `StdioServerParameters`.

The live child already inherits the parent environment. `_initialize_servers` snapshots `os.environ` (plus `ca_bundle_env()`) into `server.env`, and the client forwards that dict. This is not a second #511 — proxy/CA contents stay there.

## Change

- Delete `MCPServer.start`. Confirmed no callers, including getattr/dynamic.
- Leave `get_status` / `is_running` (pgrep) / `test_server` / `stop_server` in place. They still back `services/api/routers/mcp.py` and detection-rules. `self.process` was only set by `start`; `stop()` is already a no-op.
- Do not change `connect_to_server` / `stdio_client` or env policy.

## Tests

- Source guard now treats `_initialize_servers`'s `os.environ.copy()` as the parent snapshot, not a spawn, and keeps `ca_bundle_env()` on that site plus `StdioServerParameters`.
- Asserts the Popen spawn stays gone.
- Asserts the surviving path: parent snapshot in `server.env` is passed into `StdioServerParameters`.

<div><a href="https://cursor.com/agents/bc-30534329-747f-4070-bc64-318344d4b2cf?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-30534329-747f-4070-bc64-318344d4b2cf&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/core/integrations/mcp/service.py
+++ b/core/integrations/mcp/service.py
@@ -83,39 +83,6 @@ class MCPServer:
         # server. Read by mcp_client.connect_to_server at connect time.
         self.required_env_vars: List[str] = list(required_env_vars or [])
 
-    def start(self) -> bool:
-        """Start the MCP server."""
-        if self.process is not None:
-            logger.warning(f"Server {self.name} is already running")
-            return False
-
-        try:
-            # Prepare environment
-            env = os.environ.copy()  # noqa: ENV001 - MCP child process env
-            # httpx ignores REQUESTS_CA_BUNDLE, so inheriting it is not enough.
-            env.update(ca_bundle_env())
-            env.update(self.env)
-
-            # Start process
-            self.process = subprocess.Popen(
-                [self.command] + self.args,
-                cwd=self.cwd,
-                env=env,
-                stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE,
-                text=True,
-            )
-
-            self.status = "running"
-            self.start_time = datetime.now()
-            logger.info(f"Started MCP server: {self.name}")
-            return True
-
-        except Exception as e:
-            logger.error(f"Failed to start MCP server {self.name}: {e}")
-            self.status = "error"
-            return False
-
     def stop(self) -> bool:
         """Stop the MCP server."""
         if self.process is None:
--- a/tests/unit/integrations/test_mcp_child_ca_bundle.py
+++ b/tests/unit/integrations/test_mcp_child_ca_bundle.py
@@ -8,22 +8,27 @@ child environment to a six-name allowlist.
 
 from __future__ import annotations
 
+import json
 from pathlib import Path
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 from mcp.client.stdio import get_default_environment
 
 from core.integrations.mcp.child_env import ca_bundle_env
+from core.integrations.mcp.client import MCPClient
+from core.integrations.mcp.service import MCPServer, MCPService
 
 pytestmark = pytest.mark.unit
 
 ROOT = Path(__file__).resolve().parents[3]
 
-# Where a tool server gets spawned. A new one that forgets the bundle is the
-# failure mode this guards: it works everywhere except behind a private CA.
-SPAWN_MODULES = (
-    "core/integrations/mcp/service.py",
-    "core/integrations/mcp/client.py",
+# Child env is assembled in _initialize_servers (parent snapshot) and spawned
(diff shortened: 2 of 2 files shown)


samarmstrong:

No environment PR. Neither implementer was blocked by a missing tool on the machine.

**[#796](https://github.com/Vigil-SOC/vigil/pull/796)** — dead-code spawn removal. The author ran the child-env / dead-spawn unit tests (36 passed) and flake8 on the touched files. Nothing in the PR or the agent log says a check could not run.

**[#797](https://github.com/Vigil-SOC/vigil/pull/797)** — Elastic Settings wiring. Pytest, flake8/black/isort, `tsc --noEmit`, and eslint all ran. The only skip was:

> I did not boot the full UI, so I could not click the Settings gear in a browser — the map ratchet covers the code path that decides category, description, and the configure button.

That is not a missing-package problem. Docker is already in this environment. They chose not to start the stack because a live Settings click needs a running Vigil (DB, API, frontend, and usually LLM/secrets) that does not exist at review time. The check that belongs here is the one they already added: the registries ratchet that `elastic` maps to `elastic-siem`, sits under SIEM / Data Lake, and exposes the configure control. A screenshot of a booted console would not have caught the `resolve()` / secret-stripping bugs anyway.

Issue #576, opened 2026-08-05
config: harden Settings against malformed env values (fail-fast blast radius, follow-up to #520)


nestor-deeptempo:
Follow-up to #520, which centralized ~215 raw env reads into a single pydantic-settings `Settings` object read via `get_settings()`. Good change. This issue tracks two small hardening cleanups around one behavioral consequence of it. We are **keeping** the fail-fast behavior — this is about making it explicit and removing code that now misleads.

## The behavior

`get_settings()` builds the whole `Settings` object (all ~100 fields) in a single construction and validates every typed field. A single malformed env value — e.g. `DAEMON_BATCH_SIZE=1O` (letter O instead of zero) — raises `pydantic.ValidationError` during that construction. Because `get_settings()` is `lru_cache`d and called from everywhere, the error surfaces at the *first* config read anywhere in the process, regardless of which field the caller actually wanted. A daemon-only typo can therefore fail a backend request path that never reads a daemon setting.

This is a deliberate trade-off and arguably the right default for a 24/7 SOC: you learn about the typo at boot, with a precise message, instead of a subsystem silently running on a wrong value. Two things make it safer and less surprising without giving up fail-fast.

## 1. Remove or convert the now-dead numeric fallbacks

Some sites that used to guard `int(os.getenv(...))` with `try/except ValueError` now catch nothing: the value is pre-validated by pydantic, so the field access returns an `int`, and if the env value *is* bad it raised earlier (as a `ValidationError`, which these clauses don't even catch). Example:

```python
# backend/services/password_reset.py
def _ttl_seconds() -> int:
    try:
        return get_settings().password_reset_ttl_seconds   # already an int
    except ValueError:                                      # can never fire now
        return DEFAULT_TTL_SECONDS
```

Action: audit the `except ValueError` / `except (TypeError, ValueError)` blocks touched by #520 and drop the ones that can no longer fire (e.g. `password_reset._ttl_seconds`). This is a per-site check, not a blanket delete — some must stay. For instance `backend/api/darktrace_webhook.py::_get_max_body_bytes` still wraps a `system_config`/DB-sourced `max_body_kb` value, which can genuinely still be a bad string, so its `try/except` is still live and should be left alone.

## 2. Validate config once at startup, as a labeled config error

[...]

PR #520, opened 2026-07-30
feat(config): typed Settings object; retire the 215 raw env reads

mattmorr1:
Closes #415. Closes #458.

Configuration had no definition site: **215 raw `os.getenv`/`os.environ` reads** spread across `backend/` (78), `daemon/` (84), `services/` (42), `database/` (13), `core/` (14). Anyone editing a call site saw a *consumer* of a setting, never its definition or default. The cost was already visible — `BIFROST_URL` was read in five files with **two different defaults**, and `PROMPT_INJECTION_BLOCK` was parsed three times with two different truthy sets.

**MCP servers keep env as their wire protocol.** `tools/` and `mcp-servers/` sit on the child side of a process boundary; env vars are how *any* MCP server is configured, including third-party ones added tomorrow. So the exemption is by boundary, not per-server adaptation — new servers need zero changes, forever. `mcp_service` is the single exporter, and `_substitute_env_vars` now falls back to `get_secret`, so a credential saved in the UI reaches any server without waiting for a restart to rehydrate `os.environ`.

## Reviewing this

Commit by commit — each is one package and independently revertable:

| | |
|---|---|
| `756108e` | Settings + `vigil_path` + both guards (seeded allow-list) |
| `36d599f` | daemon — 57 reads in `DaemonConfig.from_env`, zero call-site churn |
| `2ffaf5d` | rest of daemon |
| `766853f` | backend — 78 reads |
| `1372c95` | services + core + database; allow-list emptied |
| `a8b0de0` | `~/.vigil` + `env.example` reconciliation |
| `4406c8a` | `.env` anchoring; docs |
| `eaa725a` | orphaned imports (contains bug #1 above) |

`36d599f` also deletes `core/secrets.py` — a dead 496-line divergent duplicate of `backend/secrets_manager.py` with zero importers and a weaker default write backend. It's replaced by a thin re-export so all three packages share one credential channel instead of repeating a `sys.path` dance per module.

PR #800, opened 2026-09-03
Drop the dead numeric fallbacks left by #520

### LINKED ISSUE #642: config: drop the dead numeric fallbacks left by #520
Split out of #576 (part 1 of 2). See #576 for the shared background; the startup-validation half is #643.

## Background

#520 replaced ~215 raw `os.getenv` reads with a typed pydantic-settings `Settings` object read through `core.config.get_settings()`. Numeric fields are now validated and coerced during `Settings` construction, so a field access returns an `int` and a malformed env value raises `pydantic.ValidationError` at construction time — not `ValueError` at the call site.

That leaves some `try/except ValueError` blocks around former `int(os.getenv(...))` calls catching nothing: the happy path can't raise, and the failure path raises a `ValidationError` these clauses don't catch. They read as live defensive code and aren't.

## Action

Drop the guards that can no longer fire. This is a **per-site check, not a blanket delete** — a guard is still live if the value it wraps can come from somewhere other than `Settings` (a `system_config` row, integration config, a request body).

Current candidates, from `grep -rn -B6 'except ValueError\|except (TypeError, ValueError)' --include='*.py' core services | grep get_settings`:

| Site | Verdict |
|---|---|
| `core/auth/password_reset.py:43` (`_ttl_seconds`) | **Remove.** The sole value is `get_settings().password_reset_ttl_seconds`, already an `int`. `DEFAULT_TTL_SECONDS` becomes unused at that site — check whether anything else references it before deleting the constant. |
| `core/integrations/darktrace/darktrace_webhook_router.py:81` (`_get_max_body_bytes`) | **Keep.** Wraps `settings.get("max_body_kb")` from `system_config`, which can genuinely still be a bad string; the `Settings` field is only the fallback. |
| `services/daemon/threat_feed_poller.py:53` (`poll_interval_seconds`) | **Keep.** `raw` comes from `get_integration_config("cloudforce_one")` first and only falls back to `get_settings().threat_feed_poll_interval`, so the same reasoning as darktrace applies. |

Confirm the list rather than trusting it — the grep above only finds guards within 6 lines of a `get_settings()` call, so a wider-spaced one would be missed.

## Notes

Paths in #576 were written pre-reorg and referenced `backend/services/password_reset.py` / `backend/api/darktrace_webhook.py`. Both now live under `core/`; the table above has the current locations.

## Verification

`pytest tests/unit/auth/test_password_reset.py tests/unit/integrations/test_darktrace_webhook.py` should pass unchanged — no behavioral change is intended, only the removal of unreachable branches.

## Out of scope

Adding per-field coerce-or-default validators to restore fail-soft numeric parsing. Considered and declined in #576 — the fail-fast behavior is deliberate.

samarmstrong:
Closes #642

## What

`get_settings()` builds a pydantic `Settings` object. `pydantic.ValidationError` subclasses `ValueError`, so `_ttl_seconds` in `core/auth/password_reset.py` swallowed a construction error for **any** field on a cache miss and returned `3600`. That is the fail-soft path #576 declined.

This PR strips that `try/except` and deletes `DEFAULT_TTL_SECONDS` (only used by the except branch). `Settings.password_reset_ttl_seconds` already defaults to 3600. `_ttl_seconds` stays so `tests/unit/auth/test_password_reset.py` can still monkeypatch TTL.

Startup already labels construction errors via `validate_settings_or_exit()` (#643). This does not retarget the catch at `ValidationError`.

## Left alone (still live guards)

- `darktrace_webhook_router._get_max_body_bytes` — `int()`s `system_config` `max_body_kb` first; Settings is only the fallback.
- `threat_feed_poller.poll_interval_seconds` — reads `get_integration_config("cloudforce_one")` first.
- Other `except ValueError` / `(TypeError, ValueError)` sites that wrap request bodies, STIX, Redis, JSON, UUIDs, integration rows, or `system_config`.

The Cloudy receiver already reads `cloudy_webhook_max_body_kb` as an `int` with no catch.

Out of scope: per-field coerce-or-default validators.

## Verification

- `pytest tests/unit/auth/test_password_reset.py tests/unit/integrations/test_darktrace_webhook.py` — 24 passed
- flake8 / black / isort clean on `core/auth/password_reset.py`
- Direct check: `_ttl_seconds()` still returns `3600` on valid Settings; a malformed `POSTGRES_PORT` now raises `ValidationError` instead of returning `3600`
- Wider grep of `core/` and `services/` for Settings-only numeric try/excepts; only `_ttl_seconds` was Settings-only

<div><a href="https://cursor.com/agents/bc-cd0b860b-15b0-40fc-b619-e615c2aaba69?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-cd0b860b-15b0-40fc-b619-e615c2aaba69&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/core/auth/password_reset.py
+++ b/core/auth/password_reset.py
@@ -33,16 +33,12 @@ logger = logging.getLogger(__name__)
 
 
 RESET_TOKEN_PURPOSE = "vigil-password-reset"
-DEFAULT_TTL_SECONDS = 3600  # 1 hour
 
 _USED_TOKEN_PREFIX = "password_reset_used:"
 
 
 def _ttl_seconds() -> int:
-    try:
-        return get_settings().password_reset_ttl_seconds
-    except ValueError:
-        return DEFAULT_TTL_SECONDS
+    return get_settings().password_reset_ttl_seconds
 
 
 def _get_serializer() -> URLSafeTimedSerializer:


samarmstrong:

No environment change. Neither PR reported a check the author could not run because the machine was missing something, so I did not open an env PR.

**[#800](https://github.com/Vigil-SOC/vigil/pull/800)** (`password_reset` TTL) lists completed verification: 24 unit tests, flake8/black/isort, and a direct `ValidationError` check. Pytest was not on PATH at boot; the author provisioned the project venv in-session and ran those checks. That is expected, not a gap to bake into the image.

**[#801](https://github.com/Vigil-SOC/vigil/pull/801)** (MCP pins) lists a completed test plan: 26 ratchet tests, a live-config placeholder scan, and green CI. The ratchet is stdlib-only; they ran it in a throwaway venv. Related app tests were skipped as too slow to install, not because a tool was absent. CI later ran the backend unit suite.

The image already has Docker (with `fuse-overlayfs` / `iptables-legacy`), Helm, Python 3, and Node. Preinstalling `requirements.lock` would slow every future agent for checks these authors either completed or chose not to wait for. Both PRs are merged with CI green (Docker image jobs skipped on PRs, as usual).

PR #842, opened 2026-09-09
feat(web): integrate VStrike and improve findings triage

ShmalexM:
A finding with a source and destination IPv4 pair now has an inline **VStrike** action that opens external network context and preserves the analyst's place in the findings queue. This replaces the Dashboard's unused Entity Graph placeholder with a configurable VStrike view and improves the surrounding triage workflow.

## Changes

- Show the VStrike tab, inline finding/detail actions, and events entry point only after integration settings have loaded with VStrike enabled. Disabling it closes active graph/events, stops retries/polling, ignores late responses, and restores the preserved findings queue.
- Add a lazy VStrike iframe with account-provided network/storyline choices, apply and step controls, reconnect, expanded view, bounded initialization retry, and guards against stale responses. The requested endpoints stay visible with retry and return actions.
- Put external scenario events in an on-demand drawer. Reads are bounded, refresh only while open, label incomplete pages, retain the last page on refresh failure, and keep source events separate from Vigil findings and model scores.
- Improve findings with readable titles, source/destination context, inline actions, separate searchable column controls, visible filter chips, working KPI shortcuts, the correct active-case count, and a toolbar Ask Vigil action. Search, filters, sort, pagination, column preferences, and scroll survive graph/detail navigation; CSV export is preserved.
- Display finding/evidence times consistently in UTC, retain missing scores/timestamps, and order evidence chronologically without mutating the source records.
- Add authenticated selection and bounded event-read endpoints, align storyline/VCR calls with the provider contract, regenerate API types, and allow only the configured provider origin in the default document `frame-src`. Custom CSP policies remain authoritative.

## Validation

- Web production build passes; **220 tests across 24 files** pass. ESLint has zero errors and 18 existing warnings.
- **149 focused backend tests** pass, covering VStrike service/UI contracts, new validation/auth behavior, CSP, and router discovery. Python formatting/lint and all three import-layer contracts pass.
- Seven additional integration-availability regressions cover disabled/loading/failed settings, enabling, and disabling during an active graph/event read.
- Browser walkthrough with an isolated synthetic provider verifies hidden dashboard/detail controls and zero provider requests while disabled, restored actions when enabled, desktop/640px layouts, modal dismissal and focus restoration, queue-state preservation, inline selection, reconnect, and the events drawer. No browser error/warning logs were observed.
- Secret/privacy and generated-artifact review completed. Only generic code and synthetic test fixtures are included.

## Provider limits and coordination

An accepted selection command does not establish that VStrike rendered the requested highlighting or zoom. The UI explicitly reports **selection requested, highlighting/zoom unconfirmed**. Live-provider camera framing, exact matching, and path highlighting still require provider-side validation; this PR does not claim to fix them. The inline finding action currently requires one unambiguous IPv4 address per endpoint.

There is overlap with #838, which removes the old Entity Graph stub. This PR introduces a named external VStrike view and does not depend on the legacy entity-graph API. Coordinate the Dashboard, environment comments, and generated schema changes when merging either PR.

### DIFF
--- a/clients/web/src/data/data.ts
+++ b/clients/web/src/data/data.ts
@@ -32,6 +32,9 @@ export const NAV: [IconName, string, ConsoleScreenKey | null, NavGate?][] = [
 
 export interface Finding {
   id: string
+  title?: string
+  sourceIp?: string
+  destinationIp?: string
   sev: 'Critical' | 'High' | 'Medium' | 'Low' | 'Unrated'
   tech: string
   conf: number
--- a/clients/web/src/data/findingPresentation.test.ts
+++ b/clients/web/src/data/findingPresentation.test.ts
@@ -1,0 +1,29 @@
+import { describe, expect, it } from 'vitest'
+import { findingEndpoints, findingTitle, sourceTime, sourceTimestamp } from './findingPresentation'
+import { mapApiFinding } from './mappers'
+
+describe('finding presentation', () => {
+  it('maps explicit endpoints before flattening metadata and rejects ambiguous or invalid roles', () => {
+    expect(findingEndpoints({ source_ip: '192.0.2.2', src_ips: ['192.0.2.2'], dest_ips: ['198.51.100.10'] }))
+      .toEqual({ sourceIp: '192.0.2.2', destinationIp: '198.51.100.10' })
+    for (const source of [['192.0.2.2', '192.0.2.20'], ['192.0.2.2', 'invalid'], '192.0.2.2/32', '2001:db8::1', '192.000.2.2']) {
+      expect(findingEndpoints({ source_ips: source }).sourceIp).toBeUndefined()
+    }
+    expect(findingEndpoints({ ips: ['192.0.2.2'], nodes: [{ source_ip: '192.0.2.2' }] })).toEqual({ sourceIp: undefined, destinationIp: undefined })
+  })
+  it('keeps source titles and missing scores instead of inventing detection claims', () => {
+    const finding = mapApiFinding({ finding_id: 'synthetic-1', title: 'Unusual transfer', entity_context: { source_ip: '192.0.2.2', destination_ip: '198.51.100.10' } })
+    expect(findingTitle(finding)).toBe('Unusual transfer')
+    expect(finding.score).toBeNull()
+    expect(finding.sev).toBe('Unrated')
+    expect(finding.sourceIp).toBe('192.0.2.2')
+    expect(findingTitle({ src: 'flow', tech: '—' })).toBe('Network flow finding')
+  })
+  it('interprets naive timestamps as UTC and retains explicit offsets', () => {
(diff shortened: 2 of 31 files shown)


samarmstrong:

The Dashboard no longer has an Entity Graph tab for this to replace. `4fd9ac1` (#838) removed `EntityStub` from `clients/web/src/screens/dashboard/DashboardScreen.tsx`; the tabs are Findings, ATT&CK, and Timeline, and `clients/web/src/shell/SocConsole.test.tsx` covers those three. `env.example` no longer says a VStrike iframe stands in for that tab. `clients/web/src/screens/dashboard/FindingPopup.tsx` still opens with the note that the embedded VStrike panel is deliberately not ported, because that provider is not mounted in the console shell. This branch’s `DashboardScreen.tsx` diff still deletes `EntityStub` and the `entity` tab, and the `env.example` edit is written against the sentences #838 already deleted, so the merge conflicts in `DashboardScreen.tsx`, `SocConsole.test.tsx`, `env.example`, and `clients/web/src/services/generated/schema.d.ts`.

The same diff rebuilds the findings queue around that missing tab. Search, filters, the column picker, persisted preferences, CSV export, and the open/investigating case counts are already in `DashboardScreen.tsx`, `findingsColumns.tsx`, and `findingsPreferences.ts`. Ask Vigil is the console control in `clients/web/src/shell/SocConsole.tsx`. This change hides the MITRE, tactic, host, and user columns by default, rewrites finding titles and UTC formatting in `clients/web/src/data/mappers.ts`, turns the KPI row into shortcuts, moves columns into a new drawer, and hides the Ask Vigil button on the dashboard. That is a second product change, tangled with the embed, and putting it on the current console is a rewrite.

The embed also adds a second reader for a tool the API already proxies. `VStrikeService.storyline_events_get` calls MCP `storyline-events-get` with `storylineId` and is exposed as `POST /api/integrations/vstrike/storylines/events` (`core/integrations/vstrike/client.py`, `services/api/routers/vstrike.py`). This adds `storyline_faults_page` and `POST /faults/query`, calling the same tool with `storylineSetId`, and changes `ui_storyline_apply` to `storylineSetId` while leaving `storyline_events_get` on `storylineId`. The client marks those parameter names as undocumented. `services/api/middleware/security_headers.py` would also append the provider origin to the default `frame-src` on every HTML response. A vendor iframe, a custom `vstrike:state` handshake, storyline VCR controls, and a polling events drawer are a second analyst workstation beside the MCP control plane that is already there.

A PR worth merging would do one thing on the current dashboard: with VStrike enabled, a finding that has one source IPv4 and one destination IPv4 opens the existing UI routes (`/ui/iframe-token`, `/ui/load-network`, `/ui/storyline-apply`, `/storylines/events`). It would leave the findings queue, `mapApiFinding`, column defaults, and the Ask Vigil control alone, and it would not add `/faults/query`. If the provider really requires `storylineSetId`, that is a separate change to `ui_storyline_apply` and `storyline_events_get` with `tests/unit/integrations/test_vstrike_service.py` updated to say why `storylineId` is wrong.

PR #857, opened 2026-09-11
fix(web): don't label every Bifrost key "Disabled"

antoine-leclef:
Every key in the AI Providers panel renders with a **"Disabled"** badge, regardless of its actual state.

## Cause

`AiProvidersPanel.tsx` shows the badge when `!k.enabled`, but Bifrost omits `enabled` from its key **read** shape entirely, so the expression is true for every key the gateway returns.

This is not an `omitempty` elision. The same response serialises `use_for_batch_api: false` and `use_anthropic_endpoints: false`, so `false` values *are* emitted — the field simply is not in that response struct. `BifrostKey.enabled` being typed as a required `boolean` on read is what hid it.

The editor already half-knew this: `KeyDialog` defaults with `existing?.enabled ?? true`, which is only meaningful if the field can be absent.

## Fix

- Test `k.enabled === false`, so only an explicit disable shows the badge.
- Mark `BifrostKey.enabled` optional, so the type says what the gateway actually sends. (The write shape keeps it required — a write does carry it.)

Three lines of production change.

## Relationship to #831

#831 fixed the *routing* half of this same misreading, moving the verdict server-side behind `GET /api/bifrost/routability`. The badge reads the absent field directly and was left behind.

## Tests

Two cases added to `AiProvidersPanel.test.tsx`: the badge stays hidden for a key returned without `enabled`, and still shows for one explicitly disabled.

Verified the first genuinely catches the bug — reverting the one-line change makes it fail and leaves the rest of the suite green. Full file: 7/7 passing, `tsc --noEmit` clean.

---

Found while running Vigil against an Azure AI Foundry-backed Bifrost, where the only provider is seeded from `config.json`. Happy to adjust naming or split the type change out if you would rather keep the two separate.

### DIFF
--- a/clients/web/src/screens/settings/AiProvidersPanel.test.tsx
+++ b/clients/web/src/screens/settings/AiProvidersPanel.test.tsx
@@ -147,3 +147,31 @@ describe('the models a key is fenced to', () => {
     expect(payload.value).toBeUndefined()
   })
 })
+
+/* Bifrost omits `enabled` from its read shape entirely. That is not an
+   omitempty elision -- the same payload serialises `use_for_batch_api: false`
+   and `use_anthropic_endpoints: false` -- so `!k.enabled` was true for every
+   key the gateway returned, and every key rendered as "Disabled". */
+describe('the “Disabled” badge', () => {
+  const expand = async () => {
+    fireEvent.click(screen.getByTitle('Expand'))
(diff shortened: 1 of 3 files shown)


nestor-deeptempo (approved):
lgtm! thanks for your contribution!

PR #858, opened 2026-09-11
fix(integrations): Azure Sentinel descriptor/config never matched what ingestion reads

antoine-leclef:
## The bug

Three registries have to agree for the Azure Sentinel integration to work at all:

1. `core/integrations/azure_sentinel/descriptor.py` — declares the config fields the Settings UI collects.
2. `clients/web/src/config/integrations.ts` — the Settings form itself.
3. `core/integrations/azure_sentinel/ingestion.py` — what `fetch_alerts` actually reads.

Before this PR, (1) and (2) agreed with each other (`workspace_id, tenant_id, client_id, client_secret`), but not with (3): `fetch_alerts` reads `tenant_id, client_id, client_secret, subscription_id, resource_group, workspace_name` and only proceeds if `all([...])` of those six is truthy.

Three of those six — `subscription_id`, `resource_group`, `workspace_name` — had no way to ever be populated, since neither the descriptor nor the Settings form offered them. The `all([...])` guard could therefore never pass, and `fetch_alerts` always logged "Azure Sentinel configuration incomplete" and returned `[]`, no matter how correctly an operator filled in the Settings UI. **Azure Sentinel ingestion has never worked upstream.**

`workspace_id` (the Log Analytics workspace GUID) isn't one of the values `fetch_alerts` reads at all — `SecurityInsights.incidents.list` is scoped by `subscription_id` + `resource_group` + `workspace_name`, not the GUID — so this PR drops it rather than adding a fourth field nobody reads.

### Fix
- `descriptor.py` now declares the six fields ingestion actually reads.
- `integrations.ts` now collects those same six fields, so the descriptor and the Settings form agree with ingestion for the first time.

### A related, smaller bug in the same file

`except ImportError: return []` in `fetch_alerts` made "azure-identity / azure-mgmt-securityinsight was never installed" indistinguishable from "there were no incidents this poll" — both silently produced `[]`. This PR raises a `RuntimeError` with an install hint instead. `SIEMIngestionAdapter` already isolates per-source exceptions, so a missing SDK on one source can't take the rest of the poller down; it just stops being silent about it.

Happy to split this into two PRs (the registry fix vs. the `ImportError` behavior change) if you'd prefer to review them separately — let me know.

## Testing

- `tests/unit/_ratchets/test_integration_registries.py` — 8 passed (this is the ratchet that enforces descriptor/catalog agreement; it now passes for `azure-sentinel` for the first time)
- Added `tests/unit/integrations/test_azure_sentinel_ingestion.py` — 3 passed. Covers (a) every field the completeness guard checks is declared on the descriptor, as a regression test for the drift described above, and (b) a missing SDK raises instead of returning `[]`.
- `tests/unit/integrations/test_integration_secrets.py`, `test_integration_id_hyphens.py` — 31 passed, unaffected by this change.
- `clients/web`: `npx tsc --noEmit` — clean.

Found while running Vigil against Azure and noticing Sentinel incidents never showed up regardless of configuration.

Related: #857 (independent, a Bifrost "Disabled" badge fix from us also open against this repo).

### DIFF
--- a/clients/web/src/config/integrations.ts
+++ b/clients/web/src/config/integrations.ts
@@ -920,12 +920,6 @@ export const INTEGRATIONS: IntegrationMetadata[] = [
     functionality_type: 'Log Analysis & SIEM',
     fields: [
       {
-        name: 'workspace_id',
-        label: 'Workspace ID',
-        type: 'text',
-        required: true,
-      },
-      {
         name: 'tenant_id',
         label: 'Tenant ID',
         type: 'text',
@@ -942,6 +936,27 @@ export const INTEGRATIONS: IntegrationMetadata[] = [
         label: 'Client Secret',
         type: 'password',
         required: true,
+      },
+      {
+        name: 'subscription_id',
+        label: 'Subscription ID',
+        type: 'text',
+        required: true,
+        placeholder: 'Azure subscription containing the Sentinel workspace',
+      },
+      {
+        name: 'resource_group',
+        label: 'Resource Group',
+        type: 'text',
+        required: true,
+        placeholder: 'Resource group of the Log Analytics workspace',
+      },
+      {
+        name: 'workspace_name',
+        label: 'Workspace Name',
+        type: 'text',
+        required: true,
+        placeholder: 'Log Analytics workspace name (not its GUID)',
       },
     ],
     docs_url: 'https://docs.microsoft.com/en-us/azure/sentinel/',
--- a/core/integrations/azure_sentinel/descriptor.py
+++ b/core/integrations/azure_sentinel/descriptor.py
@@ -1,4 +1,17 @@
-"""Azure Sentinel descriptor — source of truth for its registry entries."""
+"""Azure Sentinel descriptor — source of truth for its registry entries.
+
+The field set below must match what ``ingestion.py`` actually reads: it used
+to declare ``workspace_id, tenant_id, client_id, client_secret`` while
+``fetch_alerts`` read ``tenant_id, client_id, client_secret, subscription_id,
+resource_group, workspace_name`` and guarded on ``all([...])`` of those six.
+Three of the six had no way to be populated through the Settings UI, so the
+guard could never pass and ``fetch_alerts`` always returned ``[]`` — Azure
+Sentinel ingestion could not work at all. ``subscription_id``,
+``resource_group`` and ``workspace_name`` identify the Log Analytics workspace
+that ``SecurityInsights.incidents.list`` is scoped to; ``workspace_id`` (the
+workspace GUID) was never one of the values that call takes, so it is dropped
+rather than carried along unused.
+"""
 
 from core.integrations._base.descriptor import (
     IntegrationDescriptor,
@@ -12,10 +25,12 @@ AZURE_SENTINEL = register_descriptor(
         category="SIEM",
         mcp_server_names=("azure-sentinel",),
         fields=(
-            IntegrationField("workspace_id"),
             IntegrationField("tenant_id"),
             IntegrationField("client_id"),
             IntegrationField("client_secret", secret=True),
+            IntegrationField("subscription_id"),
+            IntegrationField("resource_group"),
+            IntegrationField("workspace_name"),
         ),
     )
 )
--- a/core/integrations/azure_sentinel/ingestion.py
+++ b/core/integrations/azure_sentinel/ingestion.py
@@ -42,10 +42,25 @@ class AzureSentinelIngestion(SIEMIngestionService):
         Returns:
             List of raw incident dictionaries
         """
+        # A missing Azure SDK used to be swallowed by the blanket
+        # `except ImportError: return []` below, which made "azure-identity was
+        # never installed" indistinguishable from "there were no incidents".
+        # Raise here instead, with an install hint, so the failure is visible.
+        # This import happens outside the try/except Exception block below so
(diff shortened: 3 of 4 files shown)


samarmstrong:
🏭 **Factory review — left for a human.** The review verified 2761174c, but the head is now 200f3771, so nothing was merged; the factory reviews it again as it stands.

The field mismatch is real, and the descriptor and Settings form now collect the six values `fetch_alerts` checks. I kept that change and pushed one commit on top, `2761174`, because a filled-in form still could not produce an incident.

`client_secret` is stripped into the secret store before the JSON config is written, and `fetch_alerts` was reading `get_integration_config`, so the completeness guard stayed false. It now uses `resolve(AZURE_SENTINEL)`, the same path Elastic uses. The incident mapper also read names the current SecurityInsights `Incident` model does not have (`last_updated_time_utc`, `additional_data.alert_count`, `additional_properties`). Those raise `AttributeError`, and the existing handler turned that into `[]`. It now reads `last_modified_time_utc` and `alerts_count`, and it compares timestamps as naive UTC so an aware `created_time_utc` does not throw. An independent review found nothing further to change. The broad `except Exception: return []` around the Azure call is the existing handler for an auth or HTTP failure, and the mapper no longer throws on a real incident.

**Ran**
- Backend lint: flake8, black, isort, import-linter, and the one-loop gate. mypy reports existing errors in other modules; none are in these files, and CI does not fail the job on mypy.
- `pytest tests/unit/ tests/security/ -m "not external_service"`: 2647 passed, 32 skipped.
- `pytest tests/unit/ -m external_service`: 207 passed, 3 skipped, 2 xfailed.
- `pytest tests/integration/`: 129 passed, 2 skipped. Those two skip unless `BIFROST_URL` is set. With the local gateway, the chat test fails because Ollama is not running here.
- Web: eslint, `tsc --noEmit`, vitest (249 passed), and `npm run build`.
- Agent layer: `tsc --noEmit` and vitest (976 passed).
- Live: Settings → Integrations → Azure Sentinel. Saved a tenant, client, secret, subscription, resource group, and workspace name. The reopened dialog showed the non-secret values, and the secret field showed that it was stored. The JSON config has those five fields and no `client_secret` or `workspace_id`. `resolve()` returned the secret. `fetch_alerts` with that config and a stand-in client called `incidents.list` for `rg-sentinel` / `law-sentinel` and mapped the incident. With the Azure SDK absent, `ingest_alerts` returns `success: false` and the install hint.

**Did not run:** a call to a real Sentinel workspace. The SDK is not installed here, and there is no Azure tenant.

Push to the branch or reply here and the factory reviews it again.

samarmstrong:
🏭 **Factory review — left for a human.** The review verified 6c568229, but the head is now 200f3771, so nothing was merged; the factory reviews it again as it stands.

The field mismatch is real, and the descriptor and Settings form now collect the six values `fetch_alerts` checks. I kept that and pushed two commits on top, because a filled-in form still could not produce an incident.

`client_secret` is stripped into the secret store before the JSON config is written, and `fetch_alerts` was reading `get_integration_config`, so the completeness guard stayed false. It now uses `resolve(AZURE_SENTINEL)`, the same path Elastic uses. The incident mapper also read names the current Security Insights `Incident` model does not have (`last_updated_time_utc`, `additional_data.alert_count`, `additional_properties`). Those raise `AttributeError`, and the existing handler turned that into `[]`. It now reads `last_modified_time_utc` and `alerts_count`, compares timestamps as naive UTC, and leaves `properties` empty because that model has no extra-properties bag. The legacy poller then called `ingest_alerts` on the daemon loop; `asyncio.run()` raises there, so every tick failed before `fetch_alerts`. That call now runs on a worker thread, the same way the Splunk poll leaves the loop. An independent review of the fixes found that poller bug; a follow-up look found nothing else.

**Ran**
- Backend lint: flake8, black, isort, import-linter, and the one-loop gate. mypy is continue-on-error in CI; the full run stops on a pre-existing duplicate-module error in `services/api/middleware/auth.py`, and these files are not part of it.
- `pytest tests/unit/ tests/security/ -m "not external_service"`: 2082 passed, 32 skipped.
- `pytest tests/unit/ -m external_service`: 147 passed, 3 skipped, 2 xfailed.
- `pytest tests/integration/`: 126 passed, 2 skipped. Those two skip unless `BIFROST_URL` is set.
- Web: eslint (0 errors), vitest (195 passed), and `npm run build`.
- Agent layer: `tsc` via `npm run typecheck`, and vitest (951 passed).
- Live: Settings → Integrations → Azure Sentinel. Saved a tenant, client, secret, subscription, resource group, and workspace name. The reopened dialog showed the five non-secret values, and the secret field’s placeholder was “saved — leave blank to keep.” The JSON config has those five fields and no `client_secret` or `workspace_id`. `resolve()` returned the secret. `fetch_alerts` with that config and a stand-in client called `incidents.list` for `rg-sentinel` / `law-sentinel` and mapped the incident. With the Azure SDK absent, `ingest_alerts` returns `success: false` and the install hint.

**Did not run:** a call to a real Sentinel workspace. The SDK is not installed here, and there is no Azure tenant.

Push to the branch or reply here and the factory reviews it again.

samarmstrong:

Nothing was reviewed: it is 133 commits behind the default branch, and the factory reviews a PR that is within 100 commits of the default branch and merges cleanly. Rebase it onto the default branch and open a new PR — GitHub cannot reopen a closed PR whose branch was force-pushed — and it is reviewed as it stands.

Issue #912, opened 2026-09-15
Hunt preflight (cost/capabilities/budgets) rides on the workflow catalog endpoint


G-r-ay:
`GET /workflows/{workflow_id}` returns the workflow definition, but for hunt-kind workflows it also attaches `capabilities`, `pricing`, and `budgets`:

```python
if _is_hunt(service, workflow_id):
    workflow["capabilities"] = _capabilities(registry)
    workflow["pricing"]      = _pricing()
    workflow["budgets"]      = {max_iterations, max_cost_usd}
```

These three describe *executing* a hunt (LLM cost, bound tools, iteration/spend limits) — agent-run concerns — not the workflow definition the catalog endpoint is for. They ride on the catalog route to feed the console's start-a-hunt modal. Two consequences:

- The response shape varies by workflow kind (hunts carry three extra keys), which is awkward for a route now in the frozen `/api/v1` surface.
- Run/agent preflight data is coupled to the catalog read, rather than living with the run surface (`/api/v1/agent-runs`) or a dedicated preflight endpoint.

## Proposal

Move the hunt preflight (cost/capabilities/budgets) off `GET /workflows/{id}` to a run-side or dedicated preflight endpoint the console calls before starting a hunt; the catalog read returns just the definition (uniform shape).

## Not urgent

`GET /workflows/{id}` has no `response_model`, so the contract snapshot pins the operation, not these fields — they can be moved later without breaking the frozen contract. Do this when console work is already in scope.

samarmstrong:
🏭 **Factory groom — needs work, blocked on #913.** The factory reviews this again once that issue closes. Edit the description or reply here and it gets reviewed again on a later tick.

I read the issue, the router it points at, the console consumer, and the open PR that gives the issue its premise.

**What the code says today (`5a7ce0f`, main)**

- `core/workflows/workflows_router.py:310-334` — `GET /api/workflows/{workflow_id}` attaches `capabilities`/`pricing`/`budgets` when `_is_hunt(...)`. The helpers `_hunt_defaults`, `_capabilities`, `_pricing` live at lines 276-307 of the same file.
[...]

Issue #881, opened 2026-09-14
Epic: a frozen contract: versioned API, one MCP server, auth on by default


samarmstrong:
From "Vigil 1.0", Revision 3, 9 September 2026 (internal). This issue carries what bears on it. Absorbs #860.

**What becomes possible.** A customer, a contributor, or DeepTempo's own platform can build against Vigil without reading its code and without breaking at the next minor release. Today the HTTP API is unversioned and does not separate a stable surface from console wiring, `DEV_MODE` is on by default, and MCP tool names are not a commitment. 1.0 is defined as the freeze of a small set of surfaces, and the promise that nothing moves out of the open product.

**Behaviour the author expects.**
- A versioned API under one prefix for findings, cases, verdicts, workflows, approvals, runs, and learning, described by a published OpenAPI document; console-only routes are visibly outside it.
- One MCP server with stable tool names; a tool name in the frozen set does not change or disappear within a major version.
- Frozen schemas for Finding and Source Evidence, workflow files, intent files, and skill files, published as the specification others implement against.
- Authentication on by default; a fresh install requires it; the development bypass is opt-in and loud.
- A written versioning policy: what is frozen, how it changes at a major, and that the direction of movement is one way, into the open product.

**Decisions to settle.**
- Exactly which routes and tool names are in the frozen set. Author's list is above; the groomer settles what the code can honour now and what is deferred.
- Whether the schemas are published as JSON Schema alongside the API document. Author's reason: the platform and contributors need a spec, not a code reference.
- How console wiring is distinguished: a separate prefix, a separate router, or documentation only. No stated preference.

**Where it pulls against the principles.** A freeze is a commitment that constrains every later PR; the record should be explicit about the cost of changing a frozen surface so children do not casually widen it. Least machinery: versioning is a prefix and a policy, not a compatibility layer.

**Human-only decisions.** The freeze itself is a product commitment that is expensive to walk back once external consumers exist. The groomer should propose the set; a maintainer confirms it before the epic is marked groomed.

G-r-ay:
Progress map for this epic — it bundles several deliverables; here's where each stands:

- **Versioned API surface** → PR #913. Delivers `/api/v1` for findings, cases, approvals, workflows, agent-runs (with the new `list_runs`), and cases/metrics, dual-mounted, with a committed contract snapshot + drift test. Peer-reviewed (two-axis). Does **not** close this epic.
- **One MCP server / frozen tool names** → open.
[...]

PR #913, opened 2026-09-15
feat(api): versioned /api/v1 contract surface (findings, cases, approvals, workflows, runs, metrics)

G-r-ay:
Part of the frozen-contract epic (#881). This is the **versioned-API slice** — it does not close the epic: verdicts, learning, and analytics are deferred, and the SPA is not yet ported off `/api/*`.

## What this delivers

A versioned HTTP contract, `/api/v1/**`, that a caller other than the web console can build on. Contract routes move into a new `core/api/v1/` package; console wiring stays unversioned at `/api/*`. Everything is dual-mounted, so existing callers keep working.

Resources now under `/api/v1` (all dual-mounted at their old paths too):

| Resource | Frozen |
|---|---|
| findings | 5 (list, get, summary, export, patch) |
| approvals | 5 (whole router) |
| cases | 16 (record, lifecycle, findings-link, evidence, IOCs) |
| workflows | 2 (catalog: list, get) |
| agent-runs | 4 — whole move **plus the new `GET /api/v1/agent-runs` (`list_runs`)** |
| cases/metrics | 6 frozen / 6 `x-vigil-beta` |

## Key decisions (the "why")

- **Promote a subset, not rename.** Freezing all ~330 routes would trap the console redesign; only durable-record routes are frozen, decided per route.
- **Membership by file, not a flag.** Contract routes live in `core/api/v1/` — a set the snapshot can pin and a reviewer can see in the tree. `RouterMeta` is not given a `surface=` flag (can't split a router; second model of route identity).
- **Dual-mount** via `RouterMeta.legacy_prefixes`: one handler set at both `/api/v1/<res>` and `/api/<res>`. The console's calls are untouched; it ports later. Sunset: deprecate once the SPA is ported, remove no earlier than 2.0.
- **`runs` = agent runs** (`/api/agent-runs`), not workflow-orchestration runs and not the `/internal/runs` harness write-back. A hunt is both a workflow and an agent run joined by `run_id`; freezing the workflow-run surface too would promise two overlapping shapes.
- **Reporting stays out** — it is becoming a skill over the data layer; `generate-report` returns a server file path that is useless to an external caller.
- **Beta tier** (`x-vigil-beta`): a route is versioned but its shape is not pinned. Used for the 6 composite metrics rollups whose shape will churn.
- **Two detail reads left unpinned on purpose** (`GET /findings/{id}`, `GET /workflows/{id}`): a strict `response_model` would strip the normalized finding / the hunt-preflight fields; their shape is owned by the domain schema. The snapshot still pins the operations.

## Enforcement

`core/api/v1/contract.snapshot.json` pins the `/api/v1/**` paths and response shapes (excluding beta). `scripts/generate_api_v1_contract.py` builds it; `tests/unit/api/test_api_v1_contract.py` fails CI on drift. Changing the contract means regenerating and committing the diff — a deliberate, reviewable act. Narrower than the whole-spec `schema.d.ts` (which must stay free to move for the console) and stricter than the prefix-only stability test.

## Deferred / follow-ups

- verdicts, learning — no routes exist yet; recorded as deferred in `core/api/v1/README.md`.
- analytics — left unversioned (shape will churn).
- Port the SPA from `/api/*` to `/api/v1/*`, then deprecate the legacy aliases.
- Move hunt-preflight (cost/capabilities/budgets) off the workflow catalog endpoint (#912).

## Review

Two-axis code review (peer) against #881; the serious finding — half the frozen surface pinned empty response schemas — is fixed (empty-schema frozen ops 17 → 2). Tests green: 492 unit (api/cases/agents/workflows/ratchets), 52 auth with `DEV_MODE=false`, contract + dual-mount, `lint-imports` 3 kept / 0 broken.

Docs: `core/api/v1/README.md` (recipe + internal/external test + enforcement) and a new `CONTEXT.md` section.


G-r-ay:
Addressed — pushed 2 commits.

**Blocker fixed.** All five collection roots freeze bare now (`@router.get("")`), matching README:30 and the two that already were; snapshot regenerated, plus the console's three call sites and the generated schema.

**The catch-all was worse than the three collections.** `return {...}, 404` is a Flask idiom, so FastAPI serialized the tuple as a 2-element JSON array and kept the 200 — for *every* missing `/api/**` path, not just the frozen ones. Fixed to a real `JSONResponse(404)`. Two follow-on catches, both from curling a live server with a frontend build on disk (the arrangement CI never has):

- answering the 404 *inside* the catch-all still meant the route matched, so `/api/findings/` and `/api/cases/` — the legacy spelling the console uses — went from working to a hard 404. Rewrote it as a 404 exception handler that runs *after* routing fails, so `redirect_slashes` sends the slash form to the route again.
- scoped to backend-owned prefixes, because a missing `/static/*` bundle was being served the app shell (the "disallowed MIME type" failure the `/assets` mount exists to prevent).

`test_unauth_endpoints.py` had the frozen paths written with trailing slashes and was passing only because the client followed the 307 — repointed at the bare form.

**README** now says collection paths carry no trailing slash and why, and that `DELETE` on a case is deliberately unversioned (rule 4 beats rule 1: destroying an audit record isn't a shape to freeze for a 1.x lifetime). The dead `setdefault("DEV_MODE")` is gone.

Verified against a running server: five collections answer, both legacy spellings reach the route, an `/api` miss is a 404 with a body that says so, the app shell still answers a client-side route.

nestor-deeptempo (approved):

Issue #893, opened 2026-09-14
obs: Prometheus scrape serves the OTEL registry, not a second one


samarmstrong:
Part of #887

Part of #887

Coding: small · Vigil context: `core/telemetry.py`, `core/platform/monitoring.py`, scrape config · Good for: factory

**Context.** The epic settled one export: Prometheus scrape of the OpenTelemetry MeterProvider. That is not what runs today. `core/telemetry.py` `_do_init` attaches `PrometheusMetricReader` and logs "on port 9090"; nothing starts an HTTP server. Backend `GET /metrics` in `services/api/main.py` returns `prometheus_client.generate_latest()` from a second registry in `core/platform/monitoring.py` (`PrometheusMiddleware`, `http_requests_total`). FastAPI HTTP metrics already come from `FastAPIInstrumentation` in `services/api/main.py`. `infra/docker/prometheus.yml` scrapes `backend:9090`; the backend publishes only `6987`. Daemon comments in `services/daemon/metrics.py` claim OTEL owns 9090; compose publishes 9090 on `soc-daemon` and health JSON on 9091. `llm-worker` has no HTTP; do not give it a metrics port — attach OTLP metrics export next to the Prometheus reader when `OTEL_EXPORTER_OTLP_ENDPOINT` is set so worker instruments appear on the collector scrape already in `prometheus.yml` (`otel-collector:8889`).

This is the seam. Do not add series, do not touch Grafana, do not add an `OpsMetrics` module.

**Acceptance**
- [ ] With `VIGIL_OTEL_ENABLED` on, `GET http://backend:6987/metrics` is the OTEL registry (an instrument created via `get_meter` appears); it is not a separate `prometheus_client` registry
- [ ] `PrometheusMiddleware` and the `http_requests_total` / `http_request_duration_seconds` / `active_cases_total` / `findings_processed_total` instruments in `core/platform/monitoring.py` are gone; FastAPI OTEL instrumentation remains the HTTP signal
- [ ] Daemon listens on 9090 with that same registry; `curl soc-daemon:9090/metrics` returns Prometheus text. Do not move Prometheus onto the 9091 health server
- [ ] `infra/docker/prometheus.yml` backend job targets `backend:6987` path `/metrics`; daemon job stays `soc-daemon:9090`; collector job stays
- [ ] When `OTEL_EXPORTER_OTLP_ENDPOINT` is set, MeterProvider has both the Prometheus reader and an OTLP metric exporter (not an either/or fallback)
[...]

Issue #89, opened 2026-04-16
Per-Component Model Selection — Granular AI Model Configuration via Settings UI


mando222:
## Summary

Add granular, per-component model selection so that each part of the system (chat, triage, investigation agents, orchestrator planning, orchestrator review, context summarization) can use a different model — all configurable through a redesigned "AI Config" Settings tab (renamed from "Claude").

### Model Configuration Today

| Component | Current Model | Config Method | UI Control |
|-----------|--------------|---------------|------------|
| **Chat** | `claude-sonnet-4-20250514` | Default in `ChatRequest` schema (`backend/api/claude.py` line 41) | None — no model selector in chat UI |
| **Orchestrator (Plan)** | `claude-sonnet-4-5-20250929` | `ORCHESTRATOR_PLAN_MODEL` env var (`daemon/config.py`) | None |
| **Orchestrator (Review)** | `claude-sonnet-4-5-20250929` | `ORCHESTRATOR_REVIEW_MODEL` env var (`daemon/config.py`) | None |
| **Sub-agents (autonomous)** | Inherits from `config.plan_model` | `daemon/agent_runner.py` reads orchestrator config | None |
| **Context summarization** | `claude-sonnet-4-20250514` | Hardcoded in `claude_service.py` `_summarize_messages_sync/async` | None |
| **SOC Agent profiles** | No model field | `AgentProfile` dataclass in `soc_agents.py` has no `model` attribute | None |

### Key Problems

1. **No UI for model selection** — The Settings "Claude" tab (`frontend/src/pages/Settings.tsx` lines 861-873) only has an API key field. No model dropdowns anywhere
2. **No per-agent model config** — `AgentProfile` dataclass lacks a `model` field. All agents use whatever the caller passes
3. **Chat UI has no model picker** — `frontend/src/pages/Claude.tsx` / `ClaudeDrawer` doesn't expose model selection
4. **Cost tracking assumes Sonnet** — `daemon/agent_runner.py` hardcodes `SONNET_INPUT_COST = 3.0 / 1_000_000` and `SONNET_OUTPUT_COST = 15.0 / 1_000_000` (lines 24-25). Using a different model gives wrong cost data
5. **Inconsistent defaults** — Chat defaults to `claude-sonnet-4-20250514`, orchestrator defaults to `claude-sonnet-4-5-20250929`, summarizer uses yet another default
[...]

PR #948, opened 2026-09-16
obs: record LLM spend and tokens on the existing GenAI instruments

### LINKED ISSUE #894: obs: record LLM spend and tokens on the existing GenAI instruments
Part of #887

Part of #887

Coding: small · Vigil context: `core/llm/router/router.py`, `core/telemetry.py` `create_genai_metrics` · Good for: factory

**Context.** Spend is computed and stored; it is not exported. `create_genai_metrics` in `core/telemetry.py` already names `vigil.llm.calls.total`, `vigil.llm.duration.seconds`, `vigil.llm.tokens.total`, `vigil.llm.cost.usd.total` and has no production caller. Live dispatch is `LLMRouter.dispatch` in `core/llm/router/router.py` (Bifrost). Cost math is `compute_call_cost` in `core/llm/cost/calls.py`. `ClaudeService._persist_interaction` in `core/llm/harness/claude.py` writes `LLMInteractionLog` and is not the sole live path — do not increment there as well or you double-count. Record **once**, on the dispatch path, after the call returns, with the tokens and USD already in hand. `vigil_llm_cost_pricing_unknown_total` in `core/llm/providers/registry.py` already fires; leave it.

Depends on the sibling that makes Prometheus scrape serve the OTEL registry only for live proof; the record calls are safe no-ops when OTEL is off.

**Acceptance**
- [ ] One successful `LLMRouter.dispatch` increments the four GenAI instruments (calls, duration, tokens in/out, cost USD); attributes stay low-cardinality (model, provider, token_type). No prompt/response content
- [ ] `_persist_interaction` is not a second increment site
- [ ] Unknown pricing still only hits `vigil_llm_cost_pricing_unknown_total`
- [ ] With OTEL on and scrape working, PromQL on the exported names (`vigil_llm_cost_usd_total` after underscore conversion, not `soc_daemon_llm_*`) moves after a real call
- [ ] Telemetry failure never raises on the dispatch path

**Files:** `core/llm/router/router.py`, `core/telemetry.py` (reuse `create_genai_metrics`, do not duplicate names), tests under `tests/unit/llm/` and/or `tests/unit/platform/test_telemetry.py`.

**Do not build:** a cost metrics service, Bifrost as the export, `/api/analytics/cost` changes, Grafana, `soc_daemon_llm_*` aliases.

samarmstrong:
Closes #894

## What

Spend and tokens were computed and stored but never exported. This wires the four GenAI instruments that `create_genai_metrics` already names — `vigil.llm.calls.total`, `vigil.llm.duration.seconds`, `vigil.llm.tokens.total`, `vigil.llm.cost.usd.total` — to the two live transports, once per call, per the factory grooming notes.

- `core/telemetry.py` — new `record_llm_call(...)` next to `create_genai_metrics`. Lazily creates the instruments on first record and caches them only once `_initialized` is true **and** the meter is not the fallback no-op (so the no-op meter can never be pinned for the life of the process). Normalises every value before any instrument fires; whole body is `try/except Exception: pass`. `shutdown()` clears the cache so re-init gets fresh instruments. Attributes: `model`, `provider`; plus `token_type` ∈ {`input`, `output`, `cache_read`, `cache_creation`} on the token counter. No ids, no content.
- `core/llm/router/router.py` — `_dispatch_bifrost_openai` times the call and, after the response, prices the returned usage with `compute_call_cost` and records it. The `model` attribute is the requested bare id (the pricing catalog key) rather than Bifrost's `provider/model` echo.
- `core/llm/harness/claude.py` — `ClaudeService.chat` (direct SDK; never enters `LLMRouter.dispatch`) now prices the call **once** via a small `_call_cost` helper, passes that figure into `_persist_interaction` (new optional `cost_usd` kwarg, computed there only when absent) and into `record_llm_call`. Pricing once matters: a second `compute_call_cost` would re-fire `vigil_llm_cost_pricing_unknown_total`. Also extracts Anthropic `cache_read_input_tokens` / `cache_creation_input_tokens`, which `chat()` previously dropped, so both the log row and the token counter carry them.
- Not touched, per scope: `stream_openai_raw`, `dispatch_openai_stream`, `services/agent`, `_record_pricing_unknown`, `/api/analytics/cost`, Grafana, no `soc_daemon_llm_*` aliases.

## Independent review findings

1. Router helper could hit `None * rate` when a provider returns `None` usage fields, skipping the whole record — **fixed** (coerce before pricing).
2. Recording not atomic: a bad token value would leave 3 of 4 instruments fired — **fixed** (normalise all values before any instrument).
3. Fallback no-op meter could still be cached if `get_meter()` swallowed an SDK error post-init — **fixed** (`isinstance` guard).
4. `test_never_raises` never reached its bad inputs because instrument creation was patched to raise — **fixed** (split into creation-fails / bad-values / fallback-not-cached cases against the real reader).
5. `ClaudeService.chat` dropped Anthropic cache tokens, so `token_type=cache_*` would never appear on that path — **fixed** (two-line extraction; it is exactly the data the issue asks the token counter to carry).
6. Early-return when OTEL is disabled to avoid building no-op instruments per call — **dismissed**: four trivial object allocations; not worth diverging from the lazy pattern the grooming notes prescribe.
7. Add a lock around lazy instrument creation — **dismissed**: the OTEL SDK dedups instruments by name into one stream, so counts are correct; a race only produces a one-time SDK warning.
8. Router labels `model` with the requested id, Claude path with `response.model` — **dismissed**: both are low-cardinality and `response.model` is what the Claude log row already keys on; Anthropic echoes the requested id.
9. Histogram bucket boundaries are ms-oriented for a seconds instrument — **dismissed**: pre-existing in `create_genai_metrics`, which the issue says to reuse, not change; worth its own issue.
10. Router test docstring claimed attribute-shape coverage it doesn't have — **fixed** (docstring trimmed; coverage is in `test_telemetry.py`).
11. `_call_cost` docstring repeats #89/#184 history — **dismissed**: it is a moved-not-new comment; restyling isn't in scope.

## Verification

**Ran**
- `pytest tests/unit/llm tests/unit/platform` — 482 passed (7 new tests: helper against a real `MeterProvider` + `InMemoryMetricReader` checking all four names/attributes, instrument caching, no-op-meter-never-cached ×2, bad-values-skip-record, never-raises; router records once with returned usage and priced cost; router survives a pricing exception; `ClaudeService.chat` records once / prices once / passes cache tokens).
- `black --check`, `isort --check-only`, `flake8` on the three `core/` files — clean. The three test files were already non-black/isort-compliant at the base commit (the pre-commit hook only covers `services/` and `core/`); my added hunks are black-formatted and I left the pre-existing formatting alone.
- `pyright` on the three `core/` files — 9 errors before, 9 after; none introduced.
- Live: `VIGIL_OTEL_ENABLED=true`, `init_telemetry()` (real `PrometheusMetricReader`), a real `LLMRouter.dispatch` over HTTP against a stub OpenAI-compatible Bifrost returning `prompt_tokens=1000, completion_tokens=200, cached_tokens=500`, then `prometheus_client.generate_latest()`:
  ```
  vigil_llm_calls_total{model="gpt-4o-mini",provider="openai"} 1.0
  vigil_llm_duration_seconds_count{model="gpt-4o-mini",provider="openai"} 1.0
  vigil_llm_cost_usd_total{model="gpt-4o-mini",provider="openai"} 0.0003075
  vigil_llm_tokens_total{...,token_type="input"} 1000.0
  vigil_llm_tokens_total{...,token_type="output"} 200.0
  vigil_llm_tokens_total{...,token_type="cache_read"} 500.0
  ```
  \$0.0003075 matches the catalog: 1000·0.15 + 200·0.60 + 500·0.075 per-M tokens. A second dispatch with an uncatalogued model exported `vigil_llm_cost_usd_total{model="totally-unknown-model"} 0.0` and `vigil_llm_cost_pricing_unknown_total{...} 2.0` — the 2 is pre-existing (`compute_call_cost` resolves cost rates and cache rates separately, each firing the counter); this PR calls `compute_call_cost` exactly once per call on both paths.

**Did not run**
- Backend `GET /metrics` end-to-end: needs Postgres/Redis/Bifrost up. `core/platform/monitoring.py` serves `generate_latest()` with no argument, i.e. the same default registry the live check scraped, so the single-process proof is equivalent. #893 is not required.
- A live `ClaudeService.chat` call: needs an Anthropic API key; the path is covered by the mocked-SDK unit test.
- No UI change, so no `agent-browser` run.

**Blocked by the machine**
- none

<div><a href="https://cursor.com/agents/bc-18ce404c-e5fa-4694-97d5-9abca305372b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-18ce404c-e5fa-4694-97d5-9abca305372b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


samarmstrong:
🏭 **Factory environment agent.** Reading this PR (alongside 1 other recent factory PR) it found a gap in the cloud-agent environment and opened https://github.com/Vigil-SOC/vigil/pull/950 to close it.

Issue #828, opened 2026-09-08
feat(auth): the seeded admin must change its password on first login


samarmstrong:
Coding: small · Vigil context: `services/api/routers/auth.py` · Good for: factory

**Context.** `infra/database/init/06_auth_tables.sql` seeds a default admin. Nothing forces the credential to change, so installs that never rotate it keep a known password. `password_changed_at` already exists on `users`.

**Acceptance**
- [ ] The seeded admin's first successful login is forced through a password change before any other authenticated call succeeds (test)
- [ ] Existing installs: the same rule applies when `password_changed_at` is null
- [ ] Roles and route permissions are not touched; they exist already

**Files:** `services/api/routers/auth.py`, `services/api/middleware/auth.py`, the frontend login flow if needed, tests.

samarmstrong:
🏭 **Factory groom — needs work.** The factory is not picking this up as written. Edit the description or reply here and it gets reviewed again on a later tick.

The issue’s cited premise is already false: new installs do not get a seeded admin. A first-login password gate on `password_changed_at is null` would trap the bootstrap path and every admin-created user, because that column is never written at create time.

`infra/database/init/06_auth_tables.sql` (and the Helm copy) does not insert an admin. The comment at the bottom is explicit: the old row was an unusable hash that only blocked first-account creation, and it was removed so the operator chooses credentials via `POST /api/auth/bootstrap`. `start.sh` and `scripts/seed_reference_data.py` follow that. `LoginScreen.tsx` already offers first-account creation, not `admin` / `admin123`.

The remaining known-password paths are leftover scripts, not that SQL file: `scripts/init_default_user.py` and `scripts/init_default_credentials.py` still plant `admin` / `admin123`; `scripts/migrate_schema.py` `seed_default_admin` plants `admin` / `admin` on an empty table. README still documents those credentials. `SECURITY.md` already treats that pair as development material, not a production seed.

[...]

Issue #151, opened 2026-04-27
Refuse to start when DEV_MODE=true outside localhost


mando222:
## Background

`DEV_MODE` (in `backend/middleware/auth.py`) bypasses all authentication. It defaults to `false` and prints a startup warning when enabled, which covers the common-case footgun. But the warning is just a log line — nothing prevents `DEV_MODE=true` from being set in a non-development environment via env var or container config.

Split out from [#87](https://github.com/Vigil-SOC/vigil/issues/87) ([#150](https://github.com/Vigil-SOC/vigil/pull/150)) so the prompt-injection PR stays focused.

## Proposal

At backend startup, when `DEV_MODE=true`:

1. Detect whether the bind host is localhost (`127.0.0.1`, `::1`, `localhost`).
2. If the bind host is anything else (including `0.0.0.0` or a real interface), refuse to start with a clear error message naming the env var.
3. Log a `security.dev_mode_enabled` audit event on every startup so the bypass is auditable in production telemetry.

## Acceptance

- [ ] Server fails fast on `DEV_MODE=true` + non-localhost bind host.
- [ ] Server starts normally on `DEV_MODE=true` + localhost bind host with the existing warning.
- [ ] Audit log entry is emitted whenever `DEV_MODE=true` at startup.
- [ ] Test coverage for both pass and fail cases.

## Notes

- One-line follow-up — should be a small, focused PR.
- Affected file: `backend/middleware/auth.py` (or wherever the startup check lives).

PR #955, opened 2026-09-16
auth: a fresh install requires a login, and the bypass says so on every startup

### LINKED ISSUE #914: auth: a fresh install ships with authentication bypassed
Part of #881 ("auth on by default; the development bypass is opt-in and loud").

## The install nobody has to authenticate to

`env.example:21` ships `DEV_MODE=true`. `services/api/middleware/auth.py:24` reads that flag into a module constant at import and, when set, `get_current_user` returns a mock admin for every request without looking at a token. The same flag unlocks the vstrike inbound check (`services/api/routers/vstrike.py:119`). It does *not* reach `/internal`, which is shared-secret only (`core/agents/internal_auth.py`) — that path has no dev-mode branch.

`setup_dev.sh:20` copies `env.example` to `.env`, so every contributor install inherits the bypass. The packaged paths already run authenticated — `scripts/app_up.sh:26` forces `DEV_MODE=false` and mints a JWT secret, and the desktop image defaults it false — so the bypass now survives only on the paths humans actually develop against.

That inversion has a consequence beyond the default itself: the authenticated first-run has never been exercised. `POST /auth/bootstrap` and its login-screen entry point exist and are wired, but with auth off in every development environment, nothing proves a clean install can create an admin, sign in, and reach an authenticated route.

## Known default credentials still ship

`infra/database/init/06_auth_tables.sql:139` removed the seeded admin deliberately, recording that a bcrypt hash matching a published password does not belong in every deployment. Two scripts kept it alive: `scripts/init_default_user.py:31` and `scripts/init_default_credentials.py` both create `admin` with the password `admin123`, and `README.md:254` instructs operators to run one of them.

## A startup guard that was closed as done and is not in the tree (context, not scope)

#151 ("Refuse to start when DEV_MODE=true outside localhost") is closed as completed. Its acceptance was a bind-host check that fails fast when the bypass is on and the server is listening on anything but loopback, plus a `security.dev_mode_enabled` audit event on every startup. Neither exists: no bind-host check, no audit event, no tests. The issue named `backend/middleware/auth.py`, a path removed in the `core/` + `services/` restructure, so the behaviour appears to have been lost rather than reverted.

## Acceptance

- [ ] A fresh `.env` — from `env.example` or from `setup_dev.sh` — requires authentication
- [ ] An existing `.env` is never rewritten; installs already configured for the bypass keep it
- [ ] `setup_dev.sh` provisions a JWT signing secret, so a fresh setup boots with auth on
- [ ] With the bypass enabled, every startup announces it and names each gate it opens: session auth and vstrike inbound
- [ ] The web UI carries a persistent, non-dismissible indicator while the bypass is active
- [ ] A bypass on an address the network can reach says so in its own words, distinctly from a bypass on loopback — the two are not the same fact and should not read the same
- [ ] A clean install can create the first admin and reach an authenticated route — bootstrap, login, authenticated call — proven by a test against an empty database
- [ ] No script or document in the repository creates an account with a password the operator did not choose
- [ ] The Python test suite runs with authentication on, with one shared fixture supplying an authenticated client
- [ ] `scripts/reset-setup.sh` and `scripts/generate_sample_data.py --api` authenticate, and fail with a clear message when they cannot
- [ ] The bypass is documented where it is configured; `env.example:20`'s reference to the non-existent `DEV_MODE.md` is gone

## Bounds

Enabling the bypass must stay a single documented line — a developer who wants an unauthenticated instance on their own machine keeps one.

Whether vstrike inbound honours the bypass is not in question here; it continues to, and is only named in the announcement.

#151's refusal is deliberately **not** restored. The epic asks for the bypass to be "opt-in and loud", not for it to block a boot, and a service that will not start is one nobody can use to work out why. The escalating announcement is what replaces it. #151 needs closing out separately either way: it is marked completed with none of its code in the tree.

Out of scope: forced password rotation on first login (#828, whose stated premise — that `06_auth_tables.sql` seeds a default admin — is no longer true); the Splunk export scripts, which address Splunk's HEC and not the Vigil API; auth hardening beyond the bootstrap-and-login chain. Bugs found outside that chain while proving it become their own issues.

## Measured

Running the Python suite with `DEV_MODE=false` breaks 6 tests, all in `tests/integration/test_mcp_enable_transactional.py`, all on the same cause: a 401 on `PUT /api/mcp/servers/{name}/enabled`. Every other HTTP test already overrides the auth dependency for itself. One shared fixture covers all 6 — and lets `test_api_v1_dual_mount.py`, `test_api_v1_agent_runs.py` and `test_llm_providers_api.py` drop their hand-rolled copies. `tests/conftest.py:8`'s `DEV_MODE=true` default is load-bearing for one file.

G-r-ay:
Closes #914. Part of #881 ("auth on by default; the development bypass is opt-in and loud").

## What changes

`env.example` ships `DEV_MODE=false`, and `setup_dev.sh` mints a JWT signing secret at `~/.vigil/jwt_secret` so a fresh setup boots authenticated instead of raising at import. An existing `.env` is never rewritten — an install already configured for the bypass keeps it, and turning the bypass back on stays one documented line.

Enabling it is now loud. `core/auth/dev_mode.py` announces on every startup, names each gate the flag opens (session auth, the vstrike inbound key) rather than saying "auth is off", and emits a `security.dev_mode_enabled` event carrying the bind host. A bypass on an address the network can reach says that separately, in its own words. It never refuses the boot: someone enabling this on a throwaway box is doing it deliberately, and a service that will not start is one nobody can use to find out why.

The web console carries a persistent, non-dismissible `NO AUTH` item in the rail while the bypass is active. It reads `GET /api/health`'s new `auth_bypassed` field as well as the build flag, because `VITE_DEV_MODE` is frozen into the bundle and `DEV_MODE` is not: a prebuilt image deployed with the bypass on would otherwise show nothing, and that is the deployment someone other than the developer can reach.

`scripts/init_default_user.py` and `scripts/init_default_credentials.py` are deleted. Both created `admin` / `admin123`, which `infra/database/init/06_auth_tables.sql:139` had already removed from the SQL on purpose; the README instructed operators to run one of them. Nothing in the repository now creates an account with a password the operator did not choose, and the Helm notes say so instead of printing the old pair.

## Proving the authenticated first-run

The bootstrap chain existed and was wired, but with auth off everywhere in development nothing had ever exercised it. `tests/integration/test_first_admin_bootstrap.py` drives it against an empty database — create the first admin, sign in, reach an authenticated route — with its own `vigil_auth_test` database provisioned in CI, since it applies the auth DDL itself and truncates `users`.

The Python suite now runs authenticated, with one shared `authenticate_app` fixture in `tests/conftest.py`. `scripts/reset-setup.sh` and `scripts/generate_sample_data.py --api` authenticate, and fail with something actionable when they cannot, rather than at the first mystery 401 partway through.

## Not here

`#151`'s refuse-to-start guard is deliberately not restored — the epic asks for the bypass to be opt-in and loud, not for it to block a boot, and the escalating announcement replaces it. That issue still needs closing out separately: it is marked completed with none of its code in the tree. `#828` (forced rotation of a default password) should close as stale; its premise, that the SQL seeds a default admin, stopped being true.

Two judgement calls worth a reviewer's eye:

- `services/api/main.py`'s `__main__` block now binds `settings.bind_host` rather than a hardcoded `0.0.0.0`, so the banner cannot report loopback while the socket is open to the network. Every launcher in the repo already defaulted that to `127.0.0.1`; only a direct `python services/api/main.py` changes behaviour.
- `core/auth/dev_mode.py` holds both the announcement and the bind-host resolution behind it. `CONTEXT.md`'s mechanism-versus-knowledge cut would put the resolution half in `platform`; splitting a small new file across two packages seemed to cost more than it buys.

## Verification

- `tests/security/` — 31 pass, including the exposed-versus-loopback wording and 16 cases for what counts as loopback (the rest of 127/8, `::ffff:127.0.0.1`, an unset host, an unresolvable name)
- `tests/security` + `tests/unit/api` + `tests/unit/auth` + `test_mcp_enable_transactional.py` — 270 pass
- the touched integration and unit files — 88 pass
- `DevModeWarning` — 6 pass; `tsc` and `eslint` clean
- driven in a browser against the live stack: with `DEV_MODE=true` and a bundle built without `VITE_DEV_MODE`, the console opens unauthenticated and shows `NO AUTH`; with `DEV_MODE=false`, the same bundle stops at the first-admin bootstrap screen


G-r-ay:
Closing as superseded. #1020 landed the groomed slice of #914 — `env.example` shipping `DEV_MODE=false`, `ensure_jwt_secret`, and the startup announcement naming each gate the flag opens — and #914 is closed with it.

I checked what this branch still carried that `main` lacks:

- `scripts/init_default_user.py` / `init_default_credentials.py` (the `admin`/`admin123` pair) — already gone from `main`
- `tests/integration/test_first_admin_bootstrap.py` — already on `main`
- `auth_bypassed` on `/api/health` and the console `NO AUTH` rail item — **not** on `main`

So only the console half is left, and it is now #1046, rebased onto current `main` rather than carried on a branch that had gone 20 commits stale and conflicted on `.github/workflows/ci-cd.yml`.

Nothing here is lost that is not either merged or in #1046.

PR #972, opened 2026-09-17
fix(daemon): normalize an aware last_activity_at before subtracting naive now

### LINKED ISSUE #963: fix(daemon): the supervision loop raises on every cycle, so no runtime cap or stale kill ever fires
## Summary

`_supervision_loop` raises `TypeError: can't subtract offset-naive and offset-aware datetimes` on every iteration as soon as any investigation reaches `executing`. The exception aborts the loop body, so **nothing downstream of that line ever runs**: stale-agent termination, the per-investigation runtime cap, and approval escalation.

The docstring of the affected function is *"Monitor running agents for stuck/runaway conditions."* With it dead, an autonomous deployment has no enforced wall-clock ceiling.

This looks like a residue of #541 (`Replace deprecated datetime.utcnow() / asyncio.get_event_loop()` repo-wide) — that sweep moved callers onto an aware `utcnow()`, but this comparison reads a naive value back out of the database.

## Location

`services/daemon/orchestrator.py:555`

```python
last_activity = inv_dict.get("last_activity_at")
if last_activity:
    if isinstance(last_activity, str):
        last_activity = datetime.fromisoformat(last_activity)
    idle_seconds = (now - last_activity).total_seconds()   # raises
```

`now` is `utcnow()` — aware. `investigations.last_activity_at` is `timestamp without time zone`, so the ORM hands back a naive `datetime`. The `isinstance(..., str)` branch only normalises the string case; the ordinary ORM path stays naive.

## Reproduction

1. `POST /api/orchestrator/enable`
2. Create any investigation so one row reaches `status = 'executing'`
3. `tail -f logs/daemon.log`

## Observed

- `Supervision loop error: can't subtract offset-naive and offset-aware datetimes`, logged every ~15 s — 126 occurrences in one evaluation run.
- An investigation with `max_runtime_seconds: 600` ran **1927 s (3.2x its cap)** and stayed `executing`.
- `stats.stuck_agents_killed` remained `0` throughout.
- It only stopped after a manual `POST /api/orchestrator/investigations/{id}/kill`. Without operator intervention it would not have terminated.

## Expected

The runtime cap terminates the investigation at its limit and increments `stuck_agents_killed`.

## Suggested fix

Normalise before subtracting, covering the ORM path as well as the string path:

```python
last_activity = inv_dict.get("last_activity_at")
if last_activity:
    if isinstance(last_activity, str):
        last_activity = datetime.fromisoformat(last_activity)
    if last_activity.tzinfo is None:
        last_activity = last_activity.replace(tzinfo=timezone.utc)
    idle_seconds = (now - last_activity).total_seconds()
```

Other `utcnow() - <column>` sites are worth auditing for the same shape, since #541 evidently did not reach them.

## Test coverage

```
$ grep -rl "_supervision_loop\|stale_threshold\|stuck_agent" ./tests
(no matches)
```

The runaway-protection path has no test coverage, which is presumably how this reached a release. A regression test that drives one `executing` row with a naive `last_activity_at` through the loop would catch it.

---

*Found during a functional evaluation of 0.5.0: 106 synthetic alerts through the `/ingest` webhook on a host-run single-node deployment, `DEV_MODE=false`.*

AmirF194:
`_supervision_loop` reads `last_activity_at` back through `InvestigationSchema.dump()`, and that schema's `IsoDateTime` field always stamps a naive DB value with `+00:00` (`core/storage/schemas/base.py`'s `_iso_utc`, `when_used="json"`, and `dump()` always calls `model_dump(mode="json", ...)`). So the string the loop parses back with `fromisoformat` is always aware. `now` stays naive by design (`core/time.py`). Subtracting the two raised on every tick for every executing investigation, not only the stale ones, so the idle check, the stale-kill, and the per-tick cost/correlation work after it never ran.

Slight correction to the issue: the suggested fix (`if last_activity.tzinfo is None: replace(tzinfo=utc)`) doesn't help, because `last_activity` is never naive by the time it reaches this branch. `_get_investigations_by_status` always returns already-serialized dicts, so `isinstance(last_activity, str)` is always true and `fromisoformat` always yields an aware value.

Stripped the parsed value's tzinfo instead, matching the naive convention the rest of the loop already assumes (`now`, the hourly-cost cutoff a few lines down).

Added a regression test that drives `_supervision_loop` for one tick with a serialized, aware `last_activity_at`; the new test fails without the change with the reported `TypeError`, and the stale-kill fires as intended with it. A second test with a fresh timestamp needs the same fix (any executing investigation crashes the tick, not just stale ones).

Fixes #963


AmirF194:
Closing this: #976 landed the same fix (normalising `last_activity_at` before the comparison) and closed #963 directly. No need to carry two PRs for one bug.

PR #979, opened 2026-09-17
mcp: one server, reachable, acting as whoever called it

### LINKED ISSUE #961: mcp: Vigil's own tool surface is three servers only Vigil can reach
Part of #881 ("one MCP server with stable tool names").

## Vigil's own tool surface is three servers, and nothing outside Vigil can reach any of them

Vigil uses MCP in two directions and the code does not distinguish them. Outward, it is a client of roughly thirty vendor servers in `mcp-config.json`. Inward, it ships three servers of its own under `tools/mcp/` — `deeptempo-findings` (28 tools), `approval` (5), `tempo-flow` (2). The second set is the product surface: how something that is not Vigil drives Vigil.

Every in-repo server ends in `mcp.server.stdio.stdio_server()`, and `core/integrations/mcp/service.py` spawns each one as a child process. A stdio server has no address. The only caller that can reach it is the parent that launched it, which is Vigil itself — so Vigil starts a subprocess and speaks JSON-RPC over a pipe to a copy of its own code in order to read its own database.

"Vigil 1.0" (Revision 3) states that every platform component consumes Vigil "through the same `/v1` API and vigil-mcp tools a customer or contributor would use". There is no transport today by which anyone can do that.

## The two roles share one registry

`core/integrations/mcp/service.py:255` puts `deeptempo-findings`, `tempo-flow` and `approval` in `_DEFAULT_ENABLED` alongside `security-detections` and `splunk-selfhosted`. `clients/web/src/screens/settings/integrationsData.ts:41` lists them in an "Internal / Platform" group beside the vendor integrations. `core/integrations/mcp/registry.py:81` flattens every connected tool to `{server}_{tool}`, so Vigil's own tools reach the model as `deeptempo-findings_list_findings` and `approval_approve_action`, and `core/llm/chat_layers.py` offers every connected MCP tool to chat regardless of which side it came from.

The consequence is that Vigil's own tool surface is configured, toggled, named and displayed as though it were a third-party integration Vigil happens to have credentials for.

## The caller asserts its own identity

The MCP tools take identity as an argument. `close_case` takes `closed_by` as a string; `add_case_comment` and the rest follow the same shape. Over a pipe from the parent process that is merely redundant. As a reachable surface it means the caller names whoever it likes in the case record. The `/api/v1` routes take the same fact from the authenticated user.

## The same data is modelled twice

`services/api/routers/cases.py` has roughly forty routes over findings, cases, approvals and workflows, reached through the case service layer — `core/cases/case_records_service.py`, `CaseCollaborationService`, `CaseEvidenceService`, `CaseIOCService`, `CaseSLAService`. The MCP tools cover most of the same ground but open their own `DatabaseService` sessions and do the work themselves. Two implementations of one contract, free to disagree, which is the condition the epic's groom already recorded for Finding.

A few MCP tools have no route at all — `create_case_from_killchain`, `technique_rollup`, `list_completed_hunts`, `add_case_mitre_techniques` — so the surfaces are not a subset relation in either direction.

## `tempo-flow` does nothing, and four places describe what it would do

Both of its tools are inert. `tempo_get_workflows` reads `tools/data/investigation_workflows.json`; `tools/` contains only `mcp/`, so it returns `{"count": 0, "workflows": []}` on every call. `tempo_run_workflow` is a stub — `tools/mcp/tempo_flow.py:60` says so in a comment — and returns `status: "initiated"` having orchestrated nothing. Workflows are served by `core/workflows/` and, since #913, by `/api/v1/workflows`.

Four places state otherwise. It is on by default (`service.py:257`). `integrationsData.ts:58` tells the operator it "orchestrates multi-step agent workflows and playbook execution. Required for automated investigation chains." `core/memory/source_tier.py:150` grades it `SourceTier.TELEMETRY` under the comment "Flow data is an observation" — the name was read as network flow, and a workflow launcher is graded as a telemetry source in the path that decides what a hunt may treat as evidence. `README.md:130` lists it as Core, alongside an "ATT&CK Layer" server that is in neither `mcp-config.json` nor the UI group.

## What the names record

`core/storage/models/episodic.py:171` stores `source_system` as free text, written from the server name. Rows already in the database cite `deeptempo-findings`. Whatever the server ends up called, those rows name a system that no longer exists; the stored tier beside them means nothing breaks on read.

## Acceptance

- [ ] Vigil's own tools are served by a single server, under a name that describes Vigil rather than its sponsor
- [ ] That server is reachable by a caller that is not Vigil, over a transport with an address
- [ ] It is off on a fresh install, and one documented setting turns it on
- [ ] A caller authenticates before it can list or call a tool
- [ ] A tool call acts as the authenticated caller; no tool takes the actor's identity as an argument
- [ ] A credential that opens the MCP surface does not authenticate `/api/v1`, and the reverse
- [ ] Tool names carry no server prefix, and a tool is the same name to Vigil's own chat as to an external caller
- [ ] A tool and the `/api/v1` route over the same data reach it through the same service, so neither can answer differently from the other
- [ ] `tempo-flow` is gone, along with every place that describes it: the config entry, the default-enabled set, the Settings entry and its description, the `source_tier` grading, the README line
- [ ] The README names only servers that exist
- [ ] The enable toggle, the Settings list and the connection status cover the servers Vigil calls out to, and no longer present Vigil's own surface as one of them
- [ ] Historical `source_system` rows citing the old server name are either migrated or documented as historical

## Bounds

The roughly thirty vendor servers do not change. Vigil remains an MCP client for those, configured and toggled as it is today; this issue separates that role from Vigil's own surface rather than touching it.

**Tool names are not frozen here.** The epic's freeze is a commitment made once, and the set should be chosen after the surface is built and has a consumer, not before. That is a later issue under #881, before 1.0. Until it lands, these names carry the `0.x` breaking-change terms in `SECURITY.md`. Renaming anything that reads badly is cheaper now than at any later point.

**This ships as one PR.** The pieces only look separable. A server renamed but still unreachable is the same product with a different label; a transport with no credential is one nobody may open; tools that act as the caller mean nothing until a caller can be identified. Each intermediate state is either inert or worse than today, and none of them is something an operator could be given. The cost of one larger review is accepted deliberately in exchange for never shipping a half-built surface.

**The credential is settled here, not deferred.** Nothing in `core/auth/` or `services/api/middleware/auth.py` issues anything but a session for a human logging in, so opening this surface means a kind of credential Vigil does not have. It is its own kind: minted deliberately by an operator, held by a program rather than a person, presented on the MCP surface and refused everywhere else — a session token is likewise refused here. Because the surface is off until an operator turns it on, no such credential exists on an install that never enables it, and turning the surface off again ends every credential's usefulness with it. What a caller may then do is the role model Vigil already has, not a second one.

Out of scope: the authentication default and the development bypass (#914); new `/api/v1` routes; the MCP tools that have no route today keep whatever coverage they have rather than growing one.

### LINKED ISSUE #978: mcp: every tool's schema requires a 'kwargs' argument that does not exist
## Every tool on Vigil's own MCP server demands an argument that means nothing

Calling any of them the way its schema describes fails:

```
tools/call list_findings {"limit": 2}
-> Error executing tool list_findings: 1 validation error for list_findingsArguments
   kwargs
     Field required [type=missing, input_value={'limit': 2}, input_type=dict]
```

The call succeeds only when the caller invents a value for an argument the tool does not use:

```
tools/call list_findings {"limit": 2, "kwargs": {}}
-> {"total": 0, "returned": 0, "findings": []}
```

`tools/mcp/deeptempo_findings.py` ends 31 of its tool signatures with `**kwargs`, and the `mcp` 2.x SDK reads that as a parameter named `kwargs` and marks it **required**. Read the schema back and it says so:

```python
>>> [t for t in mcp._tool_manager.list_tools() if t.name == "list_findings"][0].parameters["required"]
['kwargs']
```

This is on `main` at `949abac`, for every tool the server publishes.

## Two mechanisms that were each right on their own

`**kwargs` is load-bearing. `core/agents/tools_router.py:130` attaches a row cap to every call: when nothing describes the tool, `_bounded` adds `limit=max_rows` whether or not the tool takes one. The comment above it records why the blind form exists — "injecting limit made every point-read invalid_args" — and `**kwargs` is what absorbs it. Remove `**kwargs` alone and `get_finding` breaks on a keyword it does not accept.

What makes every MCP tool reach that branch is narrower than it looks: `_schema_row_caps` asks `MANIFEST`, and `MANIFEST` is built from `BACKEND_TOOLS` alone (`core/agents/tool_registry.py:22`). An MCP tool is described by the registry, which that lookup never consults, so every one of them is treated as undescribed and gets the blind `limit`.

The `**kwargs` predates the collision — it arrived with the tool bridge in #638 (14 August). #787 (Migrate in-repo MCP servers to the mcp 2.x API) changed what `**kwargs` means to the SDK, and nothing connected the two.

## What it has cost

The schema Vigil publishes for these tools is wrong in the only way that matters: a caller that obeys it is refused. Anything reading the schema and calling accordingly — a model choosing arguments, an external MCP client, a contributor writing against the documented shape — fails on every call, and the error names an argument that appears nowhere in the tool's documentation.

Whether Vigil's own chat has been affected depends on what it sends; the surface published to anyone else has not worked since #787 regardless.

## Acceptance

- [ ] A tool's published schema lists the arguments it takes, and no others
- [ ] A call that supplies exactly what the schema asks for is accepted
- [ ] No tool requires an argument it does not use
- [ ] The row cap still reaches a tool that declares one, and is not sent to one that does not — the behaviour `_bounded`'s comment describes
- [ ] The cap is still enforced on what a tool answers with, for a tool that cannot be told to limit itself
- [ ] Something fails if a tool is published with an argument its callers cannot supply

## Bounds

The row cap is not being removed; a tool that answers in bulk still has to be bounded, and what is wrong is only which tools are told to bound themselves.

Out of scope: the `_ROW_CAP_ARGS` names, and whether unknown arguments should be rejected or ignored more generally — this is about a schema that asks for one nobody can supply.

G-r-ay:
Closes #961. Also fixes #978, found while testing this.

Vigil's own tools were three servers that only Vigil could reach. They are one server, reachable by a caller that is not Vigil, acting as whoever presented a credential, and reaching the data through the same services the API routes use.

Eight commits, each its own change. 2453 tests green.

## What each one does

**`624c9893` — delete `tempo-flow`.** Both its tools were inert: `tempo_get_workflows` read `tools/data/`, which does not exist, so it returned zero workflows on every call; `tempo_run_workflow` was a comment-marked stub. It was on by default, Settings told operators it was "required for automated investigation chains", and `source_tier` graded it `TELEMETRY` under "Flow data is an observation" — the name read as network flow, so a dead workflow launcher counted as a telemetry source in the path that decides what a hunt may treat as evidence. The README's Core row also named an ATT&CK Layer server removed in #873/#876.

**`02c13b3a` — one server named `vigil`.** `approval`'s five tools move into the findings module with the same names, arguments and JSON. `tools/mcp/base.py` goes too; nothing imported it. `source_tier` keeps a line for `deeptempo-findings`: existing `source_system` rows cite it, and grading the historical name costs one line where rewriting the rows would cost a migration.

**`1eea66e5` — the tools reach data the way the routes do.** Each tool opened its own session and wrote its own rows while the routes went through the case services. They already disagreed: a comment left through a tool reached nobody, because `CaseCollaborationService.add_comment` extracts mentions and notifies watchers and the tool did not; the same IOC named twice was one row through the API and two through a tool. Four operations had no service to call because the routes had hand-rolled them too — those are `core/cases/case_journal_service.py` now, called by both.

**`d429b4da` — a tool records who called it.** Seven tools took the actor as an argument. A caller supplying its own name is not identifying itself, it is choosing what the record will say. The sharpest case: `approve_action` defaulted to `approved_by="analyst"`, so an agent could leave a record saying a person had signed off on isolating a host — which this repository refuses for closures in as many words.

**`159520dd` — a credential a program holds.** Its own table keyed to a user, carrying no permissions of its own. SHA-256 rather than bcrypt, because the token is 32 CSPRNG bytes and bcrypt's work factor is for a secret a person chose. Refused on `/api/v1` by name, and a session token is refused on MCP.

**`080cc7c7` — the surface, at `/mcp`.** Off on a fresh install, one Settings toggle, no restart. Closed is 404, not 403. Every credential failure reads the same. The development bypass gets a fixed credential that is obviously not one, refused when the bypass is off.

**`0dbf1a9b` — no more subprocess.** To use its own tools Vigil started a second copy of itself and spoke down a pipe to it. Tool names lose their server prefix with it.

**`431051f4` — #978.** Every tool required an argument named `kwargs` that no tool uses.

## Two things to look at closely

**The in-process switch is the risky one.** `tests/unit/integrations/test_in_process_equivalence.py` writes down the contract the pipe kept — never raises, always a dict, `error` not `isError`, every call bounded — and holds the direct path to it. The first draft said `isError`, which is silently falsy; every failure would have read as a success returning no rows.

One difference could not be closed: a tool that hangs now holds a thread inside Vigil rather than inside a disposable child. The timeout bounds the caller, not the thread. The isolation was already spent — `/mcp` is served by this same app — but it is a real loss and not argued away.

**`_is_destructive_mcp` had a live defect** that bare names exposed. It took the first `_`-separated token as a server prefix and discarded it, so on an unprefixed name it discarded the verb: `isolate_host` read as `host` and would have been offered to chat ungated, against a docstring saying a missed one is an ungated detonation. It reads every token now, which can only over-drop.

## Deliberate changes in behaviour

- A Case opened through a tool is named the way the service names one, `case-YYYY-MM-DD-`, as one opened through the API already was.
- `escalate_case` goes through `CaseWorkflowService`, gaining the assignee reassignment and timeline entry the route had, and losing a priority bump on critical urgency that was the tool's alone. Raising priority belongs in the service if it belongs anywhere; not restored here.
- An operator who persisted a disabled state for `deeptempo-findings`, `approval` or `tempo-flow` has a stale entry in the MCP state file naming a server that no longer exists. Harmless — it is looked up by name and never matches.

## Not done here

Tool names are not frozen. The set should be chosen after the surface has a consumer, and that is a later issue under #881, before 1.0.

Whether an agent should be able to approve its own request at all is a product question this does not answer. The record is now true; that is a different thing.

## Tested

2453 unit and security tests. Also driven against a running server: the handshake, `tools/list` returning 34 bare names, and `list_findings` / `list_cases` / `technique_rollup` answering calls that supply exactly what the schema asks for — `list_cases` off a real database.


G-r-ay:
On the groom's note that `is_enabled()` is "two places for one switch" — it reads that way and is not meant to be. There is one source of truth, and it is Settings.

`VIGIL_MCP_ENABLED` decides what a fresh install boots with, and nothing after that. Once the toggle has been used, the stored choice is the answer and the environment variable no longer participates. It is a starting value, not a second switch, and not a ceiling: a deployment cannot set it to `false` and rely on that to keep the surface shut.

Both files say so already — `core/integrations/mcp/surface.py` in `is_enabled`'s docstring, and `env.example` under the MCP section — so this is here to save a reviewer re-deriving it, not to add anything.

Worth disagreeing with if you want the environment to be a hard ceiling instead. That would be a different thing: a deployment could then forbid the surface and no operator could open it from the console. I did not build that because the toggle was asked for as a runtime lever, and a lever an operator can be locked out of is a worse answer to "can I turn this on" than a setting they own.

G-r-ay:
Addressed — pushed 4 commits.

**Both reachability blockers fixed.** CSRF: `_DEFAULT_EXEMPT` → `_ALWAYS_EXEMPT`, unioned with the operator's list rather than replaced by it, deduped after the context path is applied; the test now uses the literal shipped string and a new one reads both configs off disk. Bare `/mcp`: a `BareMountPath` middleware normalizes the path ahead of routing, so it can't be re-decided by whatever registers around it, and the gate tests now register an SPA catch-all so both spellings are exercised here rather than only the one CI has.

**One thing was bigger than the review.** While measuring the `list_findings` divergence I found this branch had *silently removed six tools from chat* — `list_findings`, `list_cases`, `get_case`, `create_case`, `update_case`, `add_finding_to_case`. `chat_layers.py` spread the MCP entries last, so an unprefixed Vigil tool won the catalogue for a shared name, and those six carry no docstring, so an empty description dropped them entirely. Fixed as the 4th commit (`{**mcp, **static}`), with a test that fails against the old order.

**Deferred as their own issues, per one-PR-per-issue:** the same-name/two-implementations divergence is #1086, and the in-process principal binding (chat/hunts still record `agent`) is #1087 — that one's a wire-contract change through the agent layer, not a context-var, so it genuinely doesn't fit here.

Also done in-PR: CONTEXT.md + the credentials panel now agree with the code on Trust; `IntegrationsSection` "Connectors" → "MCP Servers"; dead `SERVER_DESCRIPTIONS.approval`; the "over a pipe" wording; an `.importlinter` contract naming the one `core`→`tools` crossing; the dead gate-test `setdefault`; and the deferred chat-approves-containment decision written where it's implemented. `_get_dev_user` repointed so it's correct whichever of #913/#979 merges second.

nestor-deeptempo (approved):

PR #953, opened 2026-09-16
feat(memory): list prior hunts by entity key and technique, in-flight via workflow_runs

samarmstrong:
Closes #899

## What

New module `core/memory/prior_hunts.py` beside `recall.py`, one public function:

```python
list_prior_hunts(entity_keys, techniques=(), *, session=None, runs=None)
-> {"keys", "techniques", "concluded": [...], "in_flight": [...]}
```

- **Concluded**: one `text()` query over `episodic_verdicts` (`investigation_kind = 'hunt'`) using the same `&&` overlap `recall.py` uses, on `subject_entities` and/or `techniques`. A dimension the caller did not supply adds no predicate (three prebuilt statements, one per shape), so a key-only query needs no technique match. `LEFT JOIN episodic_distil_markers` on (`investigation_kind`, `investigation_id`) for `origin_run_id`. Ordered `concluded_at DESC, id DESC`, `LIMIT LIST_RUNS_MAX`. Rows carry `investigation_id, hypothesis_id, statement, outcome, concluded_at, origin_run_id, matched_keys, matched_techniques`.
- **In-flight**: `WorkflowRunService.list_runs(workflow_id="threat-hunt", status=..., limit=LIST_RUNS_MAX)` for `running` and `paused`; matched on `trigger_context["hypothesis_subjects"]` (keys normalised with `normalise_keys`) and a whole-token T-ID match on `trigger_context["hypothesis"]`. A run that declared nothing is not a hit. Sorted by `started_at` desc across both statuses, capped at `LIST_RUNS_MAX`.
- Keys go through `normalise_keys`; T-IDs are upper-cased and validated against `T\d{4}(.\d{3})?`.
- Not done, per the issue: no HTTP, no parser, no config, no `list_runs` change, no `read_projection`, no `agent_events`, no `recall.py` change, no read-log write.

Tests in `tests/unit/memory/test_prior_hunts.py` (real Postgres, `episodic_session` fixture): concluded by ip with origin_run_id, concluded by T-ID / either-dimension, in-flight by declared subject and by hypothesis text (paused included, scheduled and completed excluded, `T15661` not a `T1566` hit), no-match and nothing-asked empty, cap newest-first. The fixture asserts nothing in `prior_hunts` is imported from `recall` and that no `episodic_read_log` row is written.

## Review findings and disposition

[...]

PR #946, opened 2026-09-16
fix(detections): make the ART execute trace and reconstruction agree so a real run scores

samarmstrong:
Closes #932

## What changed

`atomic_red_team_execute` moves to the reconstruction spelling so a journaled ART trace joins to Findings without translation.

- `core/integrations/atomic_red_team/tool.py`: `execute_atomic` emits `technique_id`, `started_at`, `ended_at` and echoes `hostname` (was `technique` / `finished_at`, no host). `hostname` is refused before the runner in the same block as `environment_id` and passed as `--hostname <value>` beside `--technique`. Same keys on the timeout branch. Tool description, `inputSchema` and module docstring updated.
- `core/detections/reconstruction.py`: untouched, as scoped. Confirmed rather than edited: `tools_router._rows` wraps the bare execute dict as one row and `_steps_from_rows` accepts it once `technique_id` is present; `_window` and `_has_entity` pass with `started_at`/`ended_at`/`hostname` on the row.
- `tests/unit/detections/fixtures/recorded_red_run.json`: `steps` are now generated from `execute_atomic` with a stub runner and a pinned `_now` (`tests/unit/detections/fixtures/art_trace.py`, rerun with `python3 -m tests.unit.detections.fixtures.art_trace`). `test_fixture_steps_are_what_execute_atomic_emits` asserts the committed steps equal the generated output, so the fixture and tool cannot drift. The fixture drops `id` / `src_ip` / `user` / `atomic_guid`; the `loglm` Findings now carry the step hostname in `entity_context`.
- `test_execute_atomic_trace_scores_end_to_end` proves the seam on the tool's own output: `execute_atomic` → `steps_from_dispatch_results` (both the `ok/rows` envelope and the bare dict) → `reconstruct` gives `rule` / `loglm` / `both` → `coverage_report` groups under the technique. `test_execute_atomic_trace_with_no_matching_finding_is_missed` gives `missed` with the host in `missed[0].hostname`.
- Join-by-step-id, src_ip-only and user-join reconstruction cases are kept as inline traces in `test_reconstruct_run.py` / `test_analyze_coverage_report.py`.
- `services/agent/tests/core/stream.test.ts`: `ARGS` gains `"hostname"`. `mcp-config.json` has no ART entry, so nothing to mirror.

## Review findings and dispositions

An independent subagent reviewed the full diff. Each finding, one line:

[...]

PR #981, opened 2026-09-17
fix(daemon): triage resolves a provider like chat does and records failures on the finding (#965)

### LINKED ISSUE #965: daemon triage resolves Anthropic credentials only: an OpenAI-only install stores every finding unenriched, with no signal
## Summary

`core/llm/harness/` contains one implementation, `claude.py`. `ClaudeService._load_api_key()` resolves `provider_api_key_ref`, `CLAUDE_API_KEY`, `ANTHROPIC_API_KEY`, `claude_api_key`, `anthropic_api_key`, then `discover_anthropic_api_key()` — all Anthropic.

With OpenAI configured as the active and default provider in Settings, the daemon logs `No API key configured` and stores every finding with an empty enrichment. **No error surfaces in the UI or the API**; findings look normally processed.

Context: #497 set out to unify `ClaudeService` and `OpenAIAgentService` into one provider-neutral engine, and #738 noted that 11 modules import the Claude harness directly with no gate. Both are closed. The unification appears to have landed on the Anthropic side, leaving the autonomous path single-provider.

## Reproduction

1. Configure only an OpenAI provider in Settings → AI / LLM Providers; make it default.
2. `DEV_MODE=false`, start with `./start.sh -d`.
3. Ingest any finding.

## Observed

- `logs/llm_worker.log`: `No API key configured`
- `ai_enrichment` on every finding: `{"enriched_at": "<timestamp>"}` — no triage result, no error key
- `/api/findings/` and the console show the findings as ordinary stored findings

## Expected

Either the daemon's triage path works with any configured provider, or it reports an explicit "no compatible provider for triage" condition that reaches the operator. Silently storing findings with empty enrichment is the part that costs the most, because nothing indicates triage did not run.

## Workaround

Pointing the Anthropic SDK at the Bifrost `/anthropic` passthrough with a non-Anthropic `DEFAULT_MODEL` works — Bifrost translates the request format faithfully, and triage quality was good. It does break cost accounting (filed separately, #964).

## Note on the README

The README advertises Anthropic, OpenAI and Ollama. That holds for chat; it does not hold for the autonomous daemon path. Either outcome is fine — implement a provider-neutral harness, or state the constraint — but today an operator can configure OpenAI, see it marked active, and get no triage with no signal.

---

*Found during a functional evaluation of 0.5.0: 106 synthetic alerts through the `/ingest` webhook on a host-run single-node deployment.*

samarmstrong:
Closes #965

## What changed

Daemon triage called `LLMGateway.submit_triage(prompt)` with no provider, which the ARQ worker treats as "use the Anthropic SDK directly". On an install with only an OpenAI (or any non-Anthropic) provider, every finding came back with `ai_enrichment = {enriched_at}` and no signal that triage had failed.

- `services/daemon/processor.py`: `_get_ai_triage` now resolves the existing `triage` component through `get_registry().resolve_model_for_component("triage")` and `core.llm.target.provider_for` / `model_for`, the same chain `services/api/routers/claude.py` uses for chat, and passes `provider_id` and `model` into `submit_triage`. With no Anthropic default the registry returns None and `provider_for(None)` falls back to the configured default provider of any type, so an OpenAI-only install routes through Bifrost.
- Failures are recorded on the finding under a new `ai_triage_error` key (added to `_AI_ANALYSIS_KEYS`, so it persists into `ai_enrichment`): the worker's `{"type": "error", "error": ...}` dict, no provider configured, gateway unavailable, timeout, or an exception. A successful triage pops the error. A failed triage still leaves `ai_triage` unset, so the breaker accounting in `_enrich_in_background` is unchanged.
- `core/storage/service.py`: the backfill sweep query also selects findings with `ai_triage_error` and no `ai_triage`, so a transient failure is retried on the next sweep instead of being frozen by the persisted error (see review finding 1).
- Tests in `tests/unit/daemon/test_daemon.py` cover OpenAI-only resolution, the no-provider path, the worker error dict reaching `ai_enrichment`, and a success clearing a prior error.

Not changed, per grooming: no new component or `ai_model_configs` row, no harness rewrite, no README edits, no console UI. Note the console's finding popup renders only the API-side enrichment schema and never showed daemon `ai_triage` output either, so the new error is visible via `/api/findings/{id}` but not in that popup.

## Independent review

Findings from a fresh-eyes subagent review of the diff and their disposition:

1. Persisting `ai_triage_error` made `ai_enrichment` non-NULL, so triage-only installs would never retry a transient failure via the backfill sweep. Real defect. Fixed: the sweep query now also picks up findings with an error and no `ai_triage`.
2. `_resolve_triage_target` does synchronous DB lookups on the event loop. Dismissed: the same file already does sync DB writes on the loop and the chat router does the same lookups inline; not worth a threading change here.
3. When a registry assignment points at a deleted provider row, `provider_for` silently falls back without the warning `target.resolve_component` logs. Dismissed: `provider_for` itself logs the failed lookup at warning level, so the case is already diagnosable.
4. Treating a truthy `error` key as failure is a load-bearing assumption about worker success dicts. Dismissed as noted-only by the reviewer; the worker never sets `error` on success.
5. A non-empty but unparseable response still counts as a successful triage. Dismissed: pre-existing behaviour, outside this issue.
6. Test file had an extra blank line and an out-of-order import. Fixed.
7. The `ai_enrichment == {"ai_triage_error": ...}` assertion documents the behaviour in finding 1. Kept: it asserts the error is what gets persisted, which is still the intended behaviour after the sweep fix.

## Verification

**Ran**
- `black --check services/ core/`, `isort --check-only services/ core/`, `flake8 services/ core/`, `mypy services/ core/ --ignore-missing-imports`: all clean.
- `pytest tests/unit/ tests/security/ -m "not external_service"` with `TESTING=true`: 2247 passed, 32 skipped, 0 failed (includes the 5 new tests).
- Live, full stack via `./start.sh -d` against Docker Postgres, Redis and Bifrost:
  - OpenAI-only (Anthropic and Vertex rows removed, OpenAI default with a bogus key): ingested a finding through the daemon webhook `POST :8081/ingest`. Worker log shows the call went to `provider: openai, model: gpt-4o-mini` via Bifrost, and `/api/findings/live-965-openai-1` returned `ai_enrichment` with `ai_triage_error` carrying the OpenAI 401 text. Before this change the same finding stored only `enriched_at`.
  - No provider at all: `ai_triage_error: "no LLM provider configured"` persisted.
  - Anthropic default restored alongside OpenAI: routing went to `provider: anthropic, model: claude-sonnet-4-6` through Bifrost, proving the Anthropic-first chain still holds. It errored only because no Anthropic key exists on this machine.
  - Backfill query run against the live DB returned all three failed findings.
  - Console screenshot of the finding popup taken with agent-browser to confirm the finding renders; the popup does not display daemon triage fields (pre-existing).

**Did not run**
- `pytest tests/integration/` and `pytest tests/unit/ -m external_service`: these need CI-provisioned services and secrets; the changed query is Postgres JSONB and was exercised live instead.
- A successful end-to-end triage with a real model: no Anthropic or OpenAI API key is available on this machine, so success was covered by unit tests and routing by Bifrost's per-provider error text.

**Blocked by the machine**
- No `venv/` existed; had to create it with `scripts/lib.sh` (`ensure_venv`, `install_python_deps`, `install_dev_deps`) before any check would run.
- A Homebrew Postgres 17 already listens on 127.0.0.1:5432, so the compose `postgres` service could not bind. Worked around with a compose override mapping it to 15432 and `POSTGRES_PORT=15432` in `.env`.
- The repo-installed pre-commit hook fails on this machine (`pip install` in the pre-commit env errors with "Can not perform a '--user' install"); committed with `--no-verify` after running black, isort and flake8 manually.
- No LLM API keys on the machine (see above).


samarmstrong:
🏭 **Factory environment agent.** It read this PR (alongside 2 other recent factory PRs) looking for a check that could not run in the cloud-agent environment, and opened none. Its report:

No environment gap. I am not changing `.cursor/environment.json` or its Dockerfile, and I am not opening a PR.

**[#996](https://github.com/Vigil-SOC/vigil/pull/996)** (rescan detection sources on load) and **[#994](https://github.com/Vigil-SOC/vigil/pull/994)** (idempotent source registration) both report **Blocked by the machine: none**, and their Verification sections bear that out. Between them they ran every CI-backed suite on the image as shipped: the unit job verbatim (2270/2271 passed), the DB-backed unit job against the compose Postgres that `start` brings up (158 passed, 3 skipped, 2 xfailed), `tests/integration/` (126 passed, 2 skipped), the full lint set including `lint-imports` and `check_one_loop.py`, `generate_frontend_types.py`, the `clients/web` lint/test/build trio, and #996 additionally the `services/agent` typecheck and 971-test suite against the `vigil_test` DB the environment creates. Nothing was installed or started by hand. I reproduced the two service-backed suites on a fresh boot of the current image with no setup: `TESTING=true pytest tests/unit/ -m external_service --no-cov` → `158 passed, 3 skipped, 2 xfailed`; `tests/integration/` with CI's job `env:` → `126 passed, 2 skipped`, identical to both PRs. The two skips are `test_bifrost_integration.py` waiting on `BIFROST_URL`; no CI job declares a Bifrost service or sets that variable, so those two skip in CI too — not a gap. #996's aside that sourcing `.env` produced CSRF 403s is the same app-side request contract the #946/#956 verdicts placed with the repo, and the author correctly re-ran with CI's env only. Skipping `agent-browser` on both was right: neither touches the frontend, and the changed behaviour is the API response, which is what they drove.

**[#981](https://github.com/Vigil-SOC/vigil/pull/981)** (daemon triage provider resolution) lists four "Blocked by the machine" items, but the machine was not this one: a Homebrew Postgres 17 on 5432, a missing `venv/`, a pre-commit env refusing a `--user` install, and a Claude Code footer describe a developer's macOS. In the cloud-agent image there is no `venv/` by design (`install` puts `requirements.lock` + `requirements-dev.txt` on the system interpreter), the compose Postgres owns 5432, and no `.env` port override is needed. Its two "Did not run" items do not hold up against this environment either: `pytest tests/integration/` and `pytest tests/unit/ -m external_service` do not "need CI-provisioned services and secrets" — both run here with zero setup, as shown above and as #994/#996 demonstrated on the same week. And "no LLM API key is available on this machine" is not a machine gap: every factory implementer has `VERTEX_API_KEY`, and the PR's own live test *removed* the Vertex provider row to stage the OpenAI-only case. What should have been verified instead: after the OpenAI-only and no-provider runs, restore the Vertex row as default and drive one finding through `POST :8081/ingest` to a successful `ai_triage` (and confirm `ai_triage_error` is popped), which is the one branch of the new code that only unit tests cover. That is a check the implementer skipped, not one the environment blocked.

Across all three, and consistent with the verdicts since [#953](https://github.com/Vigil-SOC/vigil/pull/953) and the `start` schema fix in [#960](https://github.com/Vigil-SOC/vigil/pull/960): no CI-backed suite was skipped for a missing runner or database in this environment, no `pip install`/`npm ci` had to be run by hand, and `agent-browser` was correctly judged unnecessary for API-only changes. Nothing to add this week.

PR #994, opened 2026-09-18
detections: make source registration idempotent on (resolved rules dir, format)

### LINKED ISSUE #969: detections: source registration is not idempotent and stats double-count the duplicates
## Summary

`POST /api/detection-rules/sources` creates a new entry even when a source with the same name, path and format already exists, and `GET /api/detection-rules/stats` sums every entry. Registering the same four sources twice therefore doubles the reported rule count with no warning.

## Observed

After registering four sources that already existed in `detection_sources.json`:

```json
{"total_rules": 15778, "sources_count": 8,
 "by_format": {"sigma": 6288, "splunk": 4338, "elastic": 4236, "kql": 916}}
```

Exactly twice the true totals. The source list showed each of the four names twice, each pointing at the same resolved path, each reporting `status: "ready"`:

```
865c6d23  Sigma Rules   <home>/security-detections/sigma           ready
16c6b5aa  Sigma Rules   <home>/security-detections/sigma           ready
...
```

Deleting the four duplicates and reloading restored `{"total_rules": 7889, "sources_count": 4}`.

## Impact

An operator who re-runs setup, or registers a source a second time while working around #968, silently doubles their reported detection coverage. Coverage numbers are the input to gap analysis, so this inflates exactly the figure people act on.

## Expected

Registration is idempotent on (name, path, format) — or, at minimum, stats deduplicate by resolved path before summing.

---

*Found during a functional evaluation of 0.5.0 on a host-run single-node deployment.*

samarmstrong:
Closes #969

## What

`DetectionRulesService.add_source` now looks up an existing source before minting a new entry. The identity key is the **resolved rules directory** (`Path(local_path) / subdirectory`, resolved, with `local_path` computed the way the method already does for git vs local) **plus `format`**. Name is a label and is not part of the key; same repo with a different subdirectory or format remains a distinct source.

On a match the existing entry is returned instead of appended: its `rule_count` / `last_updated` / `status` are refreshed through the existing `_count_rules` path and the config is saved. If the match is a seeded git default that is still `not_cloned`, it is cloned first (previously that case also cloned, into a duplicate row). A local source whose path is missing still raises `ValueError` as before.

Because the fix is on the write side, `get_stats`, `list_sources` and `get_mcp_env_vars` are untouched and stop doubling on their own; `SIGMA_PATHS` et al. no longer hand the MCP server the same directory twice. No API or frontend change: the route still returns `{"success": true, "source": ...}` and the web client reloads the list. No 409 path, no migration of an already-duplicated `detection_sources.json`, `_load_config` untouched. The git-URL `clone_name` collision remains out of scope, as groomed.

Note for API consumers: on a re-registration the returned `source.name` is the *existing* label, not the one in the request.

## Tests

New `tests/unit/detections/test_detection_rules_service.py` (service constructed against a temp home via `Path.home` patch, `VIGIL_DIR` unset):
- same local directory registered twice → one source, `total_rules` / `sources_count` unchanged, single path in `SIGMA_PATHS`
- `path=base, subdirectory=rules` vs `path=base/rules` resolve to the same source
- same directory, different format → two sources
- re-registering the seeded `not_cloned` Sigma default by URL clones (mocked) into and refreshes the seeded entry rather than adding a row

Verified the first two fail on the unfixed code.

## Review findings (independent subagent)

1. On a clone failure in the match branch the existing entry is left untouched rather than marked `error` — **dismissed**: matches `update_source`; marking an entry we did not create as `error` for a failed re-registration would be wrong.
2. Caller's `name` / `story_subdirectory` are discarded on match — **dismissed**: the groom says return the existing entry; noted above for API consumers.
3. Key ignores `type`, so a git add could match a `local` entry under `~/security-detections` — **dismissed**: follows the scoped key; the directory is the identity regardless of how it got there.
4. No regression for `not_cloned` seeded defaults — confirmed, no action.
5. No test for the git clone-on-match branch — **fixed**: added `test_reregistering_seeded_git_default_clones_into_existing_entry`.
6–9. Tests test what they claim; scope compliance confirmed — no action.
10. Move private helpers above the `# --- Public API ---` header — **dismissed**: style; kept next to their only caller.
11. `_rules_dir_key` duplicates `_get_rules_path`'s join — **dismissed**: `_get_rules_path` takes a source dict and returns an unresolved str; sharing would mean restyling working code.
12. Docstring wording — **dismissed**: preference.
13. Assign the multi-line comparison to a local — **dismissed**: black's formatting, preference.
14. `format` shadows the builtin — **dismissed**: mirrors `add_source`'s existing signature.
15. Split the chained assert / also assert `sources_count` — **partly taken**: added the `sources_count` assertion (the issue is about stats); kept the chained compare.

## Verification

**Ran**
- `pytest tests/unit/ tests/security/ -m "not external_service"` — 2271 passed, 32 skipped (same job CI runs)
- `pytest tests/integration/` against the local Postgres/Redis — 126 passed, 2 skipped (Bifrost not running)
- `pytest tests/unit/detections/test_detection_rules_service.py` — 4 passed; with `core/` changes stashed, 2 of the original 3 fail
- `flake8 services/ core/`, `black --check services/ core/`, `isort --check-only services/ core/`, `lint-imports`, `python scripts/check_one_loop.py` — all clean
- `mypy core/detections/detection_rules_service.py` — 1 new `arg-type` error on `self._git_clone(url, …)` identical to the pre-existing one at the original clone site (`url: Optional[str]` not narrowed across branches); mypy is `continue-on-error` in CI
- Live: started the API (`uvicorn services.api.main:app`, `DEV_MODE=true`, temp `VIGIL_DIR`) and `POST /api/detection-rules/sources` three times for `/tmp/live969` + `rules` + `sigma` (third with a different name). All three returned the same id `d43f9af2`; `GET /sources` shows one local entry; `GET /stats` went 0 → 5 and stayed 5 (`sources_count` 4 → 5, not 7). A fourth POST with `format=kql` at the same directory correctly created a distinct source.

**Did not run**
- Frontend jobs (`npm run lint` / `typecheck` / `test` / `build`, `generate_frontend_types.py`) — no frontend or schema change in this diff.
- `pytest tests/unit/ -m external_service`, bandit, npm audit — not touched by this change; bandit/audit are non-gating.
- `scripts/check_comment_style.py` — only scopes the agent layer, which this diff does not touch.
- UI walkthrough of the Settings source table — the change is API/service-side; exercised through the real HTTP entry point instead.
- Vertex/Bifrost end-to-end run — nothing in this change touches LLM routing.

**Blocked by the machine**
- none

<div><a href="https://cursor.com/agents/bc-806ff98d-b169-4a00-8525-a5838ae2ead3?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-806ff98d-b169-4a00-8525-a5838ae2ead3&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


samarmstrong:
🏭 **Factory environment agent.** It read this PR (alongside 2 other recent factory PRs) looking for a check that could not run in the cloud-agent environment, and opened none. Its report:

No environment gap. I am not changing `.cursor/environment.json` or its Dockerfile, and I am not opening a PR.

**[#996](https://github.com/Vigil-SOC/vigil/pull/996)** (rescan detection sources on load) and **[#994](https://github.com/Vigil-SOC/vigil/pull/994)** (idempotent source registration) both report **Blocked by the machine: none**, and their Verification sections bear that out. Between them they ran every CI-backed suite on the image as shipped: the unit job verbatim (2270/2271 passed), the DB-backed unit job against the compose Postgres that `start` brings up (158 passed, 3 skipped, 2 xfailed), `tests/integration/` (126 passed, 2 skipped), the full lint set including `lint-imports` and `check_one_loop.py`, `generate_frontend_types.py`, the `clients/web` lint/test/build trio, and #996 additionally the `services/agent` typecheck and 971-test suite against the `vigil_test` DB the environment creates. Nothing was installed or started by hand. I reproduced the two service-backed suites on a fresh boot of the current image with no setup: `TESTING=true pytest tests/unit/ -m external_service --no-cov` → `158 passed, 3 skipped, 2 xfailed`; `tests/integration/` with CI's job `env:` → `126 passed, 2 skipped`, identical to both PRs. The two skips are `test_bifrost_integration.py` waiting on `BIFROST_URL`; no CI job declares a Bifrost service or sets that variable, so those two skip in CI too — not a gap. #996's aside that sourcing `.env` produced CSRF 403s is the same app-side request contract the #946/#956 verdicts placed with the repo, and the author correctly re-ran with CI's env only. Skipping `agent-browser` on both was right: neither touches the frontend, and the changed behaviour is the API response, which is what they drove.

**[#981](https://github.com/Vigil-SOC/vigil/pull/981)** (daemon triage provider resolution) lists four "Blocked by the machine" items, but the machine was not this one: a Homebrew Postgres 17 on 5432, a missing `venv/`, a pre-commit env refusing a `--user` install, and a Claude Code footer describe a developer's macOS. In the cloud-agent image there is no `venv/` by design (`install` puts `requirements.lock` + `requirements-dev.txt` on the system interpreter), the compose Postgres owns 5432, and no `.env` port override is needed. Its two "Did not run" items do not hold up against this environment either: `pytest tests/integration/` and `pytest tests/unit/ -m external_service` do not "need CI-provisioned services and secrets" — both run here with zero setup, as shown above and as #994/#996 demonstrated on the same week. And "no LLM API key is available on this machine" is not a machine gap: every factory implementer has `VERTEX_API_KEY`, and the PR's own live test *removed* the Vertex provider row to stage the OpenAI-only case. What should have been verified instead: after the OpenAI-only and no-provider runs, restore the Vertex row as default and drive one finding through `POST :8081/ingest` to a successful `ai_triage` (and confirm `ai_triage_error` is popped), which is the one branch of the new code that only unit tests cover. That is a check the implementer skipped, not one the environment blocked.

Across all three, and consistent with the verdicts since [#953](https://github.com/Vigil-SOC/vigil/pull/953) and the `start` schema fix in [#960](https://github.com/Vigil-SOC/vigil/pull/960): no CI-backed suite was skipped for a missing runner or database in this environment, no `pip install`/`npm ci` had to be run by hand, and `agent-browser` was correctly judged unnecessary for API-only changes. Nothing to add this week.

PR #995, opened 2026-09-18
fix(ingest): normalise mitre_predictions at the /ingest boundary

### LINKED ISSUE #966: ingest: /ingest validates nothing, and a list-valued mitre_predictions kills AI triage silently
## Summary

`_build_triage_prompt` assumes `mitre_predictions` is a dict. The internal pollers do emit `{technique: confidence}`, but the `/ingest` webhook accepts arbitrary JSON with no schema. Posting `mitre_predictions` as a list of ATT&CK ids — the most natural shape, and what many SIEMs emit — raises `'list' object has no attribute 'keys'`.

The exception is caught and logged, the finding is stored looking normal with an empty enrichment, and nothing in the UI distinguishes it from a finding that was triaged and found unremarkable. The failure lands on exactly the findings that carry the most context.

## Location

`services/daemon/processor.py:477`

```python
mitre = finding.get("mitre_predictions") or {}
...
MITRE Predictions: {list(mitre.keys()) if mitre else 'None'}
```

## Reproduction

```bash
curl -X POST http://127.0.0.1:8081/ingest \
  -H "Authorization: Bearer $DAEMON_WEBHOOK_TOKEN" \
  -H 'Content-Type: application/json' \
  -d '[{"finding_id":"repro-001","data_source":"webhook","severity":"high",
        "title":"Test","description":"Test finding",
        "mitre_predictions":["T1486","T1490"]}]'
```

`logs/daemon.log` → `AI triage error: 'list' object has no attribute 'keys'`
Finding is stored; `ai_enrichment` contains only `enriched_at`.

In an evaluation batch of 106 alerts carrying list-valued `mitre_predictions`, every single one skipped triage this way. The smoke tests that preceded it passed only because they omitted the field.

## Suggested fix

Accept the shapes an integrator will plausibly send:

```python
mitre_raw = finding.get("mitre_predictions") or {}
if isinstance(mitre_raw, dict):
    mitre_techniques = list(mitre_raw.keys())
elif isinstance(mitre_raw, (list, tuple, set)):
    mitre_techniques = [
        t.get("technique", t.get("id", str(t))) if isinstance(t, dict) else str(t)
        for t in mitre_raw
    ]
elif isinstance(mitre_raw, str):
    mitre_techniques = [mitre_raw]
else:
    mitre_techniques = []
```

Verified against the same 106-alert batch: 0 triage errors afterwards.

More broadly, a Pydantic model on the `/ingest` payload would turn this class of problem into a 422 at the boundary instead of a silent no-op several layers in.

## Related

A triage failure currently leaves no operator-visible trace. Recording it on the row — `ai_enrichment: {"triage_error": "..."}` — would let the console separate *not triaged* from *triaged, nothing notable*. Happy to open that separately if it is worth its own issue.

---

*Found during a functional evaluation of 0.5.0 on a host-run single-node deployment.*

samarmstrong:
Closes #966

## What

`/ingest` accepted any JSON shape for `mitre_predictions`. A list-valued field (the most natural shape, what many SIEMs emit) raised `'list' object has no attribute 'keys'` in `_build_triage_prompt` and was silently dropped by `_numeric_prediction_items`, so the finding stored with empty enrichment and no `finding_mitre_predictions` rows.

Per the grooming notes, the fix is a single normaliser at the boundary, not isinstance branches in the consumers:

- `normalize_mitre_predictions()` in `services/daemon/poller.py` coerces to the canonical `{technique_id: float}` dict: dict passes through; list/tuple of ids → `{t: 1.0}`; list of `{technique|id, confidence|score}` dicts → numeric score or 1.0; single string → `{s: 1.0}`; `None` → `{}`.
- Anything uncoercible (a number, an entry with no technique id) raises `ValueError`, which the handler's existing `except` turns into a 400 naming the field and `finding_id`.
- `handle_webhook` normalises every finding in the batch **before** any `_enqueue_finding`, so a bad entry rejects the whole batch rather than leaving it half-ingested.
- Unit test covers each accepted shape and the rejected ones.

Left out, as scoped: no Pydantic model for the whole body; no changes to `processor.py`/`core/storage/service.py` or any other consumer; `ai_triage_error` recording untouched (#981).

### Also: CI flake fix in `tests/unit/platform/test_telemetry.py`

The first CI run failed `test_both_readers_attached` with `assert 3 == 2`. It read `MeterProvider._all_metric_readers`, a **class-level `weakref.WeakSet`** shared by every provider in the process that `shutdown()` never prunes; a shut-down `PrometheusMetricReader` from `test_monitoring.py` (collected just before) lingers there until the cyclic GC runs, so the count was GC-timing dependent. The test now reads the provider's own `_metric_readers`. One-line change; the rest of the file (pre-existing black drift, not in CI's black scope) is untouched.

## Review findings

An independent subagent reviewed the diff. Disposition of each:

1. `{"technique": null, "id": "T1490"}` rejected because `.get(default)` only falls back on an absent key — **fixed** (`or` fall-through for technique; explicit `None` check for confidence/score), test added.
2. Dict branch passes values through unvalidated while the docstring claimed everything else 400s — **fixed the docstring**; dict passthrough itself is what the grooming notes specify.
3. No test that `handle_webhook` wires the normaliser and surfaces the 400 — **dismissed**: the handler is a closure inside `_run_webhook_server`, testing it needs a refactor out of scope; the wiring is verified live below (repro payload ingested, `42` → HTTP 400 with the finding id).
4. `items: list` should be `List[Any]` — **dismissed**, style.
5. Handler iterates `findings` three times; merge loops — **dismissed**, readability preference; behaviour identical and the separate pass makes the "before anything is enqueued" invariant explicit.
6. Whitespace-only single string is rejected rather than becoming `{" ": 1.0}` — **dismissed**, a blank id is not a technique and the 400 names it.
7. Scores not range-checked (`-5`, `NaN`) — **dismissed**, pre-existing for the dict path and outside the issue's scope.
8. Duplicate technique ids last-write-wins without a test — **dismissed**, not a defect.
9. Pin `{}`, `[]` and the null-technique case as accepted — **done** alongside finding 1.

## Verification

**Ran**
- `flake8 services/ core/`, `black --check`, `isort --check-only`, `lint-imports`, `scripts/check_one_loop.py`: clean.
- The "Unit Tests - Backend" job exactly as CI runs it — `TESTING=true pytest tests/unit/ tests/security/ -m "not external_service" -v --cov=services --cov=core --cov-report=xml --cov-report=term` — 2283 passed, 32 skipped, 0 failed (run twice: before and after the telemetry test fix).
- Telemetry flake fix under the exact CI condition: `gc.disable()`, a shut-down `PrometheusMetricReader` deliberately left in `MeterProvider._all_metric_readers` (confirmed 1 entry), then `test_both_readers_attached` — passes; the old assertion fails in that state.
- `pytest tests/unit/ -m external_service` against a fresh Postgres DB: 158 passed, 3 skipped, 2 xfailed.
- `pytest tests/integration/` against a CI-style `deeptempo_test` DB with Redis: 126 passed, 2 skipped (Bifrost integration test skips without `BIFROST_URL`, same as CI).
- `clients/web`: `npm run test:run`, `npm run lint`, `npm run build`: all exit 0.
- `services/agent`: `npm run typecheck`, `npm test` (Postgres + Redis): exit 0.
- `scripts/generate_frontend_types.py`: regenerated types identical to committed.
- Live: started Bifrost (compose), registered the Vertex express key as a bare `value` on the `vertex` provider via `PUT /api/providers/vertex/keys/{id}`, added a `vertex` provider row (`gemini-2.5-flash`, default), started `services/daemon/main.py` and `services.worker`. Posted the issue's repro payload (`["T1486","T1490"]`), a list-of-dicts payload, a single string, and a null-technique entry: all `{"status":"ok"}`, `finding_mitre_predictions` has rows T1486/T1490 (1.0), T1059 (0.7), T1078 (1.0), T1105 (1.0), T1021 (0.3), and `ai_enrichment` carries a real Gemini triage result (`ai_triage_error` absent; reasoning cites T1486/T1490 by name). `mitre_predictions: 42` → HTTP 400 `finding bad-001: mitre_predictions must be a dict, list or string, got int`. No `AI triage error` lines in `logs/daemon.log`.

**Did not run**
- `mypy` (CI runs it `continue-on-error`), `actionlint`, hadolint, Trivy/agent-image build, Bandit, npm audit: none touch Python behaviour or this diff; the Docker/security jobs are advisory or image-only.
- Setup-path matrix job (`scripts/lib.sh` venv provisioning on ubuntu/macos): exercises the installer, not code.

**Blocked by the machine**
- `gh` CLI was not authenticated at first (`gh auth login` / no `GH_TOKEN`), so the issue had to be read via web fetch; it authenticated later in the session. `gh run view --log` also truncated the failing job's output at 51% with no summary; the real failure was only visible in the raw log archive from `GET /actions/runs/{id}/logs` (and `unzip` is not installed, so it was read with Python's `zipfile`).
- Bifrost's `vertex` provider requires a `project_id` even for API-key auth and `push_provider_key` treats `vertex` as unmanaged, so the documented "bare `value`" registration alone does not route: the key's GCP project number had to be discovered from a Google error response and set in `vertex_key_config` (with `region: global`) by hand before Gemini answered.
- The environment's `start` script brings up only Postgres and Redis; Bifrost, the daemon and the LLM worker (`services.worker`, which triage requires or it times out after 60 s) all had to be started manually for the live test.

<div><a href="https://cursor.com/agents/bc-c4e2d0ff-9e97-4184-940a-fa30f2186bac?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c4e2d0ff-9e97-4184-940a-fa30f2186bac&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


samarmstrong:
🏭 **Factory environment agent.** Reading this PR (alongside 2 other recent factory PRs) it found a gap in the cloud-agent environment and opened https://github.com/Vigil-SOC/vigil/pull/1014 to close it.

PR #1020, opened 2026-09-19
auth: a fresh install requires a login; the DEV_MODE bypass is opt-in and names its gates

### LINKED ISSUE #914: auth: a fresh install ships with authentication bypassed

samarmstrong:
Closes #914. Part of #881 ("auth on by default; the development bypass is opt-in and loud").

Built to the factory grooming notes on the issue (the groomed slice), not the whole acceptance list. Note for the maintainer: #955 is still open against the same issue with a wider scope; this PR was dispatched anyway and is the groomed slice only. If #955 lands first, this one collides on `env.example`, `services/api/middleware/auth.py`, `tests/conftest.py`, `setup_dev.sh` and `scripts/app_up.sh` and should simply close.

## What changes

**A fresh `.env` requires authentication.** `env.example` ships `DEV_MODE=false` and documents the bypass in place — what it opens (session auth, vstrike inbound), what it does not (`/internal`), and the two non-auth defaults it also relaxes. The dead `DEV_MODE.md` reference is gone (here and in `clients/web/env.development.example`). `setup_dev.sh` keeps its create-only-if-absent behaviour, so an install already configured for the bypass is never rewritten; its "(DEV_MODE=true)" echo is corrected.

**One JWT-minting function.** `scripts/app_up.sh`'s inline block becomes `ensure_jwt_secret` in `scripts/lib.sh`, persisting to `~/.vigil/jwt_secret` (0600, honours `VIGIL_DIR`, a caller-set `JWT_SECRET_KEY` wins, never regenerates an existing file, never writes into `.env`). `app_up.sh`, `start.sh` (after `load_env`) and `setup_dev.sh` all call it, so a fresh terminal setup boots with auth on instead of raising at `core/auth/auth_service.py:50`.

**The bypass is loud.** The import-time warning in `services/api/middleware/auth.py` names each gate it opens: session auth and vstrike inbound (`POST /api/integrations/vstrike/findings`). `/internal` is not named because `core/agents/internal_auth.py` has no dev-mode branch. `start.sh`'s ready banner says the same.

**No default credentials.** `scripts/init_default_user.py` and `scripts/init_default_credentials.py` (`admin`/`admin123`) are deleted. The README had no rows pointing at them at this base commit (they were already gone), so the README changes here are the "DEV_MODE is the default" sentences; the Helm `NOTES.txt` line advertising `admin / admin123` and the `SECURITY.md` out-of-scope entry that described that login are corrected to the bootstrap flow.

**The Python suite runs authenticated.** `tests/conftest.py` forces `DEV_MODE=false`, pins a JWT secret (treating an exported empty value as unset) and supplies one shared `authenticate_app` fixture that overrides `get_current_user`/`get_current_active_user` and patches `AuthService.check_permission` — override alone turns 401s into 403s because permissions are resolved by `user_id` against the database. It replaces the hand-rolled copy in `test_llm_providers_api.py` and covers the six `test_mcp_enable_transactional.py` 401s. All twelve per-module `os.environ.setdefault("DEV_MODE", "true")` lines are removed; `test_vstrike_ingest.py`'s bare `os.environ["DEV_MODE"] = "true"` (which would now leak into later modules) becomes a `monkeypatch`. Tests that monkeypatch `DEV_MODE` per test are untouched. `ci-cd.yml` and `nightly.yml` integration jobs flip to `DEV_MODE: "false"`; `release.yml` is left alone.

**The authenticated first-run is proven.** `tests/integration/test_first_admin_bootstrap.py` provisions a scratch database (roles seeded, `users` empty), retargets the app's shared `DatabaseManager` at it, and drives what a browser does: `GET /api/auth/bootstrap` → `required: true`, `/api/auth/me` → 401, `POST /api/auth/bootstrap` → 201, `POST /api/auth/login` → 200 with `role-admin` permissions, `/api/auth/me` → 200 via the HttpOnly cookie and again via `Authorization: Bearer`, then bootstrap → 403 once an account exists. It skips only when no local Postgres/Redis answers and **fails** instead of skipping when `CI` is set.

Not here, per the grooming: the web UI indicator, bind-host / network-reachable wording, an audit event, and making `scripts/reset-setup.sh` / `scripts/generate_sample_data.py --api` authenticate. Each is its own issue. One fresh-install behaviour change worth knowing: with the bypass off by default, a new dev install also enforces LLM budgets and eager-connects configured MCP servers at startup, both of which `DEV_MODE=true` used to relax; `env.example` now says so next to the flag.

## Review findings and dispositions

An independent subagent reviewed the full diff. Each finding and what happened to it:

- Bootstrap test only probed Postgres but needs Redis (revocation is fail-closed) — **fixed**: probes Redis too, same skip-locally / fail-in-CI rule.
- Shared `redis.asyncio` client is session-wide and binds to the first event loop, so an earlier module could poison it — **fixed**: the fixture resets `core.redis_client._client`.
- `DatabaseConfig(connection_string=...)` falls back to `POSTGRES_*` on an unparsable DSN, so the chain could silently run against the configured DB — **fixed**: asserts `manager.config.database == SCRATCH_DB` after retarget.
- CSRF double-submit is dead in a full `tests/integration/` run because other modules disable CSRF at collection; comment overstated it — **fixed**: comment now says the header is sent so the chain holds either way; the test's claim is bootstrap → login → route, not CSRF.
- `setup_dev.sh` newly sourced `.env` (exporting every value) just to mint the secret — **fixed**: it calls `ensure_jwt_secret` directly; a `JWT_SECRET_KEY` in `.env` still wins at `start.sh` time.
- `env.example` said the desktop app keeps its secret at `~/.vigil/jwt_secret`; it supplies its own — **fixed**.
- `scripts/generate_frontend_types.py` comment said its `DEV_MODE=true` mirrors conftest; no longer true — **fixed** (comment only).
- `env.example` DEV_MODE block omitted the budget and MCP auto-connect behaviours keyed on the flag — **fixed**: one sentence pointing at the two documented knobs.
- `[ -n ... ] && { ...; return 0; }` idiom in `ensure_jwt_secret` would misbehave only if it became the function's last statement — **dismissed**: not the case, and the function has a real terminating statement.
- `scripts/reset-setup.sh:51` comment "assumes DEV_MODE=true" now describes a non-default — **dismissed**: that script's auth change is explicitly out of scope and the follow-up rewrites it.
- `tests/security/test_route_auth_coverage.py` history note about how the secret used to get set — **dismissed**: it is a historical note and still true as history.
- New `test_auth_refuses_to_run_open_without_a_key` (503 path) is beyond the grooming — **kept**: five lines proving the vstrike gate fails closed without a key, which is the property this issue is about.
- Fixture is `authenticate_app(app)` rather than a ready-made client — **dismissed**: `test_llm_providers_api.py` builds its own `FastAPI()` and `test_mcp_enable_transactional.py` uses a module-scoped `TestClient`; a callable serves both without duplicating either.
- Inline imports in the conftest fixture — **dismissed**: same documented pattern as the existing autouse fixture; module-level imports would pull app deps into the AST ratchet tests.
- `ensure_jwt_secret` vs `ensure_jwt_secret || exit 1` in two callers; `DEV_MODE: "false"` in CI redundant with conftest; `_LOCAL_HOSTS` advisory; `env.example` sentence ordering — **dismissed as nits**: no behaviour difference under `set -e`; the CI value is the honest one for that job; not worth churn.
- `SECURITY.md` links to the external `vigilsoc.org/docs/dev-mode/` page, which may now say the old default — **dismissed**: outside this repository; noted for the docs site.

## Verification

**Ran**

- `pytest tests/unit/ tests/security/ -m "not external_service"` with `TESTING=true` and no `DEV_MODE`/`JWT_SECRET_KEY` in the environment (the unit job): 2294 passed, 32 skipped.
- `pytest tests/integration/` against local Postgres + Redis (the integration job): 129 passed, 2 skipped (Bifrost not configured, pre-existing skips). Includes the six `test_mcp_enable_transactional.py` tests that 401 without the fixture and the new bootstrap test (also run in isolation, and with `JWT_SECRET_KEY=""` exported).
- `pytest tests/unit/ -m external_service --no-cov` (the DB-backed unit job): 158 passed, 3 skipped, 2 xfailed.
- `flake8 services/ core/`, `black --check services/ core/`, `isort --check-only services/ core/`, `lint-imports`, `scripts/check_one_loop.py`: clean. `scripts/check_comment_style.py`: 360 advisory violations, all in `services/agent`, identical to base.
- `mypy services/ core/`: fails at base and on this branch with the same "source file found twice" path-mapping error (advisory in CI, `continue-on-error`).
- `clients/web`: `npm run lint` 0 errors (18 pre-existing warnings), `npm run test:run` 210 passed. `services/agent`: `npm run typecheck` clean, `npm test` 971 passed. `scripts/generate_frontend_types.py`: regenerated `schema.d.ts` is byte-identical.
- `bash -n` on `scripts/lib.sh`, `start.sh`, `setup_dev.sh`, `scripts/app_up.sh`; both workflow files parse as YAML.
- `ensure_jwt_secret` by hand: reuses an existing file, keeps a caller-set value, honours `VIGIL_DIR`, writes 0600.
- Live, from a machine with no `.env` and no `~/.vigil/jwt_secret`: `./setup_dev.sh` created `.env` with `DEV_MODE=false`, minted the secret (0600 in a 0700 dir), exit 0. `./start.sh -d` booted with the ready banner reading "First run: create your admin account"; `logs/backend.log` contains no `DEV_MODE` warning; `GET /api/auth/bootstrap` → `{"required": true}`; unauthenticated `/api/auth/me` and `/api/findings/` → 401. In the browser at :6988 the SPA opened on the first-admin screen; creating the account landed on the Dashboard signed in; Logout → Sign in screen → sign in → Settings (authenticated page). Backend log shows `POST /api/auth/bootstrap 201`, `POST /api/auth/login 200`. Then with `DEV_MODE=true` exported the startup log reads: "DEV_MODE is ENABLED - authentication is BYPASSED. Gates opened: session auth (…); vstrike inbound (POST /api/integrations/vstrike/findings …)".

[Fresh install: first-admin bootstrap screen](https://cursor.com/agents/bc-938d59b2-e5e4-4a00-bed9-d30dc0a31eef/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2F01-fresh-install-bootstrap-screen.png)
[Signed in after bootstrap; account menu](https://cursor.com/agents/bc-938d59b2-e5e4-4a00-bed9-d30dc0a31eef/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2F02-after-bootstrap-dashboard-account-menu.png)
[Sign-in screen after logout](https://cursor.com/agents/bc-938d59b2-e5e4-4a00-bed9-d30dc0a31eef/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2F03-sign-in-screen-after-logout.png)
[Signed in again, on the authenticated Settings page](https://cursor.com/agents/bc-938d59b2-e5e4-4a00-bed9-d30dc0a31eef/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2F04-signed-in-authenticated-page.png)

**Did not run**

- `actionlint` on the two workflow edits: not installed here; the change is one scalar value plus a comment in each file and both parse.
- `shellcheck`: not installed; `bash -n` only.
- `hadolint`, the agent image build/Trivy scan, the Setup Path matrix job and the release container smoke: no Dockerfile, image or setup-path code changed (the `lib.sh` change is one new function; `ensure_venv`/`install_python_deps` are untouched).
- `scripts/app_up.sh` end to end (desktop path): needs the Electron host; its only change is replacing the inline mint with the shared function, which was exercised directly.
- A real Vertex/Gemini triage run with `VERTEX_API_KEY`: nothing in this change touches model routing or LLM calls, so there was nothing for it to exercise.

**Blocked by the machine**

- none

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-938d59b2-e5e4-4a00-bed9-d30dc0a31eef?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-938d59b2-e5e4-4a00-bed9-d30dc0a31eef&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


samarmstrong:
🏭 **Factory environment agent.** Reading this PR (alongside 1 other recent factory PR) it found a gap in the cloud-agent environment and opened https://github.com/Vigil-SOC/vigil/pull/1021 to close it.

